PR #20, opened 2013-12-28
fix cask installer error

TakatoshiMaeda:
Hello. 
I Fixed a bug that was not installed by an error when the installer executes the `brew cast info`.

Be installed on a result of this change, the name of the correct cask has been specified, the installation will be interrupted if the name of the cask that does not exist is specified.
My environment success all tests.(Mac OS X 10.8.2, ruby 2.0.0p247)

Please confirm.
Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,7 @@ rvm:
   - jruby-18mode
   - jruby-19mode
   - jruby-head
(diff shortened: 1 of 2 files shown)


andrew:
I believe I fixed this a few weeks ago.

PR #28, opened 2015-01-24
Add Homebrew "tap command" support

mikemcquaid:
If this repo is renamed to e.g. `andrew/homebrew-brewdler` (or `Homebrew/homebrew-bundler` or whatever) then with this PR merged one will be able to:
```bash
brew tap mikemcquaid/brewdler # currently working but could instead be andrew/brewdler or Homebrew/bundler or whatever.
brew brewdler
```

The previous `gem` installation method will still work and this will just provide an alternate installation method that doesn't require any messing with `PATH`, any use of `gem` and autoupdates with `brew update`.

The implementation is still a bit hacky as-is so feel free to request I change the approach here and I'm happy to do so.

### DIFF
--- a/bin/brewdle
+++ b/bin/brewdle
@@ -1,7 +1,13 @@
 #!/usr/bin/env ruby
 
 require 'brewdler'
-require 'commander/import'
+
+begin
+  require 'commander/import'
+rescue LoadError
+  Brewdler::Commands::Install.run
+  exit
+end
 
 program :version, Brewdler::VERSION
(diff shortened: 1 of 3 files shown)


andrew:
Thanks @mikemcquaid, I've renamed the repo to `homebrew-brewdler`, happy to move it over to @Homebrew if it will help get more exposure and :heart:

Just tried it out and it works great, although if we move it to homebrew maybe it should be called `Homebrew/homebrew-bundle` instead so the command makes more sense: `brew bundle`

PR #29, opened 2015-01-27
More tap tweaks

mikemcquaid:
- Fix a bunch of typos/casing pedantry
- Rename the command to `brew brewdle` to better match the documentation

@andrew I've also created a @Homebrew/bundle team and added you to it. That should give you enough permissions to be able to migrate this repository to the Homebrew organisation and rename it to whatever name you think fits best. When you've done that I'll add you and any other collaborators (just @jamesaanderson?) as administrators on the repo and you can continue to run it as you see fit. I'll then point the documentation to note this is the "official" implementation. Sound good?


andrew:
Thanks @mikemcquaid, will move it shortly :+1:

PR #30, opened 2015-01-27
Readme.md: correct homebrew-cask tap

vitorgalvao:
homebrew-cask has been an organization for a while now, no longer under `phinze/homebrew-cask`, but `caskroom/homebrew-cask`.


andrew:
Thanks @vitorgalvao

PR #34, opened 2015-01-28
Remove mime-types dependency

josh:
I have no idea why this was added in 99dab3359f3fa602fac63487703268ab83232392. Its not used by this library at all.

cc @andrew @jamesaanderson


andrew:
Thanks @josh

PR #35, opened 2015-01-28
Detect presence of brew and brew-cask commands

josh:
Simply checks if `brew` or `brew-cask` load path rather than trying to execute and dump the output of expensive commands.

Invoking `brew info` on my machine adds an additional 2 seconds to every brew install and prints useless data like `35 kegs, 91540 files, 1.4G` to the output.

cc @mikemcquaid


mikemcquaid:
Sounds good to me. Something slightly more involved that's very quick is `brew --prefix &>/dev/null`

andrew:
Thanks again @josh!

Issue #32, opened 2015-01-27
Feature Request: improve security


xu-cheng:
Brewfile is much like a Gemfile. However, the brewdler  implementation is a lot of simpler. Currently, we use the evil `eval` to parse Brewfile and string concatenation to generate system command. So there will be a lot of security vulnerabilities. Hence, the question is can we improve it like what gem did?

I'm not a security expert. But I think we can do better.

andrew:
Yeah definitely, this is a big issue :+1:

mikemcquaid:
:+1:

jamesaanderson:
It looks as if Bundler uses eval as well: https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L32. I may be skipping over something though

mikemcquaid:
I don't think `eval` is inherently bad but `instance_eval` is at least a little nicer/more secure.

andrew:
@xu-cheng are there any more changes you'd like to make regarding this issue or can I close it?

xu-cheng:
Yeah, we can close it.

PR #36, opened 2015-01-31
Change eval to instance_eval

jamesaanderson:
I think we should keep #32 open for now but I changed `eval` to `instance_eval` which should be a bit more secure. Bundler seems to use `instance_eval` as well.


mikemcquaid:
If it just works as-is: :+1: from me.

jamesaanderson:
Alright. I'm going to merge then

Issue #31, opened 2015-01-27
Feature Request: dump existing install into Brewfile


xu-cheng:
Since this becomes homebrew official method to bundle homebrew install. I hope we can add more feature to it. Shall we?

So instead of reading Brewfile and install them, I hope to do the reverse that is writing current installation to a Brewfile. I have wrote some [similar tool](https://github.com/xu-cheng/homebrew-mine/blob/master/cmd/brew-bundle.rb) before. And if it's OK, I can work on a PR.

mikemcquaid:
> So instead of reading Brewfile and install them, I hope to do the reverse that is writing current installation to a Brewfile. I have wrote some similar tool before. And if it's OK, I can work on a PR.

Sounds good!

xu-cheng:
Currently, the tool is available both on homebrew tap cmd and gem. I wonder can we drop the gem support? Because it can help to do:

* Load homebrew core library much easier.
* Make it complete dependency free.
* Parse the ARGV easier. Otherwise I have do same thing twice: one for gem and one for tap cmd.

mikemcquaid:
I'm :+1: on that if @andrew is.

andrew:
Both of those sound good, I'll push one one last release that prints out a deprecation warning when using the gem and then we can remove the rubygem specific parts.

PR #39, opened 2015-02-03
drop gem support

xu-cheng:
Per #31, #32.

This PR drops the gem support and swtiches to brew cmd only.

Things have been done:
* [x] Remove `commander` dependency.
* [x] Use homebrew ARGV to parse arguments.


andrew:
Thanks @xu-cheng, do we even need `brewdler.gemspec` at all now that it's no longer a gem? We can move the dev dependencies into `Gemfile` so travis continues to work.

mikemcquaid:
> Use homebrew internal method to install formulae.

I think we want to avoid this as the Ruby API is not public.

cmd/brew-brewdle.rb
```diff
@@ -1,5 +1,38 @@
...
+BREWDLER_LIB = Pathname.new(BREWDLER_ROOT)/"lib"
+
+$LOAD_PATH.unshift(BREWDLER_LIB)
+
+require "brewdler"
+
+usage = <<-EOS.undent
+  brew brewdle
+  brew brewdle [--version]
+  brew brewdle [-h|--help]
```
mikemcquaid: Pretty sure these won't be printed as `brew.rb` will grab them instead.

cmd/brew-brewdle.rb
```diff
@@ -1,5 +1,38 @@
...
+if ARGV.flag?("--help") || ARGV.flag?("-h")
+  puts usage
+  exit 0
+end
+
+if ARGV.empty?
+  Brewdler::Commands::Install.run
+else
+  puts usage
+  exit 1
```
mikemcquaid: `abort usage` will probably work here.

lib/brewdler/repo_installer.rb
```diff
@@ -1,10 +1,13 @@
+require "cmd/tap"
```
mikemcquaid: I'd rather we didn't include Homebrew internals here.

cmd/brew-brewdle.rb
xu-cheng: I just tested. It is printed.

mikemcquaid:
> Thanks @xu-cheng, do we even need brewdler.gemspec at all now that it's no longer a gem? We can move the dev dependencies into Gemfile so travis continues to work.

Makes sense to me :+1:

lib/brewdler/repo_installer.rb
andrew: Agreed :+1:

xu-cheng:
> I think we want to avoid this as the Ruby API is not public.

> I'd rather we didn't include Homebrew internals here.

I just want to avoid shell injection so it becomes more secure. OK I will remove them. And find another way to boost the security.

xu-cheng:
> Thanks @xu-cheng, do we even need brewdler.gemspec at all now that it's no longer a gem? We can move the dev dependencies into Gemfile so travis continues to work.

Well, since I have removed the brew internal method invoking. The CI works fine. As for `brewdler.gemspec` I'm not quite familiar with it. So I don't know how to do it.

andrew:
@xu-cheng you can remove `brewdler.gemspec` and replace the contents of `Gemfile` with:

```ruby
source 'http://rubygems.org'

gem 'rspec'
gem 'rake'
gem 'simplecov'
gem 'coveralls'
```

then everything "should" continue to work as expected.

xu-cheng:
@andrew No, it's not working. https://travis-ci.org/Homebrew/homebrew-brewdler/builds/49305423

mikemcquaid:
@xu-cheng Probably need to remove `require 'bundler/gem_tasks'` from `Rakefile`.

andrew:
@xu-cheng yeah, I just checked, what @mikemcquaid said works for me locally.

xu-cheng:
:+1: Updated.

andrew:
Also, don't merge this until I've published one last version of the rubygem please.

xu-cheng:
@mikemcquaid I have some thought regrading security issue on shell invoking. The `system` implemented by homebrew is much safer than backticks. So should I use homebrew's `system` directly or copy paste the code?

mikemcquaid:
> @mikemcquaid I have some thought regrading security issue on shell invoking. The system implemented by homebrew is much safer than backticks. So should I use homebrew's system directly or copy paste the code?

Use `Kernel.system` or `IO.popen` instead.

xu-cheng:
Is there any ETA for this PR? I have already finished implementing the `bundle` feature I mentioned in #31. But I need this one got merged first.

cmd/brew-brewdle.rb
```diff
@@ -1,5 +1,37 @@
...
+
+  Usage:
+  Bundler for non-Ruby dependencies from Homebrew
+
+  Options:
+  -h, --help        show this help message and exit
+  --version         show the version of brewdler
+EOS
+
+if ARGV.flag?("--version")
```
jamesaanderson: I'd add -v as well

xu-cheng: Shouldn't we use `-v` for the short form of `--verbose` in the future?

jamesaanderson: Didn't think about that. I generally see -v used for version but it could cause some confusion

xu-cheng: If it helps, homebrew uses `-v` for `--verbose`.

jamesaanderson: Yeah then I'd leave it

xu-cheng:
I updated the way to parse ARGV, make it more friendly for future extension. Also squash everything to single commit.

cmd/brew-brewdle.rb
```diff
@@ -1,5 +1,39 @@
...
+  -h, --help        show this help message and exit
+  --version         show the version of brewdler
+EOS
+
+if ARGV.include?("--version")
+  puts Brewdler::VERSION
+  exit 0
+end
+
+if ARGV.flag?("--help")
```
xu-cheng: I just find out `ARGV.flag? "--foo"` is equivalent to `ARGV.include? "--foo" || ARGV.include? "-f"`. So updated it a bit.

Gemfile
```diff
@@ -1,4 +1,6 @@
 source 'http://rubygems.org'
 
-# Specify your gem's dependencies in brewdler.gemspec
-gemspec
+gem 'rspec'
```
mikemcquaid: Probably want these to be `group :development do` or `group :test do` blocks (just for cleanliness)

cmd/brew-brewdle.rb
mikemcquaid: Homebrew (like Ruby) uses `-v` for verbose when other arguments are specified and for the version as a single argument.

cmd/brew-brewdle.rb
mikemcquaid: I wonder if we want to rely on Homebrew's internal methods here.

Gemfile
xu-cheng: Fixed.

mikemcquaid:
@andrew Let us know when you've done a final gem release, thanks man :+1:

cmd/brew-brewdle.rb
xu-cheng: I think this is inevitable. Since we drop the gem, the homebrew env will definite be there. Say we implement ARGV extend here again, the homebrew extension will still exist. In the result, it would be hard to see which code is acutely running, homebrew or brewdler. 

Also for the heads up, [my implement](https://github.com/xu-cheng/homebrew-brewdler/compare/Homebrew:master...xu-cheng:bundle) of #31 requires homebrew internal method too. Namely, I need the `version`.

andrew:
Just :ship: the last version ever: https://rubygems.org/gems/brewdler/versions/1.1.1

Merge away! (Once you've rebased against that last change)

xu-cheng:
I have rebased it. If @mikemcquaid has no objection of using `ARGV.flag?`, I think we are good to go.

andrew:
@xu-cheng There's still a conflict showing in the github UI

xu-cheng:
@andrew Sorry, didn't notice the master has been updated. Already fixed it.

andrew:
Great work, thanks @xu-cheng!

cmd/brew-brewdle.rb
mikemcquaid: Ok. I think this use of `ARGV` should be fine but we need to be really careful when using Homebrew internal methods as they are effectively private APIs that could change at any time. Whenever external tools have used them in the past we've ended up (accidentally) periodically breaking those tools.

PR #40, opened 2015-02-04
Change backticks to system #32

xu-cheng:


lib/brewdler/brew_installer.rb
```diff
@@ -2,14 +2,14 @@ module Brewdler
   class BrewInstaller
     def self.install(name, options = {})
       if `which brew`; $?.success?
```
mikemcquaid: Might want to use it for this command too and perhaps do `brew --prefix`.

lib/brewdler/brew_installer.rb
```diff
@@ -2,14 +2,14 @@ module Brewdler
...
+        command = [name]
         unless options[:args].nil?
           options[:args].each do |arg|
-            command << " --#{arg}"
+            command << "--#{arg}"
           end
         end
 
-        `#{command}`
+        system "brew", "install", *command
```
mikemcquaid: Might be worth checking that the stdout/stderr isn't printed now when it wasn't before (or vice-versa)

xu-cheng: I think wouldn't be better to print out current state. So if I have a long list in the `Brewfile`, I can know the current progress.

lib/brewdler/brew_installer.rb
xu-cheng: Updated. Now it's using `brew --prefix` and `brew cask --version` to check `brew` and `cask`.

lib/brewdler/brew_installer.rb
mikemcquaid: Oh, I just mean that we're consistent with how it behaved before.

xu-cheng: I understand. And I think this inconsistence is a improvement.

mikemcquaid: I think it should start with a `--verbose` flag/environment variable for now and we could talk about changing the behaviour in a future PR.

xu-cheng: OK. One thing is what is the easiest way to hide the output? `system "brew", "install", "foo", "&>/dev/null"` doesn't work. Only `system "brew install foo &>/dev/null"` works. But I don't want to use the latter one as this PR's whole point is to avoid shell injection(by using list to pass command).

Second thing, should I hide stdout only(like the current behavior) or hide both stdout and stderr(which is little strange, as it will hide errors in the mean time.)?

mikemcquaid: You probably need to use `IO.popen` for this case. It's a bit of a nastier API but it's probably worth it longer term for the flexibility it provides.

xu-cheng: Updated.

lib/brewdler/utils.rb
```diff
@@ -0,0 +1,10 @@
+module Brewdler
+  def self.system cmd, *args
+    pid = fork do
+      $stdout.reopen("/dev/null")
+      exec(cmd, *args)
+    end
```
mikemcquaid: I'd probably still recommend `IO.popen` as it gives us more flexibility in future to do things like e.g. only print `stdout` on failure or a `--verbose`, combine `stderr` into the output etc. Your call though as you're doing the work :grinning:

lib/brewdler/utils.rb
```diff
@@ -0,0 +1,10 @@
...
+    Process.wait(pid)
+    $?.success?
```
mikemcquaid: I reckon either we should make sure we always check the exit codes or raise an exception here. Your call.

lib/brewdler/utils.rb
xu-cheng: I did try `IO.popen`. But it may be too nasty for me to handle. I don't know how to use it to handle `stderr` or adding `--verbose` feature in the future. Sorry about this.

lib/brewdler/utils.rb
xu-cheng: I think exit code is better. As we can handle it more flexible on the outside.

lib/brewdler/utils.rb
mikemcquaid: This may be useful (you can ignore the timeout stuff), but again, feel free to ignore it entirely: 
https://github.com/mikemcquaid/HookHand/blob/master/hookhand.rb#L123-L150

lib/brewdler/utils.rb
mikemcquaid: :+1:

PR #41, opened 2015-02-04
Make CI work in OS X with homebrew installed

xu-cheng:
This PR didn't change too much at the moment but should be helpful for future development.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
-language: ruby
+language: objective-c # Force OS X
 rvm:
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Ooo, I like the hack to get it using OSX. Nice work. I'd also love something that we can use instead of the (annoying) @coveralls spam in here.

andrew:
@mikemcquaid I've disabled the coveralls comments now.

andrew:
:shipit:

PR #43, opened 2015-02-06
Dump installation into Brewfile 

xu-cheng:
Per #31


cmd/brew-brewdle.rb
```diff
@@ -9,15 +9,21 @@
...
   brew brewdle
+  brew brewdle dump [--force] [--gen-cleanup-installed]
   brew brewdle [--version]
   brew brewdle [-h|--help]
 
   Usage:
   Bundler for non-Ruby dependencies from Homebrew
 
+  brew brewdle                    read Brewfile and install all dependencies
+  brew brewdle dump               dump current installation into Brewfile
```
mikemcquaid: `write all currently installed packages into a Brewfile`

cmd/brew-brewdle.rb
```diff
@@ -9,15 +9,21 @@
...
+
   Options:
-  -h, --help        show this help message and exit
-  --version         show the version of brewdler
+  --force                        force overwrite existed Brewfile
+  --gen-cleanup-installed        also generate `.brew-cleanup-installed` file
```
mikemcquaid: Let's pass on this; that could be pulled into that tool instead and/or we could migrate that tool to uninstall everything not in a `Brewfile`

lib/brewdler/brew_dumper.rb
```diff
@@ -0,0 +1,42 @@
+require "json"
+require "version"
+
+module Brewdler
+  class BrewDumper
+    attr_reader :formulae
+
+    def initialize
+      if system "brew --prefix &>/dev/null"
```
mikemcquaid: Let's maybe pull this check into another function.

lib/brewdler/brew_dumper.rb
```diff
@@ -0,0 +1,42 @@
...
+        raise "Unable to list installed formulae. Homebrew is not currently installed on your system."
+      end
+    end
+
+    def to_s
+      @formulae.map do |f|
+        if f[:args].empty?
+          "brew '#{f[:name]}'"
+        else
+          "brew '#{f[:name]}', args: [#{f[:args].map { |arg| "'#{arg}'" }.join(", ")}]"
```
mikemcquaid: Nice work but I think we should pull the `args` stuff onto separate lines to make them more readable.

cmd/brew-brewdle.rb
xu-cheng: OK

lib/brewdler/cask_dumper.rb
```diff
@@ -0,0 +1,17 @@
+module Brewdler
+  class CaskDumper
+    attr_reader :casks
+
+    def initialize
+      if system "brew cask --version &>/dev/null"
```
mikemcquaid: Again, let's pull this into a function.

lib/brewdler/dumper.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def initialize
+      @brew = BrewDumper.new
+      @cask = CaskDumper.new
+      @repo = RepoDumper.new
+    end
+
+    def dump_brewfile
+      write_file Pathname.new(Dir.pwd)/"Brewfile",
+                 [repo, brew, cask].map(&:to_s).join("\n") + "\n",
```
mikemcquaid: I'd pull this into several statements to improve readability.

mikemcquaid:
Nice work here. I suspect they'll be a bit of back-and-forth but nice to see a great feature here (with tests!) in development.

lib/brewdler/brew_dumper.rb
```diff
@@ -0,0 +1,42 @@
+require "json"
+require "version"
```
xu-cheng: @mikemcquaid You may be against this, as I need to import Homebrew's `version` to do the version comparison.

xu-cheng:
Any idea why the CI test cannot find homebrew?

xu-cheng:
After so many back-and-forth, I finally make CI work.

lib/brewdler/brew_dumper.rb
mikemcquaid: What's the use-case of it below? Being stupid but it's not totally clear to me why it's required.

xu-cheng: The idea is if there are multi kegs installed, we need to choose one and use the build options of that one as the args. 

To choose, we first use the linked keg. But if none of them is linked, I would select the latest version. Hence, we require `version` to do the comparison.

lib/brewdler/brew_dumper.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    private
+
+    def formula_inspector f
+      installed = f["installed"]
+      if f["linked_keg"].nil?
+        keg = installed.max { |a,b| Version.new(a["version"]) <=> Version.new(b["version"]) }
+      else
+        keg = installed.detect { |k| f["linked_keg"] == k["version"] }
+      end
```
xu-cheng: @mikemcquaid this is where `version` is required.

lib/brewdler/brew_dumper.rb
mikemcquaid: Might be worth just using the `opt` symlink for that.

xu-cheng: I do choose `opt` aka linked keg if possible. I only use `version`, if none is linked, that is no `opt` symlink exists

mikemcquaid: The `opt` symlink should always be present whereas `linkedkeg` may not be. If it's useful we could consider adding more to the JSON output, here.

xu-cheng: How about we do the version sort in the homebrew, so we only need use the last one from the JSON output?

mikemcquaid: :+1:

xu-cheng:
Any comment @andrew ?

andrew:
This looks great, a very nice refactoring as well as the dump feature :+1:

xu-cheng:
Is it good enough for merging? Or are there any other comments?

mikemcquaid:
Looks good to me :+1: @andrew, any objections?

andrew:
:shipit:

xu-cheng:
@andrew Maybe you want to update the README or something :tada:

andrew:
@xu-cheng https://github.com/Homebrew/homebrew-brewdler/commit/dfc82915ad203e48b0ca5610e27308f40951b7a0 :snowboarder:

mikemcquaid:
@xu-cheng As a follow-up you could probably add a `cleanup` or `cleanup --force` which does what `brew cleanup-installed` does?

xu-cheng:
Yeah, I will put that in my list.

PR #47, opened 2015-02-15
decouple brewfile parsing and installation

xu-cheng:


andrew:
Thanks @xu-cheng!

lib/brewdler/dsl.rb
```diff
@@ -1,23 +1,47 @@
...
       instance_eval(@input)
+      self
+    end
+
+    def install
+      @entries.each do |entry|
+        case entry.type
+        when :brew then Brewdler::BrewInstaller.install(entry.name, entry.extra[:options])
+        when :cask then Brewdler::CaskInstaller.install(entry.name)
+        when :repo then Brewdler::RepoInstaller.install(entry.name)
```
mikemcquaid: Would be nice to be able to set just the class name and extra as options here so you don't need to repeat quite so much.

lib/brewdler/dsl.rb
```diff
@@ -1,23 +1,47 @@
...
+    attr_reader :entries
+
     def initialize(input)
       @input = input
+      @entries = []
     end
 
     def process
       instance_eval(@input)
+      self
```
mikemcquaid: Seems weird to return self here.

lib/brewdler/dsl.rb
```diff
@@ -1,23 +1,47 @@
...
     end
 
     def cask(name)
-      Brewdler::CaskInstaller.install(name)
+      @entries << Entry.new(:cask, name)
     end
 
     def tap(name)
-      Brewdler::RepoInstaller.install(name)
+      @entries << Entry.new(:repo, name)
```
mikemcquaid: Similarly would be nice to kill some of the duplication here a bit.

lib/brewdler/dsl.rb
```diff
@@ -1,23 +1,47 @@
 module Brewdler
   class Dsl
+    class Entry
```
mikemcquaid: What's an Entry, a line?

lib/brewdler/dsl.rb
xu-cheng: Could you be more specific?

lib/brewdler/dsl.rb
xu-cheng: Not too weird to me. It is used a lot in JAVA to do something like `object.foo().bar()`

lib/brewdler/dsl.rb
xu-cheng: An entry is a `brew`, `cask` or `tap` command.

lib/brewdler/dsl.rb
xu-cheng: How

lib/brewdler/dsl.rb
mikemcquaid: e.g.

```ruby
@entries.each do |entry|
  class = options = nil
  class = case entry.type
  when :brew
    options = entry.extra[:options]
    Brewdler::BrewInstaller
  when :cask then Brewdler::CaskInstaller
  when :repo then Brewdler::RepoInstaller
  end
  class.install(entry.name, options)
end
```

lib/brewdler/dsl.rb
```diff
@@ -1,23 +1,47 @@
...
+        when :brew then Brewdler::BrewInstaller.install(entry.name, entry.extra[:options])
+        when :cask then Brewdler::CaskInstaller.install(entry.name)
+        when :repo then Brewdler::RepoInstaller.install(entry.name)
+        end
+      end
     end
 
     def brew(name, options={})
-      Brewdler::BrewInstaller.install(name, options)
+      @entries << Entry.new(:brew, name, {:options => options})
```
mikemcquaid: Normally I'd expect `options` itself to be the hash rather than a subhash.

lib/brewdler/dsl.rb
mikemcquaid: Meh, I guess it doesn't really matter here but could have a `def add_entry(type, name, options={})`

lib/brewdler/dsl.rb
xu-cheng: Here, I store `options` as a sub hash inside `extra`. The reason I use `extra` rather than `options` is I want to make it easier to extend for possible need in the future. By design, we can store anything in `extra`.

mikemcquaid: I think that's probably overkill, personally. If a single hash is enough now let's just use that and we can always nest them later if we need to (there won't be external API consumers so should be easy to do).

xu-cheng: Just a reminder, this PR has been merged. You are welcome to do any refactoring as you see fit though.

PR #48, opened 2015-02-16
add cleanup feature

xu-cheng:
@mikemcquaid


cmd/brew-brewdle.rb
```diff
@@ -10,6 +10,7 @@
 usage = <<-EOS.undent
   brew brewdle
   brew brewdle dump [--force]
+  brew brewdle cleanup [--yes]
```
mikemcquaid: Maybe just go for `--force` again.

lib/brewdler/utils.rb
```diff
@@ -1,5 +1,5 @@
-module Brewdler
-  def self.system cmd, *args
+module Brewdler extend self
```
mikemcquaid: I think it's clearer as-is.

lib/brewdler/utils.rb
```diff
@@ -8,11 +8,28 @@ def self.system cmd, *args
...
+  def getc  # NOTE only tested on OS X
+    Kernel.system "/bin/stty raw -echo"
+    if STDIN.respond_to?(:getbyte)
+      STDIN.getbyte
+    else
+      STDIN.getc
+    end
+  ensure
+    Kernel.system "/bin/stty -raw echo"
   end
```
mikemcquaid: I'd maybe just make this non-interactive as Homebrew is intentionally non-interactive as a design choice.

lib/brewdler/utils.rb
xu-cheng: OK

cmd/brew-brewdle.rb
xu-cheng: OK

lib/brewdler/utils.rb
xu-cheng: I think `cleanup` is a kind of dangerous action. So I would like to ask for confirmation to users. What if you choose the wrong Brewfile? 

However, you can always make it non-interactive using `--force` option.

mikemcquaid: Yeh, I think you could just always require `--force`. I don't think introducing interactive things into Homebrew is really what we want.

xu-cheng: > I think you could just always require --force

Then, what is the point of using `--force`. IMHO, I prefer current design. It has both interactive and non-interactive features. So everyone is happy.

mikemcquaid: I'm not happy :wink: because we're introducing interactive things into Homebrew and I fear people will want them in more places. Could maybe make it not require `--force` for removing packages that were installed from bottles because it's trivial to reinstall them?

xu-cheng: > Could maybe make it not require --force for removing packages that were installed from bottles because it's trivial to reinstall them?

That's a whole different feature which is complicated and not very needed. :santa:

> I fear people will want them in more places.

There will be some place interactive thing has more benefit. In fact, this code is copied from homebrew install script. And quite frankly, this repo is not homebrew core.

I kinda want to hear more people's opinions on this. cc @andrew If indeed majority people hate interactive, I would suggest to implement this as `cleanup` and `cleanup --dry-run`. Then maybe I put interactive script into my own tap.

lib/brewdler/utils.rb
```diff
@@ -9,10 +9,27 @@ def self.system cmd, *args
...
+    Kernel.system("brew cask --version &>/dev/null")
+  end
+
+  def self.brewfile
+    File.read(Dir['{*,.*}{B,b}rewfile'].first.to_s)
+  rescue Errno::ENOENT
+    raise "No Brewfile found"
+  end
+
+  def self.getc  # NOTE only tested on OS X
```
mikemcquaid: Can probably remove this comment as the README says this only supports OSX.

lib/brewdler/utils.rb
mikemcquaid: Yeh, fair enough that it's in the installer. I concede my point :wink:

lib/brewdler/utils.rb
xu-cheng: Removed

lib/brewdler/utils.rb
xu-cheng: :tada:

cmd/brew-brewdle.rb
```diff
@@ -18,9 +19,10 @@
 
   brew brewdle                    read Brewfile and install all dependencies
   brew brewdle dump               write all currently installed packages into a Brewfile
+  brew brewdle cleanup            uninstall all Homebrew formulae not listed in Brewfile
 
   Options:
-  --force                        force overwrite existed Brewfile
+  --force                        force overwrite existed Brewfile or cleanup
```
mikemcquaid: `overwrite existing Brewfile or cleanup without prompting`?

xu-cheng: Done

cmd/brew-brewdle.rb
```diff
@@ -18,9 +19,10 @@
 
   brew brewdle                    read Brewfile and install all dependencies
   brew brewdle dump               write all currently installed packages into a Brewfile
+  brew brewdle cleanup            uninstall all Homebrew formulae not listed in Brewfile
 
   Options:
-  --force                        force overwrite existed Brewfile
+  --force                        force overwrite existed Brewfile or cleanup without prompting
```
mikemcquaid: Sorry, I think the `force overwrite existed Brewfile` bit doesn't quite make sense so we could maybe tweak that to the previously suggested string too? Sorry, last thing and then will merge!

xu-cheng: Revert back.

lib/brewdler/utils.rb
andrew: I can't see me using this feature myself but I agree with @mikemcquaid that we should keep things less interactive to match with the homebrew cli style.

xu-cheng: I guess I need to reimplement it with `--dry-run`.

xu-cheng:
@andrew @mikemcquaid I have reimplemented it with `--dry-run` feature. Please review. :smile:

mikemcquaid:
LGTM :+1:

andrew:
Very nice :+1:

PR #52, opened 2015-03-13
set $SAFE when eval Brewfile

xu-cheng:


spec/security_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect { dsl.process }.to raise_error(SecurityError)
+
+    dsl = Brewdler::Dsl.new("Kernel.system 'whomai'")
+    expect { dsl.process }.to raise_error(SecurityError)
+
+    dsl = Brewdler::Dsl.new("IO.popen 'whomai'")
+    expect { dsl.process }.to raise_error(SecurityError)
+  end
+
+  it "rasies error if there is backtick" do
```
mikemcquaid: `raises`

mikemcquaid:
:+1:

spec/security_spec.rb
xu-cheng: Oops

PR #54, opened 2015-03-13
eliminate warning during the test

xu-cheng:
This commit fixes the following warning when invoking `bundle exec rake`

    Could not find rake-10.4.2 in any of the sources (Bundler::GemNotFound)


mikemcquaid:
:+1:

xu-cheng:
:ship: ed.

Issue #57, opened 2015-03-22
Parse brewfiles line by line


cisso:
Currently the DSL seems to choke on invalid quotes, causing the whole Brewfile to fail with "missing formula" errors for each single line. I would suggest to either
- split by newline and parse each line on its own, or
- validate the whole Brewfile and throw an error if any invalid characters have been encountered.

I'm not a ruby guy and I've only briefly looked at the code, but my guess would be that the first option is actually easier to implement.

mikemcquaid:
Could you provide a sample failing `Brewfile`? Thanks!

cisso:
Sure! I've provided multiple samples here: https://gist.github.com/cisso/ccadc42b4133d810d212
The one I mentioned above is Brewfile_fails_for_all_1.

xu-cheng:
This issue is related to [the legacy codes here](https://github.com/Homebrew/homebrew-brewdler/blob/master/lib/brewdler/commands/install.rb#L6-L11). Instead of raising the error when dealing with invalid `Brewfile`, it tries to treat each line as individual formula name. In the result, some weird messages get out.

Should we just drop these legacy codes? cc @andrew

andrew:
Yeah I'm not attached to the legacy code at all!

kevinSuttle:
So, to be clear, what was the reason for requiring quotes in formula names inside of Brewfiles?

mikemcquaid:
@kevinSuttle It's a Ruby DSL which means we don't need to implement our own parser.

kevinSuttle:
@mikemcquaid Ok. So, not wanting to derail here, but that does directly affect Homebrew formulae, especially Caskroom.

mikemcquaid:
@kevinSuttle I doesn't affect anything outside of `Brewfile`s.

PR #59, opened 2015-03-24
Code Refactor

xu-cheng:
1. Remove the legacy code (#57).
2. Call `process` inside `initialize` for Dsl.
3. Improve the error message.


xu-cheng:
cc @mikemcquaid @andrew

mikemcquaid:
If it works: :+1:

xu-cheng:
It works. :ship:ed

PR #60, opened 2015-03-29
allowing custom Brewfile path in the argument

xu-cheng:
It can be useful if there're multiple Brewfile configurations.
e.g.

    brew brewdle --file=Brewfile.osx
    brew brewdle --file=Brewfile.linux

### DIFF
--- a/cmd/brew-brewdle.rb
+++ b/cmd/brew-brewdle.rb
@@ -8,9 +8,9 @@ $LOAD_PATH.unshift(BREWDLER_LIB)
 require "brewdler"
 
 usage = <<-EOS.undent
-  brew brewdle [-v|--verbose] [--file=<path>]
-  brew brewdle dump [--force] [--file=<path>]
-  brew brewdle cleanup [--dry-run] [--file=<path>]
+  brew brewdle [-v|--verbose] [--file=<path>|--global]
+  brew brewdle dump [--force] [--file=<path>|--global]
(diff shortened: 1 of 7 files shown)


mikemcquaid:
Seems good to me. I almost wonder if we want to allow a "system" `Brewfile` somehow? That `brew cleanup-installed` command I've not deprecated has the advantage that you don't need to specify a full path to a `Brewfile` (but I don't think that's a big deal). What do you think?

xu-cheng:
I'm basically fine with a system `Brewfile`. Although I think it's a little risk to read a remote file, should we add an argument to enable it? Like `brew brewdle cleanup --system` or something. And use `brew brewdle dump --system` to dump into it?

Maybe ping @andrew for his opinion.

mikemcquaid:
Yeh `--system` makes sense.

xu-cheng:
Updated.

mikemcquaid:
Looks great! Sorry to contradict myself but I think I meant `--global` rather than `--system` to match what Git does: `--local` (which is implied), `--global` for your `$HOME`, `--system` for `/etc` and specifying files directly

xu-cheng:
No problem. Updated to use `--global`. Also, I switch to use `$HOME/.Brewfile` rather than `~/.Brewfile`.

mikemcquaid:
:+1: :shipit:

PR #62, opened 2015-04-07
resolve path when overwrite Brewfile

xu-cheng:
It's useful when running `brew brewdle dump --global --force` and ~/.Brewfile
is a symlink.


mikemcquaid:
Oooh, yes, I wanted this to because mine is a symlink too :sparkles:

PR #63, opened 2015-04-17
Rename to `brew bundle`.

mikemcquaid:
I've heard a few complaints about the name and I can never remember the actual command name(s). After renaming keep the old naming around as symlinked aliases so they can still be called.


andrew:
I'll be sad to see the name go but it makes sense for it to be more predictable :+1:

xu-cheng:
So we will change the repo name as well?

cmd/brew-bundle.rb
```diff
@@ -0,0 +1,53 @@
...
+  brew bundle cleanup   uninstall all Homebrew formulae not listed in Brewfile
+
+  Options:
+  -v, --verbose          print verbose output
+  --force                force overwrite existed Brewfile
+  --dry-run              list formulae rather than actual uninstalling them
+  --file=<path>          set Brewfile path
+  --global               set Brewfile path to $HOME/.Brewfile
+  -h, --help             show this help message and exit
+  --version              show the version of bundle
```
xu-cheng: Maybe want the right part to be aligned with `Usage` part.

xu-cheng:
:+1:

mikemcquaid:
Renamed repo and merged. Thanks for the understanding @andrew :heart:.

PR #64, opened 2015-04-20
cask dumper: chomp the trailing " (!)"

xu-cheng:
Cask would add ` (!)` into cask name if the cask is not availabled or
some temporary errors.


mikemcquaid:
:+1:

PR #65, opened 2015-04-21
Update logging to make use of past participles

timlo:


lib/bundle/dsl.rb
```diff
@@ -47,10 +47,10 @@ def install
                 Bundle::RepoInstaller
               end
         if cls.install(*arg)
-          puts "Succeed to #{verb} #{entry.name}"
+          puts "Succeeded to #{verb} #{entry.name}"
```
mikemcquaid: I think this should be `Succeeded in`

timlo: If we were to update to `Succeeded in` then it would read as `Succeeded in install ffmpeg`, whereas currently this change would result in `Succeeded to install ffmpeg`.

If the preference is to have `Succeeded in` I can change the `verb` to the -ing equivalent so that it reads: `Succeeded in installing ffmpeg` etc.

mikemcquaid: > If the preference is to have Succeeded in I can change the verb to the -ing equivalent so that it reads: Succeeded in installing ffmpeg etc.

Makes sense to me!

timlo:
@mikemcquaid made the updates as per your comments

mikemcquaid:
Looks good, thanks!

PR #66, opened 2015-04-22
Support multiple brews, casks or taps in DSL

paulfri:
Would like to simplify my Brewfile a bit.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -28,14 +28,15 @@ Create a `Brewfile` in the root of your project:
 
 Then list your Homebrew based dependencies in your `Brewfile`:
 
-    tap 'caskroom/cask'
+    tap 'caskroom/cask', 'homebrew/dupes'
     brew 'emacs', args: ['cocoa', 'srgb', 'with-gnutls']
     brew 'redis'
     brew 'mongodb'
     brew 'sphinx'
     brew 'imagemagick'
     brew 'mysql'
-    cask 'google-chrome'
+    brew 'tcl-tk', 'python', 'python3', args: ['universal', 'with-brewed-tk']
+    cask 'google-chrome', 'firefox'
 
 You can then easily install all of the dependencies with one of the following commands:
 
--- a/lib/bundle/dsl.rb
+++ b/lib/bundle/dsl.rb
@@ -64,19 +64,19 @@ module Bundle
         [args, {}]
       end
 
-      names.each do |name|
+      names.flatten.each do |name|
         @entries << Entry.new(:brew, name, options)
       end
     end
 
     def cask(*names)
-      names.each do |name|
+      names.flatten.each do |name|
         @entries << Entry.new(:cask, name)
       end
     end
 
     def tap(*names)
-      names.each do |name|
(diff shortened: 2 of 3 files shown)


lib/bundle/dsl.rb
```diff
@@ -57,16 +57,28 @@ def install
...
     end
 
-    def brew(name, options={})
-      @entries << Entry.new(:brew, name, options)
+    def brew(*args)
+      names, options = if args.last.is_a? Hash
+        [args[0...-1], args.last]
+      else
+        [args, {}]
+      end
```
xu-cheng: Maybe just
```ruby
options = args.last.is_a?(Hash) ? args.pop : {}
```

spec/dsl_spec.rb
```diff
@@ -7,4 +7,90 @@
...
+        Bundle::Dsl.new("tap 'phinze/cask', 'homebrew/versions'")
+      end
+    end
+
+    context 'when array of taps is passed' do
+      it 'adds entries for each tap' do
+        Bundle::Dsl.new('tap %w(phinze/cask homebrew/versions)')
+      end
+    end
+  end
```
xu-cheng: :+1: for test.

xu-cheng:
LGTM. Will let other maintainers have a look.

mikemcquaid:
> Would like to simplify my Brewfile a bit.

Can you give an example of what it looks like looks like before and after this change?

andrew:
Would be good to add some documentation to the readme as well.

paulfri:
I added a simple example to the readme. Let me know what you think.

Readme.md
```diff
@@ -28,14 +28,15 @@ Create a `Brewfile` in the root of your project:
...
-    tap 'caskroom/cask'
+    tap 'caskroom/cask', 'homebrew/dupes'
     brew 'emacs', args: ['cocoa', 'srgb', 'with-gnutls']
     brew 'redis'
     brew 'mongodb'
     brew 'sphinx'
     brew 'imagemagick'
     brew 'mysql'
-    cask 'google-chrome'
+    brew 'tcl-tk', 'python', 'python3', args: ['universal', 'with-brewed-tk']
```
mikemcquaid: Do these all get installed with the same args?

mikemcquaid:
I'm :-1: on this change, I think. I think it makes the DSL more confusing and ambigious and may limit us if we want to expand the DSL in future.

Readme.md
paulfri: Yes, that's the idea. Homebrew ignores unapplicable arguments so it should be safe in e.g. this case.

mikemcquaid: Homebrew may not always do this; I don't think we should rely on this.

paulfri:
@mikemcquaid thanks for your feedback!

If there's any suggestion for where you might want the DSL to go, I'll happily edit this PR, otherwise I'll deal with the duplication for the time being.

mikemcquaid:
@paulfri Thanks for understanding :heart:

PR #68, opened 2015-04-26
dsl: exit 1 when there are failures.

### LINKED ISSUE #67: Return non-zero when `$ brew bundle` fails
When running `brew bundle` now, if a package fails to install, the command still returns `0`. It would be awesome if the command returned `1` or some other non-zero value whenever it encounters an error.

For example, given the following Brewfile:

```
# Brewfile
brew 'node'
brew 'some-package-that-doesnt-exist'
```

```
$ brew bundle
Succeed to install node
Error: No available formula for some-package-that-doesnt-exist 
Searching formulae...
Searching taps...
Fail to install some-package-that-doesnt-exist

Success: 1 Fail: 1

$ echo $?
0
```

mikemcquaid:
Closes #67.

### DIFF
--- a/lib/bundle/commands/install.rb
+++ b/lib/bundle/commands/install.rb
@@ -1,7 +1,7 @@
 module Bundle::Commands
   class Install
     def self.run
-      Bundle::Dsl.new(Bundle.brewfile).install
+      Bundle::Dsl.new(Bundle.brewfile).install || exit(1)
     end
   end
 end
--- a/lib/bundle/dsl.rb
+++ b/lib/bundle/dsl.rb
@@ -55,6 +55,7 @@ module Bundle
         end
       end
       puts "\nSuccess: #{success} Fail: #{fail}"
+      fail == 0
     end
 
     def brew(name, options={})


lib/bundle/dsl.rb
```diff
@@ -55,6 +55,7 @@ def install
         end
       end
       puts "\nSuccess: #{success} Fail: #{fail}"
+      exit 1 if fail > 0
```
xu-cheng: Maybe return false and exit at `Bundle::Commands#Install::run`? Your call.

mikemcquaid: Cool.

mikemcquaid: Updated.

lib/bundle/commands/install.rb
```diff
@@ -1,7 +1,9 @@
 module Bundle::Commands
   class Install
     def self.run
-      Bundle::Dsl.new(Bundle.brewfile).install
+      success = Bundle::Dsl.new(Bundle.brewfile).install
+      exit 1 unless success
+      success
```
xu-cheng: How about: `exit 1 unless Bundle::Dsl.new(Bundle.brewfile).install`?

mikemcquaid: I never like having the actual action being part of the `if` or `unless` personally. What do you think?

xu-cheng: How about `Bundle::Dsl.new(Bundle.brewfile).install || exit 1`?

IMHO, it's pointless to return `success` here. As it will always be true.

Issue #37, opened 2015-01-31
Feature Request: Upgrade Packages


AlJohri:
It would be nice to have the ability to upgrade packages. This could either be automatically (if an upgrade exists for a package in the Brewfile, then upgrade it) or a separate dsl so its a tad more explicit (such as 'brew_upgrade').

This is inspired by thoughtbot's laptop setup script: https://github.com/thoughtbot/laptop/blob/master/mac#L56

```bash
brew_install_or_upgrade() {
  if brew_is_installed "$1"; then
    if brew_is_upgradable "$1"; then
      fancy_echo "Upgrading %s ..." "$1"
      brew upgrade "$@"
    else
      fancy_echo "Already using the latest version of %s. Skipping ..." "$1"
    fi
  else
    fancy_echo "Installing %s ..." "$1"
    brew install "$@"
  fi
}
```

I'd love to submit a PR for this feature, just interested to hear which method is more likely to be accepted.

jamesaanderson:
How about something like `brewdle update`?

AlJohri:
That sounds like a good solution. Would ```brewdle update``` install packages if they aren't already installed or just upgrade the existing packages?

I can see how ```brewdle update``` is more inline with Bundler but saying ```brewdle upgrade``` may be more inline with Homebrew. Not sure. Thoughts?

jamesaanderson:
I think it would be fine to install packages as well. As for the command, I'd alias `brewdle update` and `brewdle upgrade`

AlJohri:
Cool. I haven't written tests yet but does this look alright?

https://github.com/AlJohri/homebrew-brewdler/commit/5ab122279aeea3397ac510532ea67de86cb774a1

jamesaanderson:
Looks good though not sure why you added `"install"` to `dsl_spec.rb` and not feeling how you're passing the mode into `initialize`.

AlJohri:
I was iffy on that too. I needed the ```brew``` method in the dsl to change depending on the ```@mode```. I changed ```dsl_spec.rb``` to match my changes to ```initialize``` (gets 2 arguments). I'll try to think of something else - let me know if you have a better way.

jamesaanderson:
Regarding `dsl_spec.rb` pass in install or upgrade based on the command run but the syntax of the `Brewfile` shouldn't change.

AlJohri:
It doesn't change. The current solution is:

```ruby
[...]

PR #69, opened 2015-05-05
Install or upgrade brew by default

timurvafin:
This changes behavior so `brew bundle` will install or upgrade packages from `Brewfile` by default.

Related to https://github.com/Homebrew/homebrew-bundle/issues/37

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -41,6 +41,8 @@ You can then easily install all of the dependencies with one of the following co
 
     $ brew bundle
 
+If a dependency is already installed and there is an update available it will be upgraded.
+
 ### Dump
 
 You can create a `Brewfile` from all the existing Homebrew packages you have installed with:
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -8,7 +8,7 @@ $LOAD_PATH.unshift(BUNDLE_LIB)
 require "bundle"
 
 usage = <<-EOS.undent
-  brew bundle [-v|--verbose] [--file=<path>|--global]
+  brew bundle [-v|--verbose] [--file=<path>|--global] [--install-only]
   brew bundle dump [--force] [--file=<path>|--global]
   brew bundle cleanup [--dry-run] [--file=<path>|--global]
   brew bundle [--version]
@@ -17,7 +17,7 @@ usage = <<-EOS.undent
   Usage:
   Bundler for non-Ruby dependencies from Homebrew
 
-  brew bundle            read Brewfile and install all dependencies
+  brew bundle            read Brewfile and install and upgrade all dependencies
   brew bundle dump       write all currently installed packages into a Brewfile
   brew bundle cleanup    uninstall all Homebrew formulae not listed in Brewfile
 
@@ -27,6 +27,7 @@ usage = <<-EOS.undent
   --dry-run              list formulae rather than actual uninstalling them
   --file=<path>          set Brewfile path
   --global               set Brewfile path to $HOME/.Brewfile
+  --install-only         do not upgrade existing formulae, just install new
   -h, --help             show this help message and exit
   --version              show the version of bundle
 EOS
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -8,7 +8,7 @@ $LOAD_PATH.unshift(BUNDLE_LIB)
 require "bundle"
 
 usage = <<-EOS.undent
-  brew bundle [-v|--verbose] [--file=<path>|--global] [--install-only]
+  brew bundle [-v|--verbose] [--file=<path>|--global]
   brew bundle dump [--force] [--file=<path>|--global]
   brew bundle cleanup [--dry-run] [--file=<path>|--global]
   brew bundle [--version]
@@ -27,7 +27,6 @@ usage = <<-EOS.undent
   --dry-run              list formulae rather than actual uninstalling them
   --file=<path>          set Brewfile path
   --global               set Brewfile path to $HOME/.Brewfile
-  --install-only         do not upgrade existing formulae, just install new
   -h, --help             show this help message and exit
   --version              show the version of bundle
 EOS
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -1,18 +1,53 @@
 module Bundle
   class BrewInstaller
     def self.install(name, options = {})
-      if Bundle.brew_installed?
-        command = [name]
-        unless options[:args].nil?
-          options[:args].each do |arg|
-            command << "--#{arg}"
-          end
-        end
+      new(name, options).install_or_upgrade
+    end
 
-        Bundle.system "brew", "install", *command
+    def initialize(name, options = {})
+      if Bundle.brew_installed?
+        @name = name
+        @args = prepare_args(options[:args])
       else
         raise "Unable to install #{name}. Homebrew is not currently installed on your system"
       end
+    end
+
+    def install_or_upgrade
+      if installed?
+        upgrade!
+      else
+        install!
+      end
+    end
+
+    private
+
+    def installed?
+      Bundle.installed_formulae.include?(@name)
+    end
+
+    def upgradable?
+      Bundle.outdated_formulae.include?(@name) &&
+        ! Bundle.pinned_formulae.include?(@name)
+    end
+
+    def install!
+      Bundle.system("brew", "install", @name, *@args)
+    end
+
+    def upgrade!
+      return Bundle.system("brew", "upgrade", @name) if !install_only? && upgradable?
+      true
+    end
+
+    def install_only?
+      ARGV.include?("--install-only")
+    end
+
+    def prepare_args(args)
+      args ||= []
+      args.map { |arg| "--#{arg}" }
     end
   end
 end
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -53,12 +53,8 @@ module Bundle
     end
 
     def upgrade!
-      return Bundle.system("brew", "upgrade", @name) if !install_only? && upgradable?
+      return Bundle.system("brew", "upgrade", @name) if upgradable?
       true
-    end
-
-    def install_only?
-      ARGV.include?("--install-only")
     end
 
     def prepare_args(args)
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -23,17 +23,33 @@ module Bundle
 
     private
 
+    def installed_formulae
+      @@installed_formulae ||= `brew list -1`.split("\n")
+    end
+
+    def outdated_formulae
+      @@outdated_formulae ||= `brew outdated --quiet`.split("\n")
+    end
+
+    def pinned_formulae
+      @@pinned_formulae ||= `brew list --pinned`.split("\n")
+    end
+
     def installed?
-      Bundle.installed_formulae.include?(@name)
+      installed_formulae.include?(@name)
     end
 
     def upgradable?
-      Bundle.outdated_formulae.include?(@name) &&
-        ! Bundle.pinned_formulae.include?(@name)
+      outdated_formulae.include?(@name) &&
+        ! pinned_formulae.include?(@name)
     end
 
     def install!
-      Bundle.system("brew", "install", @name, *@args)
+      if (success = Bundle.system("brew", "install", @name, *@args))
+        installed_formulae << @name
+      end
+
+      success
     end
 
     def upgrade!
--- a/lib/bundle/utils.rb
+++ b/lib/bundle/utils.rb
@@ -24,18 +24,6 @@ module Bundle
     @@cask ||= Kernel.system("brew cask --version &>/dev/null")
   end
 
-  def self.installed_formulae
-    @@installed_formulae ||= `brew list -1`.split("\n")
-  end
-
-  def self.outdated_formulae
-    @@outdated_formulae ||= `brew outdated --quiet`.split("\n")
-  end
-
-  def self.pinned_formulae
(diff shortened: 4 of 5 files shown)


ruslankhaertdinov:
Looks good

mikemcquaid:
I'm on vacation until the 13th May and I'd like to hold off merging this until I'm back and get a chance to review this. Thanks!

xu-cheng:
Some comments:
* Let's try to run `brew list` and `brew outdated` command only once.
* Use `brew outdated` (which return all outdated formulae) rather than `brew outdated #{name}` as the latter one has shell injection vulnerability.
* Remove `puts ...` which is redundancy with the message in `DSL#install` 

Some questions:
* What will/should happen when installed formulae have different build arguments than upgrade one?
* Should we add a ARGV flag to enable/disable this feature?

timurvafin:
@xu-cheng 

1. will make changes related to `brew list`, and `brew outdated`
2. we should rewrite messaging basically in the `DSL#install`, I think it should be based on object returned by `#install` method. That object should respond to `success?` and `message` (or something) to be able determine what type of action was done (install, upgrade, error). But probably in the separate pull request.
3. OK
4. "What will/should happen when installed formulae have different build arguments than upgrade one?" - I don't know. Any recommendations?
4. `--no-upgrade` was suggested in https://github.com/Homebrew/homebrew-bundle/issues/37. Do you think it should be set by default?

xu-cheng:
I just did a check on the upgrade command. It turns out `brew upgrade` only accept the formula name. So you don't need to pass the options to it.

lib/bundle/brew_installer.rb
```diff
@@ -1,17 +1,56 @@
...
   class BrewInstaller
     def self.install(name, options = {})
+      new(name, options).install_or_upgrade
+    end
+
+    def initialize(name, options = {})
       if Bundle.brew_installed?
-        command = [name]
+        @name = name
+        @command = prepare_command(name, options)
```
xu-cheng: Instead of `@commands`, let's use `@args` like this:
```ruby
@args = (options[:args] || []).map { |arg| "--#{arg}" }
```

timurvafin:
@xu-cheng done in https://github.com/Homebrew/homebrew-bundle/commit/136cca6584b50df65a057217408b991c87e07513

xu-cheng:
LGTM, but let's wait @mikemcquaid's final review.

Readme.md
```diff
@@ -41,6 +41,8 @@ You can then easily install all of the dependencies with one of the following co
 
     $ brew bundle
 
+In case dependency already installed and there is update available it will be upgraded.
```
mikemcquaid: `If a dependency is already installed and there is an update available it will be upgraded.`

cmd/brew-bundle.rb
```diff
@@ -17,7 +17,7 @@
   Usage:
   Bundler for non-Ruby dependencies from Homebrew
 
-  brew bundle            read Brewfile and install all dependencies
+  brew bundle            read Brewfile and install or upgrade all dependencies
```
mikemcquaid: `and upgrade` I'd say

cmd/brew-bundle.rb
```diff
@@ -27,6 +27,7 @@
   --dry-run              list formulae rather than actual uninstalling them
   --file=<path>          set Brewfile path
   --global               set Brewfile path to $HOME/.Brewfile
+  --install-only         do not upgrade existing formulae, just install new
```
mikemcquaid: I'm not fussed on this option, personally.

mikemcquaid:
The big problem with here I'd like to see addressed is that it'll upgrade all formulae from taps to newer versions if they are available. This maybe requires another PR from @xu-cheng or someone to tweak how we install/dump things from taps. For example if you have `redis` version `2.6.9-boxen1` installed from  then homebrew-bundle should run `brew outdated boxen/brews/redis` and `brew upgrade boxen/brews/redis` rather than `brew upgrade redis`.

cmd/brew-bundle.rb
timurvafin: @mikemcquaid please suggest alternative

mikemcquaid: I think just removing it is fine.

timurvafin: @mikemcquaid 

I'm going to use it inside Rails `bin/setup`, when I want just to make sure I have required formulae installed for specific application. Upgrading is completely different thing and related to the whole system.

Just adding `install AND upgrade` will change current behavior that is bad for backward сompatibility. So I recommend to force `install-only` by default.

Readme.md
timurvafin: Updated in https://github.com/timurvafin/homebrew-bundle/commit/6a911be2e911cfa4bacadf6db8fcc5289f3f3db8

cmd/brew-bundle.rb
timurvafin: Updated in https://github.com/timurvafin/homebrew-bundle/commit/6a911be2e911cfa4bacadf6db8fcc5289f3f3db8

timurvafin:
@mikemcquaid any chance to get it merged?

mikemcquaid:
@timurvafin See my comment from yesterday.

timurvafin:
@mikemcquaid could you please prove more detailed example, please?
Something like that?
```bash
brew install boxen/brews/redis # will install redis 2.8.6-boxen1
brew outdated redis # => redis (2.8.6-boxen1 < 3.0.1)
```

mikemcquaid:
@timurvafin Yup, that's correct. I'll try to provide a more detailed example but I'm interested in @xu-cheng's thoughts first.

timurvafin:
@xu-cheng any thoughts?

xu-cheng:
I think instead of using full qualified name in `Brewfile`, we should solve it inside Homebrew core. I'm beginning to work on this problem as part of https://github.com/Homebrew/homebrew/pull/39524. In a nut shell, `brew outdated` should use `Keg` to load the formula.

mikemcquaid:
@xu-cheng Agreed :+1: 
@timurvafin This means we'll need to wait a little while before merging this. Sorry about that and thanks for your patience.

lib/bundle/brew_installer.rb
```diff
@@ -1,18 +1,52 @@
...
+    def upgradable?
+      Bundle.outdated_formulae.include?(@name)
+    end
+
+    def install!
+      Bundle.system("brew", "install", @name, *@args)
+    end
+
+    def upgrade!
+      return Bundle.system("brew", "upgrade", @name) if !install_only? && upgradable?
```
xu-cheng: Note: need to check pin as well. Otherwise `brew update name` will overwrite the formula pin.

timurvafin: @xu-cheng isn't it `brew upgrade` responsibility to take care of pinned formulae?

timurvafin: We could implement something like `brew outdated` -- `brew list --pinned` to get list of formulae which should be upgraded.

xu-cheng: No, it's designed to overwrite the pin if you call upgrade command specific to a formula.

timurvafin: @xu-cheng fixed in https://github.com/Homebrew/homebrew-bundle/commit/92f444c4f930b1bb8f755d93adbacfeb38d434af

spec/brewinstaller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+            expect(Bundle).to receive(:system).with("brew", "upgrade", "git").and_return(true)
+            expect(do_install).to eql(true)
+          end
+
+          context "when formula pinned" do
+            before do
+              allow(Bundle).to receive(:pinned_formulae).and_return([formula])
+            end
+
+            it "upgrade formula" do
```
xu-cheng: not

timurvafin: @xu-cheng fixed in https://github.com/Homebrew/homebrew-bundle/commit/ff8f84f6f39f9bd8495c57077b1f3bdb6600ba67

lib/bundle/utils.rb
```diff
@@ -24,6 +24,18 @@ def self.cask_installed?
...
+  def self.installed_formulae
+    @@installed_formulae ||= `brew list -1`.split("\n")
+  end
+
+  def self.outdated_formulae
+    @@outdated_formulae ||= `brew outdated --quiet`.split("\n")
+  end
+
+  def self.pinned_formulae
+    @@outdated_formulae ||= `brew list --pinned`.split("\n")
```
xu-cheng: `@@pinned_formulae`

timurvafin:
@xu-cheng done

timurvafin:
@mikemcquaid @xu-cheng anything else I can do to make this pull request merged?

mikemcquaid:
@timurvafin I want to wait until we get https://github.com/Homebrew/homebrew/pull/39524 merged first, sorry.

timurvafin:
@mikemcquaid sure, thanks!

lib/bundle/utils.rb
```diff
@@ -24,6 +24,18 @@ def self.cask_installed?
...
+    @@installed_formulae ||= `brew list -1`.split("\n")
+  end
+
+  def self.outdated_formulae
+    @@outdated_formulae ||= `brew outdated --quiet`.split("\n")
+  end
+
+  def self.pinned_formulae
+    @@pinned_formulae ||= `brew list --pinned`.split("\n")
+  end
```
xu-cheng: It seems that these methods wouldn't be used outside `BrewInstaller`. So maybe put them inside `BrewInstaller`?

timurvafin: @xu-cheng btw, as I understand external commands like `brew bundle` has full access to the Homebrew API https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/External-Commands.md#ruby-commands

Could we try to use Homebrew Ruby API instead of system calls?

xu-cheng: Yes, it has full access to Homebrew API. But as a design choice, we should avoid using it. Part of reason is that Homebrew internal Ruby API is private and may be changed in the future.

timurvafin: For example `Formula.installed` instead of `brew list -1`

timurvafin:
@xu-cheng done in https://github.com/timurvafin/homebrew-bundle/commit/7d035c648e156bb9922438b46fa873b579160c99

lib/bundle/utils.rb
xu-cheng: Let's avoid calling Homebrew ruby api if possible.

xu-cheng:
This can be merged now if @mikemcquaid has no other comments/objections.

cmd/brew-bundle.rb
mikemcquaid: Let's remove this option, thanks.

mikemcquaid:
@xu-cheng One last comment and then: if it works feel free to merge :+1:

timurvafin:
:shipit:

mikemcquaid:
@timurvafin Remove the `--install-only` option, thanks!

timurvafin:
@mikemcquaid how it looks right now?

mikemcquaid:
Looks good to me. Will just get a :+1: from @xu-cheng and we're good to go.

xu-cheng:
Sorry, but there still needs a few fix to handle the full name and short name issue.

lib/bundle/brew_installer.rb
```diff
@@ -1,18 +1,65 @@
...
+    end
+
+    private
+
+    def installed_formulae
+      @@installed_formulae ||= `brew list -1`.split("\n")
+    end
+
+    def outdated_formulae
+      @@outdated_formulae ||= `brew outdated --quiet`.split("\n")
```
xu-cheng: `brew outdated` now outputs the full name, so let's do this:
```ruby
@@outdated_formulae ||= `brew outdated --quiet`.split("\n").map { |f| f.split("/").last }
```

lib/bundle/brew_installer.rb
```diff
@@ -1,18 +1,65 @@
...
       if Bundle.brew_installed?
-        command = [name]
-        unless options[:args].nil?
-          options[:args].each do |arg|
-            command << "--#{arg}"
-          end
-        end
-
-        Bundle.system "brew", "install", *command
+        @name = name
```
xu-cheng: Since the argument can be full name (e.g. `homebrew/tap/foo`) or short name (e.g. `bar`). We can do this:

```ruby
@full_name = name
@name = name.split("/").last
```

lib/bundle/brew_installer.rb
```diff
@@ -1,18 +1,65 @@
...
+      installed_formulae.include?(@name)
+    end
+
+    def upgradable?
+      outdated_formulae.include?(@name) &&
+        ! pinned_formulae.include?(@name)
+    end
+
+    def install!
+      if (success = Bundle.system("brew", "install", @name, *@args))
```
xu-cheng: `@name` -> `@full_name`

lib/bundle/brew_installer.rb
```diff
@@ -1,18 +1,65 @@
...
+        installed_formulae << @name
+      end
+
+      success
+    end
+
+    def upgrade!
+      return Bundle.system("brew", "upgrade", @name) if upgradable?
```
xu-cheng: `@name` -> `@full_name`

xu-cheng:
Merged in 12201b17c00270cf28d47bfc073e6cb7c8c9799d with a slight tweak. Thanks @timurvafin for your awesome work.

timurvafin:
@xu-cheng I was on vacation and was not able to apply you changes. Thank you!

mikemcquaid:
Thanks for all your work both of you!

PR #71, opened 2015-05-21
Skip installed cask

### LINKED ISSUE #70: Skip already installed packages
So I just created a Brewfile with a bunch of casks. If I modify the Brewfile and run `brew bundle` again, it doesn't appear to skip the casks that have already been installed. It's sitting there on each line for a while with "curl" up in the Terminal header.

xu-cheng:
Closes #70.


mikemcquaid:
Wow, so fast. I'd not even got around to adding it to me todo list! :+1: from me, code looks good.

PR #73, opened 2015-05-31
Dump full tap paths 

### LINKED ISSUE #72: Dump full tap paths
If I've installed a formula as `brew install homebrew/versions/gcc48` then it should be dumped the same way in the `Brewfile` e.g. `brew 'homebrew/versions/gcc48` rather than the current version which is just `gcc48`. Additionally we could use this to simplify only dumping `tap`s that don't belong to one of these formulae (if necessary).

Adding this so I remember to do it at some stage but if anyone wants to beat me to it: feel free!

xu-cheng:
Closes #72

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -16,10 +16,10 @@ module Bundle
     def to_s
       @formulae.map do |f|
         if f[:args].empty?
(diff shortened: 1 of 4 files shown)


mikemcquaid:
Great, thanks!

PR #76, opened 2015-06-05
bug_fix: full_name instead of name for brew bundle dump

jcrben:
When I tried 'brew bundle dump', I was just getting empty strings for all of the regular brew packages. This seems to fix it. I did not trigger the else condition but outputting the file showed that full_name was nil.

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -16,10 +16,10 @@ module Bundle
     def to_s
       @formulae.map do |f|
         if f[:args].empty?
-          "brew '#{f[:full_name]}'"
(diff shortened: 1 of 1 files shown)


mikemcquaid:
CC @xu-cheng. I'm guessing maybe we need to check that the Homebrew repo is at a given revision?

xu-cheng:
Yes, it looks like that @jcrben need to run brew update.  And we need to add homebrew minimum revision check.

jcrben:
Ah, I see. I updated and now it works fine. It's been a few weeks and that brew was tasting stale. You're welcome to close this if that helps.

PR #77, opened 2015-06-06
check homebrew revision against minimum requirement

xu-cheng:
cc @mikemcquaid

### DIFF
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -1,5 +1,12 @@
 #!/usr/bin/env ruby
 
+MINIMUM_HOMEBREW_REVISION = "bcee474bb4a989fec0fc0d4f9fbfb993208829e8"
+HOMEBREW_REPOSITORY.cd do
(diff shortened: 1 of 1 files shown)


xu-cheng:
Comments?

mikemcquaid:
Seems good :+1:

mikemcquaid:
This doesn't work with shallow clones of Homebrew, unfortunately.

mikemcquaid:
(which our installer does by default now, I believe)

xu-cheng:
You are right. I reverted this PR for now. I will consider the alternative solution. IMO, we may consider bump homebrew's version whenever its CLI API makes any change.

mikemcquaid:
Could maybe go by the date of the latest commit in `origin/master` instead of the commit hash?

xu-cheng:
I dunno. But aren't the date of commits in non-chronicle order? I mean sometimes we may push the commits authored a long time ago.

mikemcquaid:
The authorship date may be old but the commit date will not. e.g. check `git log --pretty=fuller` and the `CommitDate`.

PR #78, opened 2015-06-06
simplify file write logic

xu-cheng:


lib/bundle/dumper.rb
```diff
@@ -24,15 +24,11 @@ def dump_brewfile
...
     def write_file(file, content, overwrite=false)
-      if file.exist?
-        if overwrite && file.file?
-          file = file.realpath
-          FileUtils.rm file
-        else
-          raise "#{file} already existed."
-        end
+      if file.exist? && !overwrite
+        raise "#{file} already existed."
```
mikemcquaid: `already exists`

mikemcquaid:
:+1:

PR #79, opened 2015-06-06
Dump: sorted by dependant order

### LINKED ISSUE #74: Dump should handle dependencies order.
`brew bundle dump` should handle entry order based on dependencies. The dependencies of a formula should come first.  At the same time, requirements from cask like `java`, `mactex` should come first as well. But some casks should come later for those depend on formula.

xu-cheng:
Fixes #74.

Requires https://github.com/Homebrew/homebrew/pull/40451 merged first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ rvm:
   - 2.0.0
   - 2.2.0
 cache: bundler
+before_install:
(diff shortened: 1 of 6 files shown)


mikemcquaid:
Clever :+1:

PR #80, opened 2015-06-07
support anytap

xu-cheng:


mikemcquaid:
Nice. Would be great if `dump` was supported for this too at some point but :+1: on merging this as-is.

xu-cheng:
> Would be great if dump was supported

I'm thinking to add a `brew tap-info` command (similar to `brew info`) to the core to achieve that. Even more, I'm considering to add `Tap` class to the core so we can put all tap related code in a centralized  place.

PR #81, opened 2015-06-09
check min homebrew commit date

xu-cheng:


mikemcquaid:
:+1:

mikemcquaid:
Check it works with a shallow clone (if you haven't already)

xu-cheng:
> Check it works with a shallow clone (if you haven't already)

Right I almost forget this. Just checked, it works like a charm.

Issue #93, opened 2015-07-14
Feature Request: cask args?


joshuairl:
Trying to add the --appdir=/Applications arg to cask installs?

Maybe there could just be some sort of global setting at the top of the Brewfile and then it wouldn't need to be added to each of the casks?

Just an idea?

joshuairl:
I'm working on a pull request for this now.
Adding tests and once it passes I'll send the pull request.

PR #94, opened 2015-07-14
Added cask args for support of appdir=/Applications and such.

joshuairl:


joshuairl:
#93

mikemcquaid:
Do you personally install some stuff into `/Applications` and some into `~/Applications` or all in the same place? I'm wondering if it might make sense to allow this to be a global setting.

joshuairl:
No, just figured if cask has options may as well make the system support them so if people do use those args of cask they can.

Just my thoughts!

joshuairl:
I'm sorry, I may have misunderstood you.
You mean make the appdir directory a global setting in the Brewfile?
or something fixed inside of the codebase?

joshuairl:
Let me give it a go.  I have some ideas for defaults.

mikemcquaid:
So, I personally just want either all my applications (or all in a `Brewfile`) in `/Applications` or in `~/Applications` so I thought it might make more sense to have a global DSL value e.g. `cask_args` or something which can be applied either to individual casks or to all casks.

joshuairl:
Yep, that's precisely what I've built now.  I'm not the best as specs so I can't figure out exactly how to implement a test for it but here is the code I came up with... one sec.

joshuairl:
Alright, I added a "cask_defaults" option.

Also, since cask options are pre-conceived and we know what they all are I've changed it to be a standard hash instead of being inside of an "args" hash under options.

So new casks can be defined as this:

```ruby
cask_defaults appdir: "/Applications"

cask "google-chrome", appdir: "~/Applications" #gets assigned the appdir ~/Applications
cask "firefox" #gets assigned the appdir from "cask_defaults"
```

xu-cheng:
I would prefer to support the DSL like this:
 ```ruby
cask "google-chrome", args: [appdir: "~/Applications"]
````
So it would benefit us if we want to extend DSL further in the future.

Also I would like to remove `PERMITTED_OPTIONS` in the implementation. So we won't need to update this because any changes made in homebrew-cask.

**Comment by joshuairl** [comment]
The cask options do not change often at all.
They are simply two different animals.

Making the software better is the ultimate goal is it not?

I generally prefer to create more testable code rather than leaving it open for the sake of low maintenance reasons.

That's my vote at least.

**Comment by mikemcquaid** [comment]
> cask "google-chrome", args: {appdir: "~/Applications"}

I'd also like to see it use nested hashes like this.

File: lib/bundle/cask_installer.rb [hunk 1]
```diff
@@ -1,6 +1,8 @@
 module Bundle
   class CaskInstaller
-    def self.install(name)
+    PERMITTED_OPTIONS = %w{caskroom appdir prefpanedir qlplugindir fontdir binarydir input_methoddir screen_saverdir}
```
mikemcquaid: As @xu-cheng said: please remove these.

lib/bundle/cask_installer.rb
```diff
@@ -18,7 +20,18 @@ def self.install(name)
         return true
       end
 
-      if (success = Bundle.system "brew", "cask", "install", name)
+      if options.size
```
mikemcquaid: `options.any?` or `!options.empty?`

lib/bundle/dsl.rb
```diff
@@ -15,11 +15,8 @@ def initialize(type, name, options={})
     def initialize(input)
       @input = input
       @entries = []
-      begin
-        process
-      rescue
-        raise "Invalid Brewfile."
-      end
+      @caskdefaults = {}
```
mikemcquaid: `@cask_opts`

lib/bundle/dsl.rb
```diff
@@ -55,12 +53,28 @@ def install
       fail == 0
     end
 
+    def cask_defaults(args)
```
mikemcquaid: `cask_opts`

Readme.md
```diff
@@ -37,6 +37,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'imagemagick'
     brew 'mysql'
     cask 'google-chrome'
+    cask 'firefox', args: ['appdir=/Applications']
```
mikemcquaid: Can you add a `cask_opts` example too? Thanks!

joshuairl:
What's your reasoning for nesting it like this.  That's not how bundler does it? 
Is there a benefit?
Ruby handles putting all args in a hash anyway ?

Just curious

lib/bundle/cask_installer.rb
joshuairl: Cask options aren't the same as args passed to the build script of the installing brew...
Why would we treat cask the same?

joshuairl:
To elaborate, there is absolutely no reason we would need to extend the CASK part of the DSL.
Adding a new option to cask if it changes is apart of maintaining the project like this.
Keeping static arguments helps create repeatable tests.

The `--with-` arguments are just build parameters to the libraries installed by brew.
Cask options are a completely different animal.  Apples to Oranges.

Cask options are testable where as brew module build arguments aren't so it makes sense to leave them open or isolated into an "args:" hash but requiring cask arguments to be "key=value" is kind of an odd implementation and is not a standard practice.  If we truly care about building a nice user friendly and testable tool I vote we allow cask specific arguments to be more of a hash rather than this generic passing of Strings on to a command...?

joshuairl:
Also, I'm okay with removing the PERMITTED_OPTIONS section and validation... but I'm definitely not okay with the "key=value" syntax for cask args.  It's silly for us to only append the "--" to the string for the end-user.  Might as well just have them add the "--" in there too... otherwise I just don't see the point in barely adding a feature... why not go the extra mile and make it more usable and friendly.  :-/

joshuairl:
I mean it is a "dsl" after all... It's like adding a combo-box versus a text box... usability increases quite a bit with the combo-box and it's ability to be queries and tested... I dunno I guess I'm just missing the point of taking easy street for the sake of future maintenance of adding an option or two later when cask changes.

mikemcquaid:
> To elaborate, there is absolutely no reason we would need to extend the CASK part of the DSL.

Well, there is a reason: to avoid future clashes between the Brewfile DSL and Homebrew Cask's arguments.

> Adding a new option to cask if it changes is part of maintaining a project like this.

I don't agree, sorry. It's unnecessary work.

> but requiring cask arguments to be "key=value" is kind of an odd implementation and is not a standard practice
>> cask "google-chrome", args: {appdir: "~/Applications"}

Taking an array instead may be a better idea:

```ruby
cask "google-chrome", args: [{appdir: "~/Applications"}] 
```

xu-cheng:
> Taking an array instead may be a better idea

We can omit the braces.
```ruby
$ pry
[1] pry(main)> [appdir: "~/Applications"]
=> [{:appdir=>"~/Applications"}]
[2] pry(main)> ["a-flag", appdir: "some place", binarydir: "another place"]
=> ["a-flag", {:appdir=>"some place", :binarydir=>"another place"}]
```

mikemcquaid:
> We can omit the braces.

:+1:

joshuairl:
Okay but who are we building this for?  
It's not more work to loop over args passed directly to the cask method.  It's the same process but makes it cleaner for the end user.  It just plain reads better from the Brewfile to not add an extra args option and an array for that matter for absolutely no reason.

We aren't hard coding the options... We are doing the same process but my way makes it cleaner to understand and read when viewing a brewfile.

Adding everything to an array reads horribly.

mikemcquaid:
> It just plain reads better from the Brewfile to not add an extra args option and an array for that matter for absolutely no reason.

Two of the maintainers disagree.

joshuairl:
Okay, is there an actual reason for your disagreement or…? :)
I’m not trying to be difficult… I just get kind of passionate about these things. 

Forgive me for the persistence.

mikemcquaid:
We've explained our preferences above.

joshuairl:
`["a-flag", {:appdir=>"some place", :binarydir=>"another place"}]`
whaaaaat?
How is this a better idea?!
How does it solve anything?

I'm sorry boys, we may just be cut from different fabric.
I'm completely missing the entire purpose of over complicating the cask install options.

Cask only works one way, it's not a very complicated project.
It's absolute non-sense to make it out to be something it's not.

You explained it, yes, but it didn't have any substance.
Why would you extend the CASK portion of the DSL in the future... other than to accompany some sort of rare new feature of the Cask project?

I want to see your side of it, I'm trying... I guess your explanations above aren't enough for me.
I get WHY you are throwing things into a big array... but you don't have to... you're doing the same exact thing by flattening it out and not putting it in array... you're still just concatenating strings and prepending dashes it just removes the "code" from the brew file... why add extra syntax and code to something like a configuration file when you can accomplish the same loop and concatenation with the args passed to cask in a more clean syntax.

mikemcquaid:
We appreciate your PR and are happy to include it if it's modified to meet our requests. If you're not prepared to meet our requests please close it. I can't speak for @xu-cheng but I don't really have time for a long argument about this.

joshuairl:
Okay ignore my comments about adding a new option when the cask project changes.
My final question on this and I swear I'll drop it is...

How does nesting things under "args:" or placing them in an array under args help you maintain the project better or help there to be less clashes with the cask project if Ruby puts them in a hash that is looped over the same as an array and you aren't actually defining the options in the cask method anywhere?  I think we are both accomplishing the same thing for maintenance purposes but without the exhaustive extra step of adding args: [ "arg1":"dklfjw"].
It's the same as just doing cask "name", arg1: "dklfjw"
As far as the cask method is concerned... it's just as much code to loop over args in that syntax as it is to add them to an array.

I mean the word "cask" is technically an arg... why would we not put that in args too?  Why do we choose to draw a line at anything except for a "name" of a brew...

I don't have much time either.  And I want to contribute to the project and make it better.  That's why I'm concerned about decisions that relate to where the project is going.  This isn't a dictatorship is it? lol...
We're all on the same level here... I'm just sharing my concerns.  I'm sure all of us just want to build good software.

Think I'll just leave this alone for awhile.
Thanks for all you do.

joshuairl:
Went ahead and made some of the changes to reflect what I've been trying to say.
I simply removed the things you were asking to remove and it works the same except without an extra step of typing array syntax into the Brewfile...

mikemcquaid:
> How does nesting things under "args:" or placing them in an array under args help you maintain the project better

If we decide to add global arguments that affect all DSL elements (e.g. both `brew` and `cask`) then we don't want to have to deal with clashes with current or future Cask arguments.

Readme.md
```diff
@@ -28,6 +28,7 @@ Create a `Brewfile` in the root of your project:
 
 Then list your Homebrew based dependencies in your `Brewfile`:
 
+    cask_opts appdir: "/Applications"
```
mikemcquaid: Use single-quotes here for consistency.

Readme.md
```diff
@@ -37,6 +38,8 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'imagemagick'
     brew 'mysql'
     cask 'google-chrome'
+    cask 'firefox', appdir: "~/Applications" # override individual cask options per cask
+    cask "filezilla" # installs using the "cask_opts" defaults at the top.
```
mikemcquaid: Use single-quotes here for consistency.

Readme.md
```diff
@@ -37,6 +38,8 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'imagemagick'
     brew 'mysql'
     cask 'google-chrome'
+    cask 'firefox', appdir: "~/Applications" # override individual cask options per cask
```
mikemcquaid: Worth checking if this gets expanded if you use `~` here.

Readme.md
mikemcquaid: Actually, I think you can just kill this line.

Readme.md
mikemcquaid: Also you can kill the comment; think it's self-explanatory.

joshuairl:
Okay, that makes more sense to me.
Thanks for clarifying.
I don’t mind making the revision to move toward a nested version of this.

You guys rock! Again, I never meant to be difficult! :)
<3

spec/caskinstaller_spec.rb
```diff
@@ -45,7 +49,7 @@ def do_install
 
     context "when cask is installed" do
       before do
-         allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return(["google-chrome"])
+         allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return(["google-chrome","firefox"])
```
mikemcquaid: Total pedantry but a space after the `,` would be great

joshuairl:
Oh and also, forgive me I'm very new to Ruby so I don't always know to use certain features of the language to achieve things.  I'm learning so thank you for clarifying things for me!

mikemcquaid:
No worries, I appreciate the contribution :heart:

Readme.md
joshuairl: It does expand it, I confirmed it.

Readme.md
```diff
@@ -37,6 +39,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'imagemagick'
     brew 'mysql'
     cask 'google-chrome'
+    cask 'firefox', opts: {appdir: '~/Applications'}
```
mikemcquaid: Let's use `args` for consistency.

Readme.md
```diff
@@ -28,6 +28,8 @@ Create a `Brewfile` in the root of your project:
 
 Then list your Homebrew based dependencies in your `Brewfile`:
 
+    cask_opts appdir: '/Applications'
```
mikemcquaid: Sorry, I know I've changed my mind here but `cask_args` feels like it may be more consistent.

mikemcquaid:
Looks great, a few final nitpicks and the merge conflict needs resolved. Thanks for all your work here :heart:

mikemcquaid:
Hey @joshuairl! I've got a bit more bandwidth to help now but I am up for including this but it'll need rebased first. Thanks!

lib/bundle/cask_installer.rb
```diff
@@ -18,7 +18,14 @@ def self.install(name)
         return true
       end
 
-      if (success = Bundle.system "brew", "cask", "install", name)
+      if options.any?
```
mikemcquaid: You can probably just loop through without this check.

lib/bundle/cask_installer.rb
```diff
@@ -18,7 +18,14 @@ def self.install(name)
...
+        @args = []
+        options.each do |key, array|
+          @args << "--#{key}=#{options[key]}"
+        end
+      end
+
+      if (success = Bundle.system("brew", "cask", "install", name, *@args))
```
mikemcquaid: This will fail if `@args` is `nil`. Removing the `.any` check shoudl fix that.

mikemcquaid: Although might be nicer to use `args` instead of `@args` too.

spec/caskinstaller_spec.rb
```diff
@@ -23,7 +25,9 @@ def do_install
         allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return([])
         expect(Bundle).to receive(:system).with("brew", "install", "caskroom/cask/brew-cask").and_return(true)
         expect(Bundle).to receive(:system).with("brew", "cask", "install", "google-chrome").and_return(true)
+        expect(Bundle).to receive(:system).with("brew", "cask", "install", "firefox", "--appdir=/Applications").and_return(true)
         expect(do_install).to eq(true)
+        expect(do_install_with_args).to eq(true)
```
mikemcquaid: Please put this call into a new test rather than adding to the existing one, thanks.

Readme.md
```diff
@@ -28,6 +28,8 @@ Create a `Brewfile` in the root of your project:
 
 Then list your Homebrew based dependencies in your `Brewfile`:
 
+    cask_args appdir: '/Applications'
+
```
mikemcquaid: Kill this empty line, thanks.

kylegordy:
:+1: If you all can get this merged. I'm really looking forward to this feature.

agh:
Ditto. This looks great. :+1:

alexhernandez:
So whats the final decision on syntax?

mikemcquaid:
If anyone else wants to try and make a new PR and merge these conflicts that'd be great.

PR #96, opened 2015-07-15
Fix #88, Add DSL for handling services

### LINKED ISSUE #88: Add DSL for handling services
When `nginx` is installed it has a `launchd` `plist` that it includes. This isn't (yet) copied or started but it'd be nice if it could be.

Perhaps it'd be nice to have an API like:
```ruby
brew 'nginx', restart_service: 'true'
```
that will run tap https://github.com/Homebrew/homebrew-services and then run `brew services restart nginx` to setup and start `nginx`.

voanhduy1512:


xu-cheng:
LGTM, but if I understand correctly, some formulae require `sudo` to control services.

voanhduy1512:
How about change the dsl to 

```ruby
brew 'nginx', services: { restart: true, sudo: true }
```

Which make it cleaner and easier to extend.

xu-cheng:
New DSL looks good. cc @mikemcquaid for comments as you are the one who requests such feature.

lib/bundle/brew_installer.rb
```diff
@@ -9,11 +9,18 @@ def initialize(name, options = {})
...
         @name = name.split("/").last
         @args = options.fetch(:args, []).map { |arg| "--#{arg}" }
+        @restart = options.fetch(:restart_service, false)
       else
         raise "Unable to install #{name}. Homebrew is not currently installed on your system"
       end
     end
 
+    def run
+      install_or_upgrade && restart
```
mikemcquaid: I'd add the (optional) `restart` into the `install_or_upgrade` method

lib/bundle/brew_installer.rb
```diff
@@ -22,7 +29,10 @@ def install_or_upgrade
       end
     end
 
-    private
+    def restart
+      return BrewServices.restart(@name) if @restart
+      true
```
mikemcquaid:
```
return true unless @restart
BrewServices.restart(@name)`
```

lib/bundle/brew_services.rb
```diff
@@ -0,0 +1,26 @@
+module Bundle
+  class BrewServices
+    def self.restart(name)
+      new(name).restart
+    end
+
+    def initialize(name)
+      ensure_brew_services_installed
+      @name = name
+    end
```
mikemcquaid: Make this the first method.

mikemcquaid:
A few comments (that @xu-cheng may disagree with me on) but otherwise :+1:

mikemcquaid:
Oh, could you also add an example to the `README`?

lib/bundle/brew_services.rb
voanhduy1512: Sorry, what do you mean the first method?

mikemcquaid: Sorry, the first method in the class (e.g. before `restart`); just a typical convention.

lib/bundle/brew_installer.rb
voanhduy1512: How about `sudo` option, I think we need to change to DSL the support that. And in case the DSL change like i suggest, i don't think add optional params into  `install_or_upgrade` is the right way to do it

mikemcquaid: I don't think we'll want to have `sudo` support in this DSL and I don't think this method needs to change name to add optional support for restarting a service. Sorry!

xu-cheng:
@voanhduy1512 Shouldn't you update the DSL to `brew 'nginx', services: { restart: true, sudo: true }`?

voanhduy1512:
I don't change the dsl because @mikemcquaid https://github.com/Homebrew/homebrew-bundle/pull/96#discussion_r34972600

lib/bundle/brew_installer.rb
voanhduy1512: I fix other comment but can you explain more why you will add the (optional) `restart` into `install_or_upgrade`. Because currently i fetch it from the init of `BrewInstaller`. 
Moreover, i think use a boolean parameter in a method to control logic flow should be avoided.

mikemcquaid: Instead of a boolean you can `options = {}` and `options[:restart] = true`

xu-cheng:
@mikemcquaid Any reason you don't want to support sudo? In my understanding, certain services(e.g. web services listening to resolved ports) require sudo to work. Without it, it would be useless for them,

mikemcquaid:
@xu-cheng Most things don't require it and we need to work out the permissions; would this require running `brew bundle` as root or would it ask for the `sudo` password or what? Feels a bit icky security-wise no matter what we do.

lib/bundle/brew_installer.rb
voanhduy1512: Sorry i still don't get it. Can you make a quick pseudo what `self.install` and `install_or_upgrade` will be what you add a parameter like that. 
I think it will be like this https://gist.github.com/voanhduy1512/e6993637ca204f630539 but i feel it isn't as clean as current implement

xu-cheng:
> Most things don't require it and we need to work out the permissions

There are 154 formulae which have services. Among that, 24 of them require `sudo`.

> would this require running brew bundle as root or would it ask for the sudo password or what?

It would ask user for user password.

> Feels a bit icky security-wise no matter what we do.

This is a problem we have to address. Maybe we should solve it in `homebrew/services`'s end. (Create a sudo-helper or something).

mikemcquaid:
I think this is a big and difficult enough problem that we should skip it for now. We can't just call out to `sudo` as it's a really bad habit to encourage users to do (as after that we can run whatever command we want with root privileges). As you pointed out, the vast majority of services do not require `sudo` and it doesn't feel like there should be a duplication between the `sudo` state in the Brewfile and Formula anyway.

voanhduy1512:
@mikemcquaid @xu-cheng do you need anything else or it's good to merge?

lib/bundle/brew_installer.rb
mikemcquaid: I've added my suggested changed here: https://gist.github.com/mikemcquaid/a8c08cfe5eec3416ed22/revisions

voanhduy1512:
All done :dancer:

mikemcquaid:
LGTM now :+1: @xu-cheng thoughts? If you're happy: :shipit:

lib/bundle/brew_installer.rb
```diff
@@ -2,6 +2,7 @@ module Bundle
   class BrewInstaller
     def self.install(name, options = {})
       new(name, options).install_or_upgrade
+      BrewServices.restart(name) if options[:restart_service]
```
xu-cheng: This way you fail to process the exit state from `install_or_upgrade`.

voanhduy1512:
```

Comment (voanhduy1512) [inline lib/bundle/brew_installer.rb]:
```ruby
  if new(name, options).install_or_upgrade  && options[:restart_service]
     BrewServices.restart(name)
  else
     false
  end
end
```
It seem a bit ugly this way

mikemcquaid: The `if` can be postfix and you don't need the `else` (as you want to return it unconditionally)

voanhduy1512: i don't think we can do that, because if we change if to postfix and remove `else`. What will happened if `result` is `true` but `restart_service` is `false`, the method will return `false`, which should be `true`

mikemcquaid:
```ruby
  result = new(name, options).install_or_upgrade
  result = BrewServices.restart(name) if result && options[:restart_service]
  result
end
```

xu-cheng:
:ship: it after test turns green.

voanhduy1512:
:shipit:

xu-cheng:
Merged, thanks for your contribution @voanhduy1512

voanhduy1512:
:hand: you are welcome :smile:

mikemcquaid:
Thanks @voanhduy1512!

Issue #100, opened 2015-08-17
Generate less output when not actually installing anything


technicalpickles:
I'm looking to add a `Brewfile` to a project, and run `brew bundle` when [bootstrapping it](http://githubengineering.com/scripts-to-rule-them-all/).

If I have a pretty simple `Brewfile` like:

```ruby
brew 'nagios'
```

And I run `brew bundle` multiple times, I always get this output, even when nagios is already installed:

```
$ brew bundle
Succeeded in installing nagios

Success: 1 Fail: 0
```

I could imagine there being a flag like `--quiet` which only outputs if something was actually installed. Or, maybe having a `--check` option which is used for scripting, so you can do something like this: https://github.com/github/scripts-to-rule-them-all/blob/master/script/bootstrap#L10-L13

josh:
:+1: 

> [Rule of Silence: When a program has nothing surprising to say, it should say nothing.](http://www.linfo.org/rule_of_silence.html)

mikemcquaid:
:+1:

voanhduy1512:
@mikemcquaid do you need a hand :happy:

mikemcquaid:
@voanhduy1512 That would be great!

PR #101, opened 2015-08-24
Cleanup and tweaks

### LINKED ISSUE #91: Make `brew bundle cleanup` safer
I wonder if it makes sense to make `brew bundle cleanup` not do anything without either e.g. `--dry-run`, `--global` or `--force`. A coworker was just trying out `brew bundle` and managed to blow away almost all his installed formulae by mistake. I've personally never used `cleanup` without `--global` and I can see why it would be dangerous.

mikemcquaid:
- General cleanups
- dumper: allow `-` to read from `STDIN`
- cleanup: use `--force` instead of `--dry-run`
- cleanup: also uninstall casks.
- Add check command
- Readme: add example of using `system`.
- gitignore: ignore `bin` directory.
- 100% test coverage.

CC @xu-cheng @josh @technicalpickles

Closes #91 and #100.


mikemcquaid:
@xu-cheng Whoops, yep, I did (although for consistency I might make it write to `STDOUT`). Any other objections or thoughts?

lib/bundle/commands/check.rb
```diff
@@ -0,0 +1,37 @@
...
+      current_casks = Bundle::CaskDumper.new.casks
+      requested_casks - current_casks
+    end
+
+    def self.formulae_to_install
+      @@dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      requested_formulae = @@dsl.entries.select { |e| e.type == :brew }.map(&:name)
+      current_formulae = Bundle::BrewDumper.new.formulae.map { |f| f[:full_name] }
+      upgradable_formulae = Bundle::BrewInstaller.upgradable_formulae
+      requested_formulae - current_formulae - upgradable_formulae
```
xu-cheng: You may want to turn all of formulae into short name(as opposite to full name).

xu-cheng:
> although for consistency I might make it write to STDOUT

no objection. If you do this, you need bypass file existing check as `/dev/stdout` will always exist.

lib/bundle/commands/check.rb
mikemcquaid: Any idea what format `upgradable_formulae` or `Dsl` will give you?

xu-cheng: `upgradable_formulae` will give you short names. `DSL` can return both decided by users' Brewfile.

xu-cheng: Also why do we need `upgradable_formulae` here? If it's needed, shouldn't it be `requested_formulae - (current_formulae - upgradable_formulae)`?

mikemcquaid: I'll play around.

xu-cheng:
Another the thing come to my mind, would `-y` or `--yes` be better than `--force` for cleanup?

mikemcquaid:
@xu-cheng I think that implies there's a prompt without. `--force` feels more appropriate given the destructiveness too.

xu-cheng:
:cool: it's just `--force` feels like something you would try after normal invoking failed.

mikemcquaid:
@xu-cheng Addressed feedback. Let me know what you think.

lib/bundle/commands/check.rb
```diff
@@ -0,0 +1,37 @@
...
+    def self.casks_to_install
+      @@dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      requested_casks = @@dsl.entries.select { |e| e.type == :cask }.map(&:name)
+      current_casks = Bundle::CaskDumper.new.casks
+      requested_casks - current_casks
+    end
+
+    def self.formulae_to_install
+      @@dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      requested_formulae = @@dsl.entries.select { |e| e.type == :brew }.map(&:name)
```
xu-cheng: You can turn this into short name by `.map { |f| f.split("/").last }`

mikemcquaid: Done! Let me know what you think.

lib/bundle/dumper.rb
```diff
@@ -8,14 +8,17 @@ class Dumper
     def initialize
       @brew = BrewDumper.new
       @cask = CaskDumper.new
-      @repo = RepoDumper.new
+      @repo = TapDumper.new
```
xu-cheng: Small nit `repo` -> `tap`. Everything else are good :+1:

Issue #1, opened 2011-07-05
Awesome!  Any plans for other *nix compatibility?


bglusman:
We had plans for a project a little like this but the idea was to make it cross platform so if you develop on Mac and deploy to Linux, dependencies would install and resolve properly in both places...  currently of course Homebrew is Mac only, but I saw someone has an experimental Solaris port of homebrew (https://github.com/rmyers/homebrew)... Curious if this is a personal itch you'd like to scratch as well, and if so, I know some people (myself included) who'd like to help if we can!

adamliter:
@andrew I just tried running `brew bundle` on a Linux machine (Ubuntu 14.04), and I get the error:

    /usr/bin/env: ruby -W0: No such file or directory

The odd thing is that explicitly typing `/usr/bin/env ruby -W0` at the command line works. I'm honestly not really sure why the shebang isn't working, and I don't know what to do in order to debug this, especially since it works when manually typing the shebang.

mikemcquaid:
@adamliter I'm afraid I don't know what to suggest here beyond trying to submit a pull-request.

adamliter:
@mikemcquaid @andrew 
It seems that the problem is that the system is trying to call the executable "ruby -W0". See this [Stack Overflow question](http://stackoverflow.com/q/24032641/2571049) and its [answer](http://stackoverflow.com/a/24032743/2571049).

That being said, I have no idea why this works on a Mac but not Linux.

According to this [Ruby forum post](https://www.ruby-forum.com/topic/96121#199230), the best practice for shebangs with arguments in Ruby is to do:

```sh
#!/bin/sh
exec ruby -x "$0" "$@"
#!ruby -W0
# Ruby code goes here ...
```
If this is something you're open to changing, I'm happy to submit a pull request.

I tested it on both a Mac (10.10.5) and Ubuntu (14.04), and this seems to work in both cases.

mikemcquaid:
@adamliter Yeh, feel free to submit a PR. We could probably just try `#!/usr/bin/env ruby`.

adamliter:
@mikemcquaid I've created the pull request using the solution I suggested. I haven't tested it, but your suggested solution of just using `#!/usr/bin/env ruby` should also work. If you'd rather do that, I can change the pull request, or you can just close the pull request and make the changes yourself, if that's quicker. Thanks! :smile:

Nakilon:
[...]

PR #102, opened 2015-08-27
Fix shebang so brew bundle works on Linux

adamliter:
See the discussion in https://github.com/Homebrew/homebrew-bundle/issues/1.

This pull request fixes the shebang in `/cmd/brew-bundle` so that it can run on Linux. In Linux, the `#!/usr/bin/env` shebang previously tried to run the executable "ruby -W0", which returned the error:

    /usr/bin/env: ruby -W0: No such file or directory

(See, for example, [Bash trace mode (bash -x) in the shebang](http://stackoverflow.com/q/24032641/2571049) on Stack Overflow.)

By changing this to `#!/bin/sh` and making the call to ruby with `exec ruby -x "$0" "$@"`, this problem is solved.

`man ruby` tells us for the flag `-x` that:

>      -x[directory]  Tells Ruby that the script is embedded in a message.  Leading garbage will be discarded until the first line that starts with ``#!'' and con-
                    tains the string, ``ruby''.  Any meaningful switches on that line will be applied.  The end of the script must be specified with either EOF,
                    ^D (control-D), ^Z (control-Z), or the reserved word __END__.  If the directory name is specified, Ruby will switch to that directory before
                    executing script.

This is why `/cmd/brew-bundle` now has `__END__` at the end of the file.

I've tested this on both Mac (10.10.5) and Ubuntu (14.04).

### DIFF
--- a/cmd/brew-bundle
+++ b/cmd/brew-bundle
@@ -1,4 +1,7 @@
-#!/usr/bin/env ruby -W0
+#!/usr/bin/env ruby
+
+# Set verbosity level to 0
+$VERBOSE = nil
 
 # Ruby version check
 unless RUBY_VERSION.split(".").first.to_i >= 2
--- a/Readme.md
+++ b/Readme.md
@@ -8,7 +8,9 @@ Bundler for non-Ruby dependencies from Homebrew
 
 ## Requirements
 
-[Homebrew](http://github.com/Homebrew/homebrew) is used for installing the dependencies, it only works on OS X and so does this tool.
+[Homebrew](http://github.com/Homebrew/homebrew) or [Linuxbrew](https://github.com/homebrew/linuxbrew) are used for installing the dependencies.
+Linuxbrew is a fork of Homebrew for Linux, while Homebrew only works on Mac OS X.
+This tool is primarily developed for use with Homebrew on Mac OS X but should work with Linuxbrew on Linux, too.
 
 [brew tap](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/brew-tap.md) is new feature in Homebrew 0.9, adds more GitHub repos to the list of available formulae.
 
--- a/lib/bundle/utils.rb
+++ b/lib/bundle/utils.rb
@@ -17,15 +17,15 @@ module Bundle
   end
 
   def self.brew_installed?
-    @@brew ||= Kernel.system("brew --version &>/dev/null")
+    @@brew ||= Kernel.system("brew --version >/dev/null 2>&1")
   end
 
   def self.cask_installed?
-    @@cask ||= Kernel.system("brew cask --version &>/dev/null")
+    @@cask ||= Kernel.system("brew cask --version >/dev/null 2>&1")
   end
 
   def self.services_installed?
-    @@services ||= Kernel.system("brew services --help &>/dev/null")
+    @@services ||= Kernel.system("brew services --help >/dev/null 2>&1")
   end
 
   def self.brewfile


cmd/brew-bundle
```diff
@@ -1,4 +1,6 @@
-#!/usr/bin/env ruby -W0
+#!/bin/sh
```
mikemcquaid: Let's just try `/usr/bin/env ruby` first, thanks.

adamliter:
@mikemcquaid Okay, I've changed the commit to just do `#!/usr/bin/env ruby` as the shebang, and I've tested it on Ubuntu 14.04. It works.

mikemcquaid:
@xu-cheng Any objections here?

xu-cheng:
Maybe add `$VERBOSE = nil` which is the equivalence of `-W0`

mikemcquaid:
:+1:. Mind giving that a go @adamliter? Also, can you update the `README` to remove the stuff about this not working on Linux?

adamliter:
@mikemcquaid Yes, I can do that, but it'll probably have to wait a day or two. I'm currently on vacation and in a place with very sparse internet connection. :stuck_out_tongue: Thanks!

mikemcquaid:
No rush!

adamliter:
@mikemcquaid @xu-cheng I made the suggested changes, but now I'm getting:

```sh
Error: Unknown command: cask
Error: Unknown command: cask
0.9.5
```

When trying to run any `brew bundle` command on Linux. I'm not sure why this is showing up now. Let me see if I can figure out where it is coming from. I definitely wouldn't merge this pull request yet, though.

mikemcquaid:
@adamliter thanks!

dunn:
For some reason the verbosity, in particular `Kernel.exit` is leading to this final item on `brew bundle dump`s:
```
[...]
cask 'Kernel.exit'

**Comment by adamliter** [comment]
@dunn Hmm, that's interesting. Any idea why?

I haven't had much of a chance to do any debugging yet, but I just switched back to what was in the original pull request:

```sh
#!/bin/sh
exec ruby -x "$0" "$@"
#!ruby -W0
...
__END__
```

And I'm getting the same messages:

```sh
Error: Unknown command: cask
Error: Unknown command: cask
0.9.5
```

I'm not sure if something has changed or if I just missed that in my original testing (no idea how I would have?).

I'll try both things again with a fresh install of Linuxbrew and Homebrew-bundle sometime later.

It almost seems as if the output from `brew.installed?` and `cask.installed?` in `utils.rb` is getting through, but that wouldn't make sense since it is piped to `/dev/null`. And it also wouldn't explain the different behavior from Mac to Linux.

I'll try to do some closer looking tomorrow, but this might be beyond me to figure out. I'm not really a programmer. :stuck_out_tongue:

mikemcquaid:
Thanks for digging in here @adamliter. We'll make a programmer out of you yet :wink:

dunn:
Still investigating, but dropping these findings in case someone else connects the dots faster:
- `Bundle.cask_installed?` returns true even if Cask isn't installed.
- internal calls to `brew cask` fail even if Cask _is_ installed.

dunn:
@adamliter was on the right track.  Maybe we were using a non-standard redirection idiom, since this fixes it:
```diff
diff --git a/lib/bundle/utils.rb b/lib/bundle/utils.rb
index 8c12fa9..5e6b6a8 100644
--- a/lib/bundle/utils.rb
+++ b/lib/bundle/utils.rb
@@ -17,15 +17,15 @@ module Bundle
   end

   def self.brew_installed?
-    @@brew ||= Kernel.system("brew --version &>/dev/null")
+    @@brew ||= Kernel.system("brew --version >/dev/null 2>&1")
   end

   def self.cask_installed?
-    @@cask ||= Kernel.system("brew cask --version &>/dev/null")
+    @@cask ||= Kernel.system("brew cask --version >/dev/null 2>&1")
   end

   def self.services_installed?
-    @@services ||= Kernel.system("brew services --help &>/dev/null")
+    @@services ||= Kernel.system("brew services --help >/dev/null 2>&1")
   end

   def self.brewfile
```

adamliter:
@dunn Hmm, I wonder why the non-standard redirect idiom worked on Mac but not Linux. Interesting.

But anyway, should I add your changes to the pull request?

mikemcquaid:
It will probably depend on the shell it's calling out to. No harm in adding it.

adamliter:
With the fixes to the redirects proposed by @dunn, it seems like everything is working. I tested all of the `brew bundle` commands on a Linux VM (Ubuntu 14.04).

dunn:
Awesome! Most of the credit goes to you for narrowing in on the problem.

DomT4:
`&>` isn't supported by POSIX `sh`, so if the call is sent to `/bin/sh` rather than `/bin/bash` it can fail.

adamliter:
@DomT4 Thanks for the clarification! So the fix that @dunn suggested and that I added to this pull request would be the better way to go, then?

DomT4:
Yeah, you'll hit less potential issues using the `>/dev/null 2>&1` syntax, which should work well for dash/bash/zsh/fish/sh.

DomT4:
You may have hit this issue on Linux in particular because the Linux distros started to favour `sh` calling `dash` a while ago, and `echo "dog" &>/dev/null` doesn't work as intended with `dash`.

adamliter:
@DomT4 The odd thing is that your suggested test string doesn't print any output on my Linux VM (Ubuntu 14.04), whereas I *was* seeing the gratuitous output from `brew bundle` prior to the fix from @dunn.

DomT4:
So
```bash
sh -c "echo dog &>/dev/null"
```
Doesn't throw anything to stdout on Ubuntu? That's interesting. What does
```bash
sh --version
```
Throw, out of curiosity?

adamliter:
@DomT4 Ah, so `sh -c "echo dog &>/dev/null"` does print "dog" to stdout.

`sh --version` returns `sh: 0: Illegal option --`.

DomT4:
> does print "dog" to stdout.

:+1:. I presume `/bin/sh -c "echo dog >/dev/null 2>&1"` doesn't?

> returns sh: 0: Illegal option --.

Heh, forgot `dash` doesn't support that, or version checks in general. Sorry about that!

adamliter:
> :+1:. I presume sh -c "echo dog >/dev/null 2>&1" doesn't?

Correct. :smile_cat:

DomT4:
Cool. Unless anyone objects, I think adopting Alex's syntax is the path forwards. It's friendlier towards a wider variety of shells than `&>`.

PR #106, opened 2015-09-25
Expand the path of the `--file` argument

gabebw:
`brew bundle --file=~/Brewfile` didn't work because the `~` was not expanded.

Now we expand the whole path so that people can pass more idiomatic paths to
`brew bundle.`

About tests: I couldn't find any tests for the functions in `utils.rb`. I don't know what this project's policy on tests for the functions in that file is (maybe I missed something!) but if you'd like, I can add unit tests for at least `Bundle.brewfile`.

### DIFF
--- a/lib/bundle/utils.rb
+++ b/lib/bundle/utils.rb
@@ -33,8 +33,13 @@ module Bundle
       file = "#{ENV["HOME"]}/.Brewfile"
     else
       file = ARGV.value("file")
-      file = "/dev/stdin" if file == "-"
-      file ||= "Brewfile"
+      if file == "-"
+        file = "/dev/stdin"
+      elsif file.nil?
+        file = "Brewfile"
+      else
+        file = File.expand_path(file)
+      end
     end
     File.read(file)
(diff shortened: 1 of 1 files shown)


lib/bundle/utils.rb
```diff
@@ -33,8 +33,13 @@ def self.brewfile
...
-      file ||= "Brewfile"
+      if file == "-"
+        file = "/dev/stdin"
+      elsif file.nil?
+        file = "Brewfile"
+      else
+        file = File.expand_path(file)
+      end
     end
     File.read(file)
```
xu-cheng: Can you do this instead?

```ruby
if ARGV.include?("--global")
  file = Pathname.new("#{ENV["HOME"]}/.Brewfile")
else
  filename = ARGV.value("file")
  filename = "/dev/stdin" if filename == "-"
  filename ||= "Brewfile"
  file = Pathname.new(filename).expand_path(Dir.pwd)
end
file.read
```

Thanks.

gabebw:
Thanks!

PR #108, opened 2015-09-28
Don't use reserved keywords

teoljungberg:


mikemcquaid:
There's merge conflicts to be resolved here.

mikemcquaid:
Otherwise: :+1:

teoljungberg:
@mikemcquaid This seems to already be fixed on master by `Pathname.new(filename).expand_path(Dir.pwd)`. Which means that if `filename` is a relative path, it will be expanded to `Dir.pwd`.

So the only relevant change is to rename the local variables to not use ruby keywords.

Do you want me to change this PR to reflect that, or close it entirely?

mikemcquaid:
@teoljungberg Change the PR to reflect that, thanks.

teoljungberg:
@mikemcquaid Rebased and updated

mikemcquaid:

PR #113, opened 2015-10-09
some cleanup

xu-cheng:


lib/bundle/brew_dumper.rb
```diff
@@ -69,18 +73,18 @@ def formula_inspector f
...
-        args: args,
-        version: keg["version"],
-        dependencies: f["dependencies"],
-        requirements: f["requirements"],
+        :name => f["name"],
+        :full_name => f["full_name"],
+        :args => args,
+        :version => keg["version"],
+        :dependencies => f["dependencies"],
+        :requirements => f["requirements"],
```
mikemcquaid: Why, out of interest? I prefer the existing style.

mikemcquaid:
Could we get a Rubocop to handle this automatically in future?

lib/bundle/brew_dumper.rb
xu-cheng: This is changed by Rubocop(Using `brew style --fix <path to homebrew/bundle>`). Maybe we can change the rubocop rule if you want.

mikemcquaid: Ok, cool. I guess it's better to be consistent for now.

PR #115, opened 2015-10-26
Rebased/fixed: Added cask args for support of appdir=/Applications and such.

Liquidsoul:
Hi!

Here is my attempt to rebase and fix #94 on `master`.
I wanted to get it out as soon as possible to have some feedback so it may still need some work :smile: 

Hope this helps!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -41,7 +41,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'mysql'
     cask 'google-chrome'
     cask 'java' unless system '/usr/libexec/java_home --failfast'
-    cask 'firefox', args: {'appdir=/Applications'}
+    cask 'firefox', args: { appdir: '/Applications' }
 
 You can then easily install all of the dependencies with one of the following commands:
 
--- a/Readme.md
+++ b/Readme.md
@@ -31,7 +31,6 @@ Create a `Brewfile` in the root of your project:
 Then list your Homebrew based dependencies in your `Brewfile`:
 
     cask_args appdir: '/Applications'
-
     tap 'caskroom/cask'
     tap 'telemachus/brew', 'https://telemachus@bitbucket.org/telemachus/brew.git'
     brew 'emacs', args: ['with-cocoa', 'with-gnutls']
--- a/Readme.md
+++ b/Readme.md
@@ -30,6 +30,8 @@ Create a `Brewfile` in the root of your project:
 
 Then list your Homebrew based dependencies in your `Brewfile`:
 
+    cask_args appdir: '/Applications'
+
     tap 'caskroom/cask'
     tap 'telemachus/brew', 'https://telemachus@bitbucket.org/telemachus/brew.git'
     brew 'emacs', args: ['with-cocoa', 'with-gnutls']
@@ -40,6 +42,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'mysql'
     cask 'google-chrome'
     cask 'java' unless system '/usr/libexec/java_home --failfast'
+    cask 'firefox', args: ['appdir=/Applications']
 
 You can then easily install all of the dependencies with one of the following commands:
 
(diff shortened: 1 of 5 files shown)


mikemcquaid:
Looks good to me if it does to @xu-cheng :+1:

xu-cheng:
:+1: although should we squash commits?

lib/bundle/dsl.rb
```diff
@@ -63,12 +66,21 @@ def install
...
     end
 
-    def cask(name)
-      @entries << Entry.new(:cask, name)
+    def cask(name, options={})
+      if options[:args]
+        options = @cask_args.merge(options[:args])
+      else
+        options = @cask_args
+      end
```
xu-cheng: You should update `options[:args]` here not `options`. In fact, you can replace if condition to:
```ruby
options[:args] = @cask_args.merge options.fetch(:args, {})

File: Readme.md [hunk 2]
```diff
@@ -40,6 +41,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'mysql'
     cask 'google-chrome'
     cask 'java' unless system '/usr/libexec/java_home --failfast'
+    cask 'firefox', args: ['appdir=/Applications']
```
xu-cheng: `args: { appdir: '/Applications' }`

Liquidsoul:
I wanted to squash the commits from the previous PR but I preferred to sent the full branch before doing this.
Do you want me to squash them and update the PR?

Readme.md
```diff
@@ -40,6 +41,7 @@ Then list your Homebrew based dependencies in your `Brewfile`:
     brew 'mysql'
     cask 'google-chrome'
     cask 'java' unless system '/usr/libexec/java_home --failfast'
+    cask 'firefox', args: {'appdir=/Applications'}
```
xu-cheng: Sorry but you still fail to get it right. It should be: `{ appdir: '/Applications' }`

spec/caskinstaller_spec.rb
```diff
@@ -57,7 +70,7 @@ def do_install
 
     context "when cask is installed" do
       before do
-        allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return(["google-chrome"])
+        allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return(["google-chrome","firefox"])
```
xu-cheng: Why change this?

spec/caskinstaller_spec.rb
```diff
@@ -37,6 +41,15 @@ def do_install
...
       end
+
+      it "works fine with arguments" do
+        allow(Bundle).to receive(:cask_installed?).and_return(false, true)
+        allow(ARGV).to receive(:verbose?).and_return(false)
+        allow(Bundle::CaskInstaller).to receive(:installed_casks).and_return([])
+        expect(Bundle).to receive(:system).with("brew", "install", "caskroom/cask/brew-cask").and_return(true)
+        expect(Bundle).to receive(:system).with("brew", "cask", "install", "firefox", "--appdir=/Applications").and_return(true)
+        expect(do_install_with_args).to eq(true)
+      end
```
xu-cheng: This test may be more suitable to put after [Line 90](https://github.com/Liquidsoul/homebrew-bundle/blob/master/spec/caskinstaller_spec.rb#L90)

spec/caskinstaller_spec.rb
Liquidsoul: I don't really know. This comes from #94.
It does not seem relevant to me. I will remove this.

Readme.md
Liquidsoul: Ok I'll fix the Readme.
I am stil quite new to ruby and this code base. So I don't know if [this code](https://github.com/Liquidsoul/homebrew-bundle/blob/master/spec/caskinstaller_spec.rb#L9) tests this correctly.

mikemcquaid:
Would be good to squash down the commits to e.g. 1 per author.

Liquidsoul:
I've squashed the commits of the previous PR and add the fixes following your comments.

PR #116, opened 2015-10-27
Support aliases.

mikemcquaid:
Also some other general cleanup and refactoring and ensuring that tests have 100% coverage.

CC @xu-cheng

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -7,12 +7,7 @@ module Bundle
 
     def initialize
       if Bundle.brew_installed?
-        formulae_info = begin
-          JSON.load(`brew info --json=v1 --installed`) || []
-        rescue
-          []
-        end
-        @formulae = formulae_info.map { |info| formula_inspector info }
+        @formulae = BrewDumper.formulae_info
       else
         raise "Unable to list installed formulae. Homebrew is not currently installed on your system."
       end
@@ -61,7 +56,20 @@ module Bundle
 
     private
 
-    def formula_inspector(f)
+    def self.formulae_info
+      @@formulae_info ||= begin
+        installed_formulae = JSON.load(`brew info --json=v1 --installed`) || []
+        installed_formulae.map { |info| formula_inspector info }
+      rescue JSON::ParserError
+        []
+      end
+    end
+
+    def self.formulae_info_reset!
+      @@formulae_info = nil
+    end
+
+    def self.formula_inspector(f)
       installed = f["installed"]
       if f["linked_keg"].nil?
         keg = installed[-1]
@@ -75,6 +83,7 @@ module Bundle
       {
         :name => f["name"],
         :full_name => f["full_name"],
+        :aliases => f["aliases"],
         :args => args,
         :version => keg["version"],
         :dependencies => f["dependencies"],
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -24,14 +24,50 @@ module Bundle
       end
     end
 
-    def self.upgradable_formulae
-      outdated_formulae - pinned_formulae
+    def self.formula_installed_and_up_to_date?(formula)
+      formula_installed?(formula) && !formula_upgradable?(formula)
+    end
+
+    def self.formula_in_array?(formula, array)
+      array.include?(formula) || array.include?(resolved_formula_name(formula))
     end
 
     private
 
+    def self.formula_installed?(formula)
+      formula_in_array?(formula, installed_formulae)
+    end
+
+    def self.formula_upgradable?(formula)
+      formula_in_array?(formula, upgradable_formulae)
+    end
+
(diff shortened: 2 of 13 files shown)


lib/bundle/brew_dumper.rb
```diff
@@ -61,6 +56,18 @@ def expand_cask_requirements
 
     private
 
+    def self.formulae_info
+      @@formulae_info ||= begin
```
xu-cheng: I wonder do we need to cache this? Considering, it will only be run once for all of `brew bundle` subcommands. Beside by removing the cache, we can remove `formulae_info_reset!` at the same time.

If you want to keep it, you want to change `@@formulae_info` to `@formulae_info`

lib/bundle/brew_installer.rb
```diff
@@ -24,14 +24,43 @@ def install_or_upgrade
...
+      installed_formulae.include?(resolved_formula_name(formula))
+    end
+
+    def self.formula_upgradable?(formula)
+      upgradable_formulae.include?(resolved_formula_name(formula))
+    end
+
     def self.installed_formulae
-      @@installed_formulae ||= `brew list -1`.split("\n")
+      @@installed_formulae ||= Bundle::BrewDumper.new.formulae.map { |f| f[:name] }
```
xu-cheng: `Bundle::BrewDumper.new.formulae` this is probably what we want to cache to avoid duplicated formula topology sort.

lib/bundle/brew_dumper.rb
```diff
@@ -75,6 +82,7 @@ def formula_inspector(f)
       {
         :name => f["name"],
         :full_name => f["full_name"],
+        :aliases => f["aliases"],
```
xu-cheng: Beside the alias, we need to store `oldname` as well, which however may complicate implementation.

xu-cheng:
You need to update `Cleanup::formulae_to_uninstall` to support alias.

Another problem is that only using name to check formula is not enough. We have to handle the case when full name is used in `Brewfile` like `foo/bar/baz` or `foo/bar/baz-alias` but current installation is from another tap or core.

lib/bundle/brew_dumper.rb
mikemcquaid: I don't think we do need to (or should).

lib/bundle/brew_dumper.rb
mikemcquaid: It's a class-level cache so it can be reused across instances.

mikemcquaid:
> You need to update Cleanup::formulae_to_uninstall to support alias.

:+1: 



> Another problem is that only using name to check formula is not enough. We have to handle the case when full name is used in Brewfile like foo/bar/baz or foo/bar/baz-alias but current installation is from another tap or core.

Will test that case.

lib/bundle/brew_dumper.rb
xu-cheng: Actually I mean it make more sense to cache the result after it is sorted based on dependency order. 

In the core, we prefer

lib/bundle/brew_dumper.rb
xu-cheng: It can cause some problem. For example, `objective-caml` is not an alias but an old name for `ocaml`. 

Maybe we can simplify this. Instead of supporting old name. We create alias explicitly in such case, since it will never be used by another formula.

lib/bundle/brew_dumper.rb
mikemcquaid: We're using class variables elsewhere so I thought it was better to just be consistent here. Caching after sorted makes sense though :+1:

lib/bundle/brew_dumper.rb
mikemcquaid: I don't think we need to handle that case for now unless it's specifically requested.

mikemcquaid:
It's :green_heart:. @xu-cheng let me know what you think.

lib/bundle/brew_installer.rb
mikemcquaid: Not sure I see the problem here.

mikemcquaid: Can you elaborate (or can this be done in a follow-up PR)?

xu-cheng: It's just a better performance here. I can do it in a follow-up PR.

xu-cheng:
> You need to update Cleanup::formulae_to_uninstall to support alias.

You seem to miss that :wink: 

> I don't think we need to handle that case for now unless it's specifically requested.

We still need to somehow support old name. Otherwise, `check` and `cleanup` will be broken to users whenever we rename formulae.

mikemcquaid:
> You seem to miss that

Whoops, yep, I'll get it.

> We still need to somehow support old name. Otherwise, check and cleanup will be broken to users whenever we rename formulae.

That's fine. I don't want to expose `old_name` as a public API. In those cases users will need to rename. What do you propose we do.

PR #119, opened 2015-11-02
Use singleton

xu-cheng:
* Use singleton(`Class.method`) instead of object(`Class.new.method`).
* Handle case when using alias to refer tap formula, i.e. `foo/bar/alias -> foo/bar/baz` or `alias -> foo/bar/baz`
* Improve test case.
* Sanitize input name in DSL.
* Update minimum Homebrew version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ test/version_tmp
 tmp
 .ruby-gemset
 .ruby-version
+.envrc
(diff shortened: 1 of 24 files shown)


xu-cheng:
cc @mikemcquaid

mikemcquaid:
👍

PR #124, opened 2015-11-13
Uninstall to handle version conflicts

josh:
The high level problem I'd love to see `brew bundle` solve is working with version'd taps. A project may depend on a specific version of mysql. At the moment, homebrews main method of installing older versions of a formula is to use the `homebrew-versions` tap. However, you can only have one version of each of these installed at the same time. If you just add `homebrew/versions/mysql56` to your Brewfile, you'll run into conflicts if the user already has `mysql` installed.

While its still not ideal, it'd be great if `brew bundle` could resolve this conflict by removing any conflicting versions and installing the specific version required to use the project.

``` ruby
brew "mysql", uninstall: true, stop_service: true
brew "homebrew/versions/mysql56", restart_service: true
```

The current implementation attached is pretty terrible. I'm all for rewriting it or doing this completely differently. Just wanted to a discussion going with some sort of proposed approach.

To: @mikemcquaid

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -7,8 +7,13 @@ module Bundle
     end
 
     def self.install(name, options = {})
-      result = new(name, options).install_or_upgrade
-      result = BrewServices.restart(name) if result && options[:restart_service]
+      if options[:uninstall]
+        result = new(name, options).uninstall!
+        result = BrewServices.stop(name) if result && options[:stop_service]
+      else
+        result = new(name, options).install_or_upgrade
+        result = BrewServices.restart(name) if result && options[:restart_service]
+      end
       result
     end
 
@@ -28,6 +33,12 @@ module Bundle
       else
         install!
       end
+    end
+
+    def uninstall!
+      puts "Uninstalling #{@name} formula." if ARGV.verbose?
+      success = Bundle.system("brew", "uninstall", @full_name, *@args)
+      success
     end
 
     def self.formula_installed_and_up_to_date?(formula)
--- a/lib/bundle/brew_services.rb
+++ b/lib/bundle/brew_services.rb
@@ -5,6 +5,11 @@ module Bundle
       Bundle.system "brew", "services", "restart", name
     end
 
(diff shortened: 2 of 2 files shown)


mikemcquaid:
How about if this took an array like: 

`brew "homebrew/versions/mysql56", uninstall: ["mysql", "mysql55"], restart_service: true`

josh:
> How about if this took an array like:

Yeah, that could be interesting. I dig the idea of there being a clear relationship between "uninstall mysql, cause i'm trying to install mysql56".

xu-cheng:
To me, `uninstall` seems to be too dramatic behavior. Maybe use `unlink` instead? As for mysql case here, there is a catch that database may be corrupted if a old version mysql tries to read/write a database created by new version one.

IMHO, it would be better to do pre-install conflict check and warn users the situation. Instead of `uninstall` or `unlink` silently.

mikemcquaid:
@xu-cheng I don't like telling users what commands to run when we know. I don't think it should be silent but I don't think for an automation tool like `brew bundle` that we should tell users what to do when we know what commands to run ourselves.

xu-cheng:
> when we know what commands to run ourselves.

My point is we don't know what to do in this situation. In the case of mysql, we could potentially corrupt the existed database.

mikemcquaid:
> My point is we don't know what to do in this situation. In the case of mysql, we could potentially corrupt the existed database.

Corrupt, no. Refuse to start, perhaps depending on the state.

I'm open to other ideas but we definitely need a way to handle the move to versions within `brew bundle` itself or otherwise make `brew bundle` do a bunch of pinning/installing old versions by default. The issue this is fixing is: we upgrade to MySQL 5.7 in Homebrew core, a gem that relies on 5.6 breaks for everyone running `brew bundle`. Short of adding decent versions support to Homebrew core (a bigger project that I'm thinking we need to tackle) this seems like the best solution I've come across.

CC @DomT4 for thoughts here too.

DomT4:
Supplied thoughts in the chat this morning, but let me know if you want me to elaborate or comment on anything else here.

PR #86, opened 2015-06-16
drop support for ruby 1.x

xu-cheng:
Closes #85.

PR #126, opened 2015-11-20
Rename command to `brew-bundler.rb`.

mikemcquaid:
This means it'll use the Homebrew Ruby rather than whatever Ruby is in your path. This also means if you're using it to e.g. install some project dependencies to install a Ruby that does not already exist then
you do not need to do e.g. `RBENV_VERSION=system` first.

If we want to make it easier for people to run e.g. on 10.9 with an old Ruby (although I think it might be worth waiting for user requests first) we can loosen the `HOMEBREW_DEVELOPER` check in https://github.com/Homebrew/homebrew/pull/45816 to not require `HOMEBREW_DEVELOPER` to be set.

CC @xu-cheng for thoughts.

### DIFF
--- a/cmd/brew-bundle
+++ b/cmd/brew-bundle
@@ -1,98 +1,25 @@
-#!/usr/bin/env ruby
-
-# Set verbosity level to 0
-$VERBOSE = nil
-
-# Ruby version check
-unless RUBY_VERSION.split(".").first.to_i >= 2
-  abort "Ruby 2.0 or above is required. You can install it with `brew install ruby`."
-end
-
-# Load Homebrew Library
-require "pathname"
-HOMEBREW_LIBRARY_PATH = Pathname.new(ENV["HOMEBREW_LIBRARY_PATH"])
-$LOAD_PATH.unshift HOMEBREW_LIBRARY_PATH
-require "global"
-
-# Homebrew version check
-# commit f3cf43acfc764acba84fa20b4c42a3e0b5382589
-MIN_HOMEBREW_COMMIT_DATE = Time.parse "Wed Oct 14 17:26:41 2015 +0800"
-HOMEBREW_REPOSITORY.cd do
-  if MIN_HOMEBREW_COMMIT_DATE > Time.at(`git show -s --format=%ct`.to_i)
-    odie "Your Homebrew is outdated. Please run `brew update`."
-  end
-end
-
-BUNDLE_ROOT = File.expand_path "#{File.dirname(__FILE__)}/.."
-BUNDLE_LIB = Pathname.new(BUNDLE_ROOT)/"lib"
-
-$LOAD_PATH.unshift(BUNDLE_LIB)
-
-require "bundle"
-
-usage = <<-EOS.undent
-  brew bundle [-v|--verbose] [--file=<path>|--global]
-  brew bundle dump [--force] [--file=<path>|--global]
-  brew bundle cleanup [--force] [--file=<path>|--global]
-  brew bundle check [--file=<path>|--global]
-  brew bundle [--version]
-  brew bundle [-h|--help]
-
-  Usage:
-  Bundler for non-Ruby dependencies from Homebrew
-
-  brew bundle            install or upgrade all dependencies in a Brewfile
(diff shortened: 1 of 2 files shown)


xu-cheng:
Didn't we do the opposite in #86 as we only support Ruby 2.0 for this project.

mikemcquaid:
@xu-cheng We did. I think that was a bad idea, in hindsight. I don't mind whether we want to support Ruby 1.8 or 2.0, 10.9 and above only (which seems like a good idea, really) or the custom Homebrew Ruby but I do think we should do something to handle this issue.

mikemcquaid:
@xu-cheng Relatedly: I think it's probably been long enough that we can now drop the aliases too. Let me know what you think there.

xu-cheng:
First thing first, clearly we cannot just rename it back to `*.rb` as it will break this tool hard on old OS. That to be said, we may want to apply this PR when we move Homebrew core to ruby 2.0 in the future.

> I do think we should do something to handle this issue.

Looks to me is we should have a better ruby interpreter selection logic, at least for short term. To fix this, we can create a bash script similar to logic in `brew-cask`, where we can find suitable ruby interpreter in a defined order e.g. System ruby -> ruby in HOMEBREW_PREFIX -> ruby in PATH. Only if we cannot find any ruby 2.x, we will ask user to install ruby themselves or we install ruby for them.

>  I think it's probably been long enough that we can now drop the aliases too.

I don't see any harm to keep them around. Therefore, I would like to keep them instead of accidentally breaking someone's workflow.

mikemcquaid:
> First thing first, clearly we cannot just rename it back to *.rb as it will break this tool hard on old OS.

How long do we want to support old versions of OS X for this tool? 

> where we can find suitable ruby interpreter in a defined order e.g. System ruby -> ruby in HOMEBREW_PREFIX -> ruby in PATH

If we do this I think we should just look at the system Ruby and then Ruby in the PATH. That fixes my particular issue without breaking old versions. I don't think auto-install or Ruby in the prefix are necessary to search.

xu-cheng:
> How long do we want to support old versions of OS X for this tool?

Currently, I think bundle as a semi core command as it will be auto installed when invoking. Therefore, I think we should support old OS as long as Homebrew core supports it. 

I'm ok with the search logic you proposed.

mikemcquaid:
@xu-cheng See what you think now.

mikemcquaid:
@xu-cheng If we're going to take this approach: can you suggest how I can profile with `ruby-prof`? The wrapper script appears to mess with it.

mikemcquaid:
@xu-cheng It still feels really weird to me that if you call `brew bundle` with the system Ruby you're now gonna have a bash wrapper -> Ruby code -> bash wrapper -> Ruby code.

xu-cheng:
Unfortunately, this has to be it before we move core to Ruby 2.0. FWIW, currently cask is invoked in the same flow.

xu-cheng:
So this is a short term fix.

mikemcquaid:
Are you actively working on making Homebrew Ruby >2.0 only? If not, I don't think you can say it's short-term.

It doesn't _have_ to be this way, it just has to be if we want to use Ruby 2.0 in Homebrew Bundle and we want to support older OS X versions using it.

mikemcquaid:
And I don't think it's acceptable that this script is impossible to profile right now, really.

xu-cheng:
I'm currently active work on new update system which will one step before moving to 2.0. But still, I would require help as I have little access on old system. 

As for your problem, wait a bit, I may come out a more clever back.

mikemcquaid:
> I'm currently active work on new update system which will one step before moving to 2.0. But still, I would require help as I have little access on old system.

:cool:. We have the Vagrant images that should give you access to those old versions.

> As for your problem, wait a bit, I may come out a more clever hack.

Currently the best I've got is commenting out some lines in `brew.rb` so it loads the `.rb` file. For that reason I'd like if we can keep this in `cmd` for now.

mikemcquaid:
> As for your problem, wait a bit, I may come out a more clever hack.

Let's iterate on this in future PRs.

PR #128, opened 2015-11-20
check: speed up by checking less.

mikemcquaid:
We’ve been doing more checking here than is strictly necessary to verify. Unfortunately the slowest case is everything being fine and it’s really hard to improve as they processes we shell out to are slow.
I’ve tried speeding this up with `Thread`s and async `IO.popen` but neither actually improve things. Further speedup will have to happen in those commands instead.

CC @xu-cheng for review and @josh and @scottjg as this gets the github `brew bundle check` down from ~1.45s to ~1.15s on my machine. Not fast enough but getting better.


xu-cheng:
:+1:

josh:
:metal:

Issue #125, opened 2015-11-18
Error: undefined method `map' for nil:NilClass


nate-fm:
When I use the option to `dump`, I get the following error:

**Error: undefined method `map' for nil:NilClass**
Please report this bug:
    https://github.com/Homebrew/homebrew-bundle/issues/
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:78:in `formula_inspector'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `block in formulae_info'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `map'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `formulae_info'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:14:in `formulae'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:34:in `cask_requirements'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/dumper.rb:17:in `dump_brewfile'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/commands/dump.rb:4:in `run'
/usr/local/Library/Taps/homebrew/homebrew-bundle/cmd/brew-bundle:73:in `<main>'

**More output below to diagnose the issue:**

 $ ls Brewfile
Brewfile

 $ brew bundle dump
Error: undefined method `map' for nil:NilClass
Please report this bug:
    https://github.com/Homebrew/homebrew-bundle/issues/
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:78:in `formula_inspector'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `block in formulae_info'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `map'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:64:in `formulae_info'
/usr/local/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:14:in `formulae'
[...]

PR #129, opened 2015-11-24
BrewDumper: handle bad keg JSON more strictly.

mikemcquaid:
If we didn't get a keg back then ensure we don't try to read it in any form. Also, tweak how we read it so bad or missing data won't make it explode.

Closes https://github.com/Homebrew/homebrew-bundle/issues/125


xu-cheng:
:+1:

PR #130, opened 2015-11-26
Massively speedup `brew bundle check`.

mikemcquaid:
The main slowdown with `brew bundle` check was the repeated shelling out to other `brew` commands unnecessary. We can't avoid this for `brew cask` (for now) but we can do so for the others. Additionally
now that `brew bundle` and `brew services` are core commands we don't need to check if they are installed and we don't need to check if Homebrew is installed (as it's require for `brew bundle` to work).

As well all this: rename specs to be more consistent.

Depends on https://github.com/Homebrew/homebrew/pull/46400

CC @xu-cheng for review.

Testing this on a work project (`github/github` for @scottjg @josh) the speed difference is from:
```
brew bundle check  1.04s user 0.24s system 90% cpu 1.415 total
```

to:
```
brew bundle check  0.31s user 0.09s system 92% cpu 0.443 total
```


josh:
Wow, nice!

mikemcquaid:
@josh Happy thanksgiving :wink:

lib/bundle/brew_dumper.rb
```diff
@@ -59,12 +55,8 @@ def self.formula_aliases
...
 
     def self.formulae_info
-      begin
-        installed_formulae = JSON.load(`brew info --json=v1 --installed`) || []
-        installed_formulae.map { |info| formula_inspector info }
-      rescue JSON::ParserError
-        []
-      end
+      require "cmd/info"
+      Homebrew.formulae_json.map { |info| formula_inspector info }
```
xu-cheng: Where is `formulae_json` defined?

lib/bundle/cask_dumper.rb
```diff
@@ -5,11 +5,7 @@ def self.reset!
     end
 
     def self.casks
-      @casks ||= if Bundle.cask_installed?
-        `brew cask list -1 2>/dev/null`.split("\n").map { |cask| cask.chomp " (!)" }
-      else
-        []
-      end
+      @casks ||= `brew cask list -1 2>/dev/null`.split("\n").map { |cask| cask.chomp " (!)" }
```
xu-cheng: `Bundle.cask_installed?` is still required. Otherwise `brew bundle check` will automatically install cask, which would be unexpected.

lib/bundle/brew_dumper.rb
mikemcquaid: Sorry, haven't pushed that to the other PR yet. It may be able to be simplified further into using `Formula.hash` after I finish the outdated refactoring.

lib/bundle/cask_dumper.rb
mikemcquaid: :cool:

mikemcquaid:
@xu-cheng updated and ready for rereview.

lib/bundle/utils.rb
```diff
@@ -16,16 +16,10 @@ def self.system(cmd, *args)
...
-  end
-
   def self.cask_installed?
-    @cask ||= Kernel.system("brew command cask >/dev/null 2>&1")
-  end
-
-  def self.services_installed?
-    @services ||= Kernel.system("brew command services >/dev/null 2>&1")
+    @cask ||= begin
+      which("brew-cask") || which("brew-cask.rb")
```
xu-cheng: Why not use `brew command cask`?

xu-cheng:
:+1:

lib/bundle/utils.rb
mikemcquaid: More performant to not shell out and I figured this was nicer than `require "cmd/command"`

xu-cheng: I doubt it will reduce noticeable overhead as it's not hotspot here. And it feels too hacky to me.

xu-cheng:
Also, one thing I almost forget, you need to update minimal Homebrew revision after core PR is merged.

lib/bundle/utils.rb
mikemcquaid: Literally having it increases the runtime of `brew bundle check` by 10%. Trust me, I've actually been using `ruby prof bundle check` for all this stuff :wink:

mikemcquaid:
> Also, one thing I almost forget, you need to update minimal Homebrew revision after core PR is merged.

Yup :+1:

lib/bundle/utils.rb
xu-cheng: OK, your call.

(To be honest, the only thing I worry about is that after seeing these codes, other people start using private API in their own external commands, which could be easily broken)

mikemcquaid: Yeh. It makes me wonder about moving `brew bundle` into core.

PR #133, opened 2015-12-13
Require bundler explicitly

voanhduy1512:
Currently, when we want to run the test suite, we must run `bundle exec rake spec`. When run `rake spec` alone, `without bundler exec`, an exception is thrown `NameError: uninitialized constant Bundler` because we when run with `bundle exec`, `bundler` is automated required for us.

This a confused behaviour because in the code, we don't use bundler. And in fact, if we remove all the bundler stuffs, then just run `rake spec`, the test will pass.

So i make some changes in the `spec_helper`. First, require `bundler` explicitly. Second, run all specs under `Bundler.with_clean_env`, which make the test easy to understand and separate setup test env with the actual test.

### DIFF
--- a/spec/brew_installer_spec.rb
+++ b/spec/brew_installer_spec.rb
@@ -6,7 +6,7 @@ describe Bundle::BrewInstaller do
   let(:installer) { Bundle::BrewInstaller.new(formula, options) }
 
   def do_install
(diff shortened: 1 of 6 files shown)


mikemcquaid:
Makes sense! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #140, opened 2016-01-03
Use RFC2822 style when checking git commit date

cheungpat:
When git is configured with `date=relative`, `git show -s --format=%cd`
prints commit date as `38 minutes ago`, which `Time.parse` is unable to parse.

Using `%cD` force git to print in RFC2822 style regardless of git config.


xu-cheng:
Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #141, opened 2016-01-10
Add --zap command line switch

lvillani:
This allows cleanup to "zap" casks instead of uninstalling them,
performing a more thorough removal.

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -28,7 +28,13 @@ module Bundle::Commands
         end
       else
         if casks.any?
-          Kernel.system "brew", "cask", "uninstall", "--force", *casks
+          if ARGV.include?("--zap")
+            action = "zap"
+          else
+            action = "uninstall"
+          end
+
+          Kernel.system "brew", "cask", action, "--force", *casks
           puts "Uninstalled #{casks.size} cask#{casks.size == 1 ? "" : "s"}"
         end
 
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -29,7 +29,7 @@ require "bundle"
 usage = <<-EOS.undent
   brew bundle [-v|--verbose] [--file=<path>|--global]
   brew bundle dump [--force] [--file=<path>|--global]
-  brew bundle cleanup [--force] [--file=<path>|--global]
+  brew bundle cleanup [--force] [--zap] [--file=<path>|--global]
   brew bundle check [--file=<path>|--global]
   brew bundle [--version]
   brew bundle [-h|--help]
@@ -45,6 +45,7 @@ usage = <<-EOS.undent
   Options:
   -v, --verbose          print verbose output
   --force                uninstall dependencies or overwrite existing Brewfile
+  --zap                  zaps all files associated with Casks to be uninstalled
   --file=<path>          set Brewfile path
   --global               set Brewfile path to $HOME/.Brewfile
   -h, --help             show this help message and exit
--- a/spec/cleanup_command_spec.rb
+++ b/spec/cleanup_command_spec.rb
@@ -77,6 +77,22 @@ describe Bundle::Commands::Cleanup do
     end
   end
 
+  context "there are casks to zap" do
+    before do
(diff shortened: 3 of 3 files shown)


mikemcquaid:
I'm not quite sure what `--zap` does (even after reading `man brew-cask`)? Can you give examples about files that would be deleted?

lib/bundle/commands/cleanup.rb
```diff
@@ -28,7 +28,12 @@ def self.run
         end
       else
         if casks.any?
-          Kernel.system "brew", "cask", "uninstall", "--force", *casks
+          action = "uninstall"
+          if ARGV.include?("--zap")
+            action = "zap"
+          end
```
xu-cheng:
```ruby
if ARGV.include?("--zap")
  action = "zap"
else
  action = "uninstall"
end
```

xu-cheng:
Per @mikemcquaid's question, it is usually used to remove configure files. e.g. files inside `~/Library/`.

lvillani:
Just pushed a new version of this patch that includes @xu-cheng suggestion.

adityadalal924:
`zap` also implicitly runs `uninstall`, FWIW. Mostly used for preference files and the like.

mikemcquaid:
I'm not that keen on this as-is just because I think it would be good to do something similar for Homebrew rather than exposing something Cask-specific.

mikemcquaid:
@xu-cheng What are your thoughts on this? I'm probably :-1: as it's basically just `brew cask` behaviour but I don't feel super strongly.

xu-cheng:
I can see the value to support it. `brew cask zap` can be quite useful to purge all garbage files. So I would probably :+1: on this.

> I think it would be good to do something similar for Homebrew rather than exposing something Cask-specific.

Back then, I was thinking to add something similar to Homebrew. But gave up eventually due to too many edge cases.

mikemcquaid:
@xu-cheng Would you object if we passed on this? After all, users can still `brew cask zap` without this, right?

xu-cheng:
I'm OK either way. I think it mainly depends on how demand for this feature is.

mikemcquaid:
Will pass on this for now, then, sorry @lvillani and thanks for the PR.

lvillani:
Hi @MikeMcQuaid and @xu-cheng,

`brew cask zap` has changed behaviour some time last year and removed the ability to "zap" casks that have been removed (IIRC, it could zap casks regardless of their state before). This means I can no longer `brew bundle cleanup` to uninstall casks and then "zap" the stragglers, since the following usually happens after a cask has been uninstalled (e.g.):

```
(0) lvillani@Lorenzos-MacBook-Pro ~/D/_/homebrew-bundle (master=) $
brew cask zap imageoptim
==> Implied "brew cask uninstall imageoptim"
==> Moving App 'ImageOptim.app' back to '/usr/local/Caskroom/imageoptim/1.7.3/ImageOptim.app'.
==> Dispatching zap stanza
==> Running zap process for imageoptim; your password may be necessary
==> Trashing files:
~/Library/Caches/net.pornel.ImageOptim
~/Library/Caches/ImageOptimResults.db
~/Library/Preferences/net.pornel.ImageOptim.plist
~/Library/Saved Application State/net.pornel.ImageOptim.savedState
==> Removing all staged versions of Cask 'imageoptim'

(0) lvillani@Lorenzos-MacBook-Pro ~/D/_/homebrew-bundle (master=) $
brew cask zap imageoptim
==> Implied "brew cask uninstall imageoptim"
Error: It seems the App source '/Applications/ImageOptim.app' is not there.

```

Please note that after a `brew bundle cleanup` 90% of the time I fall in the second case, in which the source app bundle has been removed but configuration files are still on my system.

I'm kinda unsure whether to add a feature here or submit a PR to `homebrew-cask` on this.

Would you reconsider merging this PR (or an updated version thereof)?

MikeMcQuaid:
If `zap` definitely isn't run automatically on installation: yes, I'd consider merging an updated version of this PR (or reopening it if you undelete the branch).

lvillani:
I wasn't able to undelete the branch and I ended up opening a new PR instead (#323)

PR #145, opened 2016-02-02
Don't unconditionally return false when checking for upgradability

### LINKED ISSUE #144: brew bundle doesn't update outdated packages
I have this `rbenv-aliases` version `1.0.1` installed

```shell
$ brew info rbenv-aliases
rbenv-aliases: stable 1.1.0, HEAD
Make aliases for Ruby versions
https://github.com/tpope/rbenv-aliases
/usr/local/Cellar/rbenv-aliases/1.0.1 (7 files, 6.5K)
  Built from source
From: https://github.com/Homebrew/homebrew/blob/master/Library/Formula/rbenv-aliases.rb
```

If I try to list the outdated packages, the `rbenv-aliases` package is there

```shell
$ brew outdated
...
rbenv-aliases (1.0.1 < 1.1.0)
...
```

But if I run `brew bundle --file=Brewfile` againt the following `Brewfile`, nothing gets updated.

```shell
$ cat Brewfile
brew 'rbenv'
brew 'rbenv-vars'
brew 'rbenv-aliases'
brew 'ruby-build'

$ brew bundle --file=Brewfile
Succeeded in installing rbenv
Succeeded in installing rbenv-vars
Succeeded in installing rbenv-aliases
Succeeded in installing ruby-build
```

After that I still have `rbenv-aliases` version `1.0.1`.
any idea?

ilovezfs:
Closes #144

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -59,11 +59,11 @@ module Bundle
     end
 
     def self.outdated_formulae
-      @outdated_formulae ||= Bundle::BrewDumper.formulae.select { |f| f[:outdated?] }
+      @outdated_formulae ||= Bundle::BrewDumper.formulae.map { |f| f[:name] if f[:outdated?] }.compact
     end
 
     def self.pinned_formulae
-      @pinned_formulae ||= Bundle::BrewDumper.formulae.select { |f| f[:pinned?] }
+      @pinned_formulae ||= Bundle::BrewDumper.formulae.map { |f| f[:name] if f[:pinned?] }.compact
     end
 
     def installed?
--- a/spec/brew_installer_spec.rb
+++ b/spec/brew_installer_spec.rb
@@ -41,7 +41,7 @@ describe Bundle::BrewInstaller do
         { :name => "b", :outdated? => true },
         { :name => "c", :outdated? => false },
       ])
-      expect(Bundle::BrewInstaller.outdated_formulae.map { |f| f[:name] }).to eql(%w[a b])
+      expect(Bundle::BrewInstaller.outdated_formulae).to eql(%w[a b])
     end
   end
 
@@ -53,7 +53,7 @@ describe Bundle::BrewInstaller do
         { :name => "b", :pinned? => true },
         { :name => "c", :pinned? => false },
       ])
-      expect(Bundle::BrewInstaller.pinned_formulae.map { |f| f[:name] }).to eql(%w[a b])
+      expect(Bundle::BrewInstaller.pinned_formulae).to eql(%w[a b])
     end
   end
 


mikemcquaid:
@ilovezfs Could you consider adding a failing test that this fixes? I'm not sure I understand the fix.

lib/bundle/brew_installer.rb
```diff
@@ -55,7 +55,8 @@ def self.installed_formulae
     end
 
     def self.upgradable_formulae
-      outdated_formulae - pinned_formulae
+      upgradable = outdated_formulae - pinned_formulae
+      upgradable.nil? ? nil : upgradable.map { |u| u[:name] }
```
mikemcquaid: It seems to me this method should always return an array.

ilovezfs:
> Could you consider adding a failing test that this fixes? I'm not sure I understand the fix.
@mikemcquaid This is the minimal change to fix it, but you may actually want more extensive changes.

The problem is that installed_formulae is an array of formula names, whereas outdated_formulae and pinned_formulae are arrays of formulae. The output of upgradable_formulae is treated as if it's the same as installed_formulae, an array of formula names, but in reality it was the difference between outdated_formulae and pinned_formulae, both of which are arrays of formulae not formula names.

mikemcquaid:
> The problem is that installed_formulae is an array of formula names, whereas outdated_formulae and pinned_formulae are arrays of formulae. The output of upgradable_formulae is treated as if it's the same as installed_formulae, an array of formula names, but in reality it was the difference between outdated_formulae and pinned_formulae, both of which are arrays of formulae not formula names.

Thanks for explaining!

lib/bundle/brew_installer.rb
```diff
@@ -55,7 +55,8 @@ def self.installed_formulae
     end
 
     def self.upgradable_formulae
-      outdated_formulae - pinned_formulae
+      upgradable = outdated_formulae - pinned_formulae
+      upgradable.nil? ? [] : upgradable.map { |u| u[:name] }
```
mikemcquaid: `(outdated_formulae - pinned_formulae).map { |f| f[:name] }` would work?

ilovezfs: Yes. Or we could return names from outdated_formulae and pinned_formulae.

mikemcquaid: Actually, yeh, that seems like the best fix :+1:

ilovezfs:
@mikemcquaid How's that, with compact?

lib/bundle/brew_installer.rb
```diff
@@ -59,11 +59,11 @@ def self.upgradable_formulae
     end
 
     def self.outdated_formulae
-      @outdated_formulae ||= Bundle::BrewDumper.formulae.select { |f| f[:outdated?] }
+      @outdated_formulae ||= Bundle::BrewDumper.formulae.map { |f| f[:name] if f[:outdated?] }.compact
```
xu-cheng: `compact` is unnecessary as there won't be `nil` inside the input list.

xu-cheng:
You will need to update corresponding tests https://github.com/Homebrew/homebrew-bundle/blob/master/spec/brew_installer_spec.rb#L36-L58

lib/bundle/brew_installer.rb
ilovezfs: Um,
```
irb(main):001:0> [ 3, 4, 5, 6 ].map { |n| (n+3) if n > 3 }
=> [nil, 7, 8, 9]
```

xu-cheng: Oh, sorry my fault. I missed the `select` here.

ilovezfs: No problem. This was in lieu of doing .select {g(x)}.map {f(x)}. Is there any even better way than .map {f(x) if g(x)}.compact?

ilovezfs:
@xu-cheng better?

xu-cheng:
:+1: LGTM. But one last nit: could you update the commit message? Thanks.

ilovezfs:
@xu-cheng commit message is now not `nil`.

xu-cheng:
Thanks @ilovezfs

ilovezfs:
@xu-cheng You're welcome :)

PR #152, opened 2016-02-23
restart_service only if formula installed or upgraded

ktheory:
First off, thanks for making homebrew and Brewfiles. It's a boon for Mac development. :cake: 

This changes `restart_service: true` to only restart services if they’re installed or upgraded. Avoiding unnecessary restarts is faster, and less disruptive for developers who may be using those services while `brew bundle`'ing.

This changes the current default of restarting services on each `brew
bundle run`. If a user would like to keep the current behavior, they can specify `restart_service: :always` in their Brewfile.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -73,6 +73,18 @@ You can check there's anything to install/upgrade in the `Brewfile` by running:
 
 This provides a successful exit code if everything is up-to-date so is useful for scripting.
 
+### Restarting services
+
+You can choose whether `brew bundle` restarts a service every time it's run, or
+only when the formula is installed or upgraded. In you `Brewfile`:
+`Brewfile`:
+
+    # Always restart myservice
+    brew 'myservice', restart_service: true
+
+    # Only restart when installing or upgrading myservice
+    brew 'myservice', restart_service: :changed
+
 ## Note
 
 Homebrew does not support installing specific versions of a library, only the most recent one so there is no good mechanism for storing installed versions in a .lock file.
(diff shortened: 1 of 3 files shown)


mikemcquaid:
So, the advantage of `restart_service` running every time is it makes sure that the service is running after a `brew bundle`. I can see why your behaviour is useful: perhaps it could be `restart_service: :installed` or something?

ktheory:
> So, the advantage of restart_service running every time is it makes sure that the service is running after a brew bundle

Sure! TBH, I didn't know if that was a feature, or just incidental. :smile: 

AFAIK, other package managers that control daemons (like `apt` & `yum`) only restart the daemons if the underlying package changed (e.g., running `apt-get install nginx`), so I figured that'd be a good convention to follow by default. But granted, Brewfiles are a somewhat different use case.

> perhaps it could be `restart_service: :installed` or something?

How about `restart_service: :changed` ? That's a bit more clear that the restart applies to both installing & upgrading formula.

Let me know your preference, and I'll update the PR accordingly.

mikemcquaid:
> How about restart_service: :changed ? That's a bit more clear that the restart applies to both installing & upgrading formula.

Sounds good :+1:

ktheory:
@mikemcquaid: :ok: I amended the commit per the discussion, and added documentation to the Readme.

:ship: :question:

mikemcquaid:
Looking good! Looks like there's a Travis CI test failure and the test coverage has dropped a bit; would be good to get it back to 100%. Thanks!

ktheory:
Travis tests are hung, but pass locally for me.

mikemcquaid:

PR #154, opened 2016-02-29
Unlink and stop service for conflicting formulae.

mikemcquaid:
Unlink and stop the service for conflicting formulae or those specified
by a `conflicts_with` argument. This is useful when using `brew bundle`
to manage the transition between e.g. `homebrew/versions` versions.

Similar to https://github.com/Homebrew/homebrew-bundle/pull/124

CC @josh @xu-cheng for thoughts.

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -14,14 +14,16 @@ module Bundle
       @full_name = name
       @name = name.split("/").last
       @args = options.fetch(:args, []).map { |arg| "--#{arg}" }
+      @conflicts_with_arg = options.fetch(:conflicts_with, [])
       @restart_service = options.fetch(:restart_service, false)
     end
 
     def run
-      install_or_upgrade and restart_service!
+      install_change_state! && service_change_state!
     end
 
-    def install_or_upgrade
+    def install_change_state!
+      return false unless resolve_conflicts!
       if installed?
         upgrade!
       else
@@ -38,8 +40,13 @@ module Bundle
       @changed
     end
 
-    def restart_service!
-      restart_service? ? BrewServices.restart(@full_name) : true
+    def service_change_state!
+      if restart_service?
+        puts "Restarting #{@name} service." if ARGV.verbose?
+        BrewServices.restart(@full_name)
+      else
+        true
+      end
     end
 
     def self.formula_installed_and_up_to_date?(formula)
@@ -88,6 +95,41 @@ module Bundle
 
     def upgradable?
       BrewInstaller.formula_upgradable?(@name)
+    end
+
+    def conflicts_with
+      @conflicts_with ||= begin
+        conflicts_with = Set.new
+        conflicts_with += @conflicts_with_arg
+
+        if (formula_info = Bundle::BrewDumper.formula_info(@full_name))
+          if (formula_conflicts_with = formula_info[:conflicts_with])
+            conflicts_with += formula_conflicts_with
+          end
+        end
+
+        conflicts_with.to_a
+      end
+    end
+
+    def resolve_conflicts!
+      conflicts_with.each do |conflict|
+        if BrewInstaller.formula_installed?(conflict)
+          if ARGV.verbose?
+            puts <<-EOS
+Unlinking #{conflict} formula.
+It is currently installed and conflicts with #{@name}.
+EOS
+          end
+          return false unless Bundle.system("brew", "unlink", conflict)
+          if @restart_service
+            puts "Stopping #{conflict} service (if it is running)." if ARGV.verbose?
+            BrewServices.stop(conflict)
+          end
+        end
+      end
+
+      true
     end
 
     def install!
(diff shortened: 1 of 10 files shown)


lib/bundle/brew_installer.rb
```diff
@@ -14,21 +14,35 @@ def initialize(name, options = {})
...
-        upgrade!
+        if unlink?
+          unlink!
+        else
+          upgrade!
+        end
       else
         install!
       end
     end
```
xu-cheng: So for `brew "foo", unlink: true`, we will have:
* Unlink it if it's installed.
* Install (and link it) if it's not installed.

This is very counter-intuitive. From what I'm understanding, what you really need is an unlink-if-conflict feature.

mikemcquaid: > This is very counter-intuitive.

Agreed, will fix.



> From what I'm understanding, what you really need is an unlink-if-conflict feature. This also can be applied to services. I think what you want is stop-services-if-conflict feature(this may be better to be handled inside brew services).

That relies on conflicts being declared consistently and correctly and I don't think we can rely on that particularly when e.g. you're mixing third-party taps.

xu-cheng: I don't think we should use `brew "foo", unlink: :true` as DSL to handle what really is conflicting list. The reasons are (a) `brew "foo"` as a DSL is used to declare a dependency, and this is not. (b) it will interfere with the functions of `check` and `cleanup`. 

As I'm still thinking it's formula's duty to declare its conflicted formulae, I acknowledge this info can be missing. So how about a `brew "foo", conflicts_with: %w[a b]` to supply missing information. Therefore, rest of things can be handled as unlink/stop if conflict.

mikemcquaid:
@xu-cheng I really liked the approach you suggested, feels much cleaner, thanks. Let me know what you think of the PR now.

lib/bundle/brew_installer.rb
```diff
@@ -90,6 +97,41 @@ def upgradable?
...
+
+        conflicts_with.to_a
+      end
+    end
+
+    def unlink_conflicts!
+      conflicts_with.each do |conflict|
+        if BrewInstaller.formula_installed?(conflict)
+          if ARGV.verbose?
+            puts <<-EOS
```
xu-cheng: Use `EOS.undent`?

lib/bundle/brew_installer.rb
```diff
@@ -14,14 +14,16 @@ def initialize(name, options = {})
       @full_name = name
       @name = name.split("/").last
       @args = options.fetch(:args, []).map { |arg| "--#{arg}" }
+      @conflicts_with_arg = options.fetch(:conflicts_with, false)
```
xu-cheng: It seems to make more sense to:
```ruby
@conflicts_with_arg = options.fetch(:conflicts_with, [])
# OR
@conflicts_with_arg = options.fetch(:conflicts_with, Set.new)
```

Mixing boolean and array/set is kinda weird to me.

lib/bundle/brew_installer.rb
```diff
@@ -90,6 +97,41 @@ def upgradable?
...
+          if (formula_conflicts_with = formula_info[:conflicts_with])
+            conflicts_with += formula_conflicts_with
+          end
+        end
+
+        conflicts_with.to_a
+      end
+    end
+
+    def unlink_conflicts!
```
xu-cheng: Maybe call it `resolve_conflicts!`?

xu-cheng:
:+1:

lib/bundle/brew_installer.rb
mikemcquaid: :+1:

lib/bundle/brew_installer.rb
mikemcquaid: :+1:

lib/bundle/brew_installer.rb
mikemcquaid: Not available in Homebrew Bundle 😢

xu-cheng: It may be not available for test. But it's definitely available in bundle as it loads all of Homebrew core code. I think all you need to do is adding a dumpy `String#undent` method in `spec_helper.rb`.

lib/bundle/brew_services.rb
```diff
@@ -1,5 +1,11 @@
 module Bundle
   class BrewServices
+    def self.stop(name)
+      started = !`brew services list | grep "#{name}" | grep -q "started"`.chomp.empty?
```
xu-cheng: Oh, a final improvement. It may be a better idea to use `split/include?` instead of shelling out using `grep`. This is to avoid shell injection.

mikemcquaid: 👍 will open a new PR.

lib/bundle/brew_installer.rb
mikemcquaid: 👍 will open a new PR.

lib/bundle/brew_services.rb
mikemcquaid: 👍

PR #159, opened 2016-03-20
Improve style of `conflicts_with` implementation.

mikemcquaid:
Address @xu-cheng's comments from after https://github.com/Homebrew/homebrew-bundle/pull/154 was merged.

### DIFF
--- a/lib/bundle/brew_services.rb
+++ b/lib/bundle/brew_services.rb
@@ -1,7 +1,7 @@
 module Bundle
   class BrewServices
     def self.stop(name)
-      started = !`brew services list | grep "#{name}" | grep -q "started"`.chomp.empty?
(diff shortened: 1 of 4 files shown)


lib/bundle/brew_services.rb
```diff
@@ -1,7 +1,7 @@
 module Bundle
   class BrewServices
     def self.stop(name)
-      started = !`brew services list | grep "#{name}" | grep -q "started"`.chomp.empty?
+      started = `brew services list`.lines.grep(/^#{name} +started/).any?
```
xu-cheng: Probably need a regex escape here.

mikemcquaid: Can you think of a valid formula name that'd generate invalid regex?

xu-cheng: Formula with + in the name.

PR #160, opened 2016-03-24
Add brew-bundle exec command

mistydemeo:
brew-bundle exec is a wrapper around bundler, which executes bundler within Homebrew's superenv build environment. Using superenv helps make gems' Homebrew dependencies discoverable when building C extensions thanks to Homebrew's compiler shims and other flags set by superenv. It also helps prevent gems from detecting certain dependencies which weren't specified in the Brewfile, which can help make Brewfile/Gemfile combos more reliable longterm. The Brewfile is used to supply the list of formulae to use as dependencies.

This was written because a common issue, when converting projects to use a `Brewfile`, was that gems specified in the Gemfile would opportunistically pick up unspecified libraries. This led to dependencies being forgotten.

For example, the charlock_holmes gem requires icu4c, and is able to pick it up opportunistically even though it's keg-only. However, if the `bundle install` is run within superenv with icu4c not specified as a dependency, the build is properly isolated, and `brew bundle exec -- bundle install` will only succeed if icu4c is in the Brewfile. There are some superenv improvements to be made that can make this even more reliable.

~~(Why brew-burlap? You wrap up a bundle in burlap!)~~


xu-cheng:
If I understand correctly, you may want to make it more general as `brew bundle exec` command. So you could run `brew bundle exec gem/bundle/pip/npm` and so on.

mistydemeo:
This passes any arguments to the command to `bundle`, so `bundle exec` equivalents do work; e.g. `brew burlap exec [...]`. The intention is for it to act very transparently as a wrapper around bundler, with any bundler commands working.

mistydemeo:
Rereading, I get your suggestion now - making it more generic probably is a better idea, so, e.g., it can wrap other non-bundler external commands. That's a good idea, better than making this bundler specific! I'll look at

mistydemeo:
Added a commit which migrates this to a `brew bundle wrap` command. (Decided to name it `wrap` instead of `exec` because `bundle exec` already exists, and running `brew bundle exec bundle exec` would be confusing.) Any arguments after `wrap` are the full commandline which will be executed, e.g. `brew bundle wrap bundle install --binstubs --path vendor/gems`.

mistydemeo:
> /Users/travis/build/Homebrew/homebrew-bundle/lib/bundle/commands/wrap.rb:1:in `require': cannot load such file -- extend/ENV (LoadError)

That's interesting, is the CI run without access to Homebrew's own code?

mikemcquaid:
Yeh, tests should have Homebrew's code stubbed out. You can test locally with `rake`. I think `wrap` is a nicer name than `burlap` but `brewenv` `setenv` `superenv` `homebrewexec` or similar might be a nicer/more self-explanatory. We'll need enough tests here to maintain the 100% coverage we had before.

Otherwise: neat idea and 👍

cmd/brew-bundle.rb
```diff
@@ -31,6 +31,7 @@
   brew bundle dump [--force] [--file=<path>|--global]
   brew bundle cleanup [--force] [--file=<path>|--global]
   brew bundle check [--file=<path>|--global]
+  brew bundle wrap [command]
```
mikemcquaid: Could you add example usage to the README and identically to `dsl_spec.rb`?

cmd/brew-bundle.rb
```diff
@@ -71,6 +73,8 @@
     Bundle::Commands::Cleanup.run
   when "check"
     Bundle::Commands::Check.run
+  when "wrap"
+    Bundle::Commands::Wrap.run ARGV[1..-1]
```
xu-cheng: You may want to handle ARGV in a smarter way, so it allows certain flags passed to `brew-bundle` instead of whatever program is wrapped. For example, to support:
```bash
brew bundle wrap --file=/path/to/file command --command-flag
# OR
brew bundle wrap --verbose -- command --command-flag
```
You could change `ARGV.named[0]` in the case block to `ARGV.shift`, then 
```ruby
  args = []

  ARGV.each_with_index do |arg, i|
    if arg == "--"
      args = ARGV.slice!(i+1..-1)
      break
    elsif !arg.start_with?("-")
      args = ARGV.slice!(i..-1)
      break
    end
  end
```

So the rest of ARGV will be used in `brew bundle` while `args` will be used by the wrapper.

cmd/brew-bundle.rb
mistydemeo: Added to the README. What did you want in `dsl_spec.rb`?

cmd/brew-bundle.rb
mistydemeo: Added with a little tweaking, to make sure it works if an option precedes the command name.

mistydemeo:
I've been having a lot of trouble coming up with good names, which has me coming back around to `exec` despite the possibility of `brew bundle exec bundle exec rake` coming up. To fit with the other ones it feels like it should be a verb of some kind, and I felt that to many users something like `superenv` might not communicate very much. Currently using `isolate`, with the implication that the command is being isolated from the global environment.

mikemcquaid:
@mistydemeo I think I like `exec`; with `isolate` it's not clear what and how it's being isolated. Presumably the reason for this not being a separate command is that's it's using superenv to restrict to only the stuff in the `Brewfile`?

mikemcquaid:
@mistydemeo About the README: that was when I thought this was adding a DSL value; I see and understand now what this is doing so that nothing is needed in `dsl_spec.rb` 👍

cmd/brew-bundle.rb
```diff
@@ -61,8 +63,12 @@
   exit 0
 end
 
+# Pop the named command from ARGV, leaving everything else in place
+command = ARGV.named.first
+i = ARGV.find_index(command)
+ARGV.slice!(i)
```
xu-cheng: You probably want to make it `ARGV.delete_at(ARGV.index(command) unless command.nil?` to handle the case when command is nil.

mistydemeo:
Renamed to `exec`.

cmd/brew-bundle.rb
mistydemeo: Updated!

mikemcquaid:
Looks good to me 👍. :shipit:

PR #163, opened 2016-04-02
Cleanup: don't untap homebrew/core

bfontaine:
We currently get an error when running `brew bundle cleanup -f` if the bundle was dumped before tapping `homebrew/core` because `brew-bundle` tries to untap it while [Homebrew forbids it](https://github.com/Homebrew/homebrew/blob/e13ff6294a0bbc28fc98d6f797b1b56ce93ecd67/Library/Homebrew/cmd/untap.rb#L9).

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -66,7 +66,7 @@ module Bundle::Commands
       @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
       kept_taps = @dsl.entries.select { |e| e.type == :tap }.map(&:name)
       current_taps = Bundle::TapDumper.tap_names
-      current_taps - kept_taps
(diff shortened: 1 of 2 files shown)


ilovezfs:
I think maybe we should just make `brew untap homebrew/core` not an error but a noop + warning to avoid having to do this.

mikemcquaid:
@ilovezfs I think that the warning in that case will still be confusing in this case but 👍 in the general case.

mikemcquaid:
@bfontaine Merge conflicts but otherwise 👍

mikemcquaid:
Actually: @xu-cheng fixed this in https://github.com/Homebrew/homebrew-bundle/pull/164.

xu-cheng:
Sorry, I missed this PR when I submit the change.

Issue #167, opened 2016-04-14
Investigate Mac App Store support


mikemcquaid:
Could be done perhaps using e.g. https://github.com/argon/mas if it's reliable.

PR #172, opened 2016-04-20
Add Mac App Store support using mas-cli.

mikemcquaid:
Closes https://github.com/Homebrew/homebrew-bundle/issues/167.
Closes https://github.com/argon/mas/issues/21.


kibitan:
wow! its awesome!

mikemcquaid:
@kibitan Thanks 🙇

Readme.md
```diff
@@ -40,6 +42,7 @@ brew 'emacs', args: ['with-cocoa', 'with-gnutls']
 cask 'google-chrome'
 cask 'java' unless system '/usr/libexec/java_home --failfast'
 cask 'firefox', args: { appdir: '~/my-apps/Applications' }
+mac_app_store '1Password', 443987910
```
xu-cheng: Can we call this DSL `app` or `mas`? `mac_app_store` is unnecessary long.

Also the number is confusing. Can we make it `app '1Password', id: 443987910`?

lib/bundle/dumper.rb
```diff
@@ -20,6 +20,7 @@ def self.dump_brewfile
       content << cask_before_formula
       content << BrewDumper.dump
       content << cask_after_formula
+      content << MacAppStoreDumper.dump
```
xu-cheng: Personally, I would ask a flag to disable this. Because not everyone want to use `brew bundle` to manage apps.

lib/bundle/mac_app_store_installer.rb
```diff
@@ -0,0 +1,21 @@
+module Bundle
+  class MacAppStoreInstaller
+    def self.install(name, id)
```
xu-cheng: Fail to check whether mas has been installed here. Also should we auto install it?

mikemcquaid: 👍 I don't know about autoinstalling.

Readme.md
mikemcquaid: Could do `mas`. Could do the `id`.

mikemcquaid: Although, technically the name is only used in the UI and not passed to commands but makes it more readable.

mikemcquaid: And my concern was that `id:` made it look like an option when it is not optional.

lib/bundle/dumper.rb
mikemcquaid: I think it's enough to assume that if you don't have `mas` installed you don't want this behaviour, right?

lib/bundle/mac_app_store_installer.rb
xu-cheng: You could see git history on how we auto install brew-cask(when it used to be a formula)

lib/bundle/dumper.rb
xu-cheng: I don't think we should make such assumption, which feels like you have to either use mas without bundle installed and use bundle without mas installed in order to separate manage homebrew and apps.

lib/bundle/mac_app_store_installer.rb
mikemcquaid: Yeh, sorry, I mean I know how to do so but I'm just unsure if we want to. I guess if someone is running `brew bundle` on a `Brewfile` that contains `mas` entries: we probably do.

lib/bundle/dumper.rb
mikemcquaid: Do you have `mas` installed and not want it dumping to your `Brewfile`?

xu-cheng: You may want to use `mas` to install some apps. And you want to use `Brewfile` to manage dev env. That doesn't necessary mean you want to put them together. That's because your `Brewfile`(hence dev setup) is stable, but the list of **all** your mac apps is not.

Even more, I would publicly share my `Brewfile` in Github but not necessary mac apps list.

Readme.md
mikemcquaid: Note that taps also use a non-option here.

xu-cheng: One could easily understand what tap remote is. But I would certainly be confused by these strange magic numbers.

lib/bundle/dumper.rb
mikemcquaid: It will already not dump anything unless `mas` is installed. I think this argument about Mac Apps applies equally to brews, casks and taps. You can always edit your `Brewfile` after dumping if you have `mas` installed but don't want to use it for some reason. I don't think an option is useful unless it defaults to not dumping at which point this feature is not discoverable or useful. If you'd like to add an option in a follow-up PR for personal usage: that's fine.

lib/bundle/mac_app_store_installer.rb
mikemcquaid: Yeh, gonna auto-install like we did with `cask`.

mikemcquaid:
All feedback except the option addressed.

kibitan:
:tada:

PR #174, opened 2016-04-29
Fix to dumps fomula args in same order

ravelll:
Fix to dumps args in same order when a formula installed with args specified in different order on different machines.

In current implementation, as I present below, there is the case that has difference about its order.

```diff
-brew 'gcc', args: ['with-jit', 'with-nls', 'with-all-languages']
+brew 'gcc', args: ['with-all-languages', 'with-jit', 'with-nls']
```

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -20,7 +20,7 @@ module Bundle
         if f[:args].empty?
           "brew '#{f[:full_name]}'"
         else
-          args = f[:args].map { |arg| "'#{arg}'" }.join(", ")
+          args = f[:args].map { |arg| "'#{arg}'" }.sort.join(", ")
           "brew '#{f[:full_name]}', args: [#{args}]"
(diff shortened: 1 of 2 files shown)


mikemcquaid:
👍 great PR.

> Fix to dumps args in same order when a formula installed with args specified in different order on different machines.

Can you elaborate a bit on this? I'm interested in hearing more about it, thanks.

ravelll:
Thanks ☺️ 

I reinstalled gcc with some args on two OS X machines.
Following commands are these executed on each machines.

```sh
# on machine 1
brew reinstall gcc --with-all-languages --with-jit --with-nls
# on machine 2
brew reinstall gcc --with-jit --with-nls --with-all-languages
```

Then 'brew bundle dump' on two, I found diff in Brewfiles as I wrote to #.
(Diff is one checked on machine2.)

mikemcquaid:
Seems reasonable 👍

Issue #176, opened 2016-05-16
Formulae that have been renamed should be handed better


technicalpickles:
`brew bundle check` was introduced over in https://github.com/Homebrew/homebrew-bundle/issues/100 , and we use it a `script/bootstrap` script like:

```shell
  brew update >/dev/null
  brew bundle check &>/dev/null || brew bundle
```

Ideally, `brew bundle` if anything isn't installed, but I ran into a project that was triggering `brew bundle` every run of `bootstrap`. To try to identify why, I started making changes to `lib/bundle/commands/check.rb` to try to figure out what specifically was triggering it. I'll follow up with a PR when I can, but I found the thing that was causing `brew bundle check` to return false:

```
$ brew bundle check --verbose
brew bundle can't satisfy your Brewfile's dependencies.
Install missing dependencies with `brew bundle install`.

Formulae to install:
- heroku-toolbelt
Kernel.exit
Error: Kernel.exit

$ brew install heroku-toolbelt
Warning: heroku-toolbelt already installed, it's just not migrated
You can migrate formula with `brew migrate heroku`
Or you can force install it with `brew install heroku --force`
```

I am able to fix this on my machine by running `brew unlink heroku-toolbelt; brew install heroku --force`, and updating `Brewfile` to use `heroku` instead of `heroku-toolbelt`. However, I'm worried about others using this project not being able to see the information they need to fix their environment.

I did try reproducing this with installing herokut-toolbelt, but wasn't quite able to get it to fail in the same way.

mikemcquaid:
Migrations are done by `brew update`. Can you try to check out an old version of `Library/Taps/Homebrew/homebrew-core` before the `heroku` rename, install `heroku-toolbelt`, and then run the `script/bootstrap` again and see if you can reproduce?

MikeMcQuaid:
> Can you try to check out an old version of Library/Taps/Homebrew/homebrew-core before the heroku rename, install heroku-toolbelt, and then run the script/bootstrap again and see if you can reproduce?

@technicalpickles Gentle ping on this, thanks.

technicalpickles:
The rename commit is https://github.com/Homebrew/homebrew-core/commit/11432e047c3373497fb12285190dc0d5fbe597df so I checked out it's parent. `brew uninstall heroku; brew uninstall heroku-toolbelt` to get to a clean slate, then `brew install heroku-toolbelt`. I ran `brew update` and saw among other things:

```
[...]

PR #177, opened 2016-05-16
Include verbose output for check to list out what needs to be done

technicalpickles:
I mentioned this in https://github.com/Homebrew/homebrew-bundle/issues/176 . I needed a way to figure out why `brew bundle` wasn't installing all the dependencies to satisfy `brew bundle check`.

This can be tested with `brew bundle check --verbose`.

This diff probably could be smaller, but I factored out most of the variables to their own methods so I knew they could be used in the verbose section.

One last thing is: I assume that there's checks like `return false if requested_taps.empty?` to avoid load time / memory for loading the taps tapped if none were requested.

### DIFF
--- a/lib/bundle/commands/check.rb
+++ b/lib/bundle/commands/check.rb
@@ -9,9 +9,28 @@ module Bundle::Commands
     end
 
     def self.run
-      if any_taps_to_tap? || any_casks_to_install? || any_apps_to_install? || any_formulae_to_install?
+      if unsatisfied?
         puts "brew bundle can't satisfy your Brewfile's dependencies."
         puts "Install missing dependencies with `brew bundle install`."
+        if ARGV.verbose?
+          puts
+          if taps_to_tap.any?
+            puts "Taps to tap:"
+            taps_to_tap.each {|tap| puts "- #{tap}" }
+          end
+          if casks_to_install.any?
+            puts "Casks to install:"
+            casks_to_install.each {|cask| puts "- #{cask.name}" }
+          end
+          if apps_to_install.any?
+            puts "Apps to install:"
+            apps_to_install.each {|app| puts "- #{app}" }
+          end
+          if formula_to_install.any?
+            puts "Formulae to install:"
+            formula_to_install.each {|formula| puts "- #{formula}" }
+          end
+        end
         exit 1
       else
         puts "The Brewfile's dependencies are satisfied."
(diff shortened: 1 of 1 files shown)


lib/bundle/commands/check.rb
```diff
@@ -20,36 +39,77 @@ def self.run
...
-    def self.any_formulae_to_install?
-      @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
-      requested_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
-      requested_formulae.any? do |f|
-        !Bundle::BrewInstaller.formula_installed_and_up_to_date?(f)
+    def self.requested_formulae
+      @requested_formulae ||= dsl.entries.select { |e| e.type == :brew }.map(&:name)
+    end
+
+    def self.formula_to_install
```
mikemcquaid: `self.formulae_to_install`

mikemcquaid:
> One last thing is: I assume that there's checks like return false if requested_taps.empty? to avoid load time / memory for loading the taps tapped if none were requested.

Indeed, the goal is to make `brew bundle check` do as little as possible. As soon as it figures out there's missing dependencies it fails. I'm 🆒 with doing more/everything in the verbose case, though.

You've probably seen from CI but just to be explicit: we'll need some more test cases for this.

mikemcquaid:
Oh, and, 👍 and thanks for making a PR 🙇

technicalpickles:
```
  2) Bundle::Commands::Check when formulae are not installed raises an error
     Failure/Error: expect { do_check }.to raise_error(SystemExit)
     
       expected SystemExit, got #<NoMethodError: undefined method `verbose?' for ["--pattern", "spec/**{,/*/**}/*_spec.rb"]:Array> with backtrace:

```

Is there something homebrew/bundle includes that would include `verbose?` on ARGV? I copied this from other commands using it, and it works when run, but not in the tests. Digging more, but figured I'd ask in case someone knows off hand.

technicalpickles:
Looks like it's https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ARGV.rb . Now to figure out how to get that in testing and tested. It seems there's stubbing that happens like:

```ruby
      allow(ARGV).to receive(:include?).and_return(true)
```

I could just do that for verbose, but that'd end up in several places.

MikeMcQuaid:
> I could just do that for verbose, but that'd end up in several places.

Yeh, that's fine, thanks 👍

technicalpickles:
I haven't gotten a chance to get back to this, and I'm about to go on leave so won't get to it any time soon 😓 

This PR itself isn't super important, but it was useful to help identify what was going on in https://github.com/Homebrew/homebrew-bundle/issues/176

MikeMcQuaid:
Ok, thanks for the heads up. Will pass on this for now.

PR #183, opened 2016-06-18
Bundle.system: better way to handle verbose

xu-cheng:
Use Kernel.system instead of our own logic to handle verbose output. This
allows us to show progress bar for downloading.


MikeMcQuaid:
👍

PR #184, opened 2016-06-28
[Doc] Readme.md - No longer need to tap caskroom/cask

fanquake:


MikeMcQuaid:
You don't _need_ to but it's still not a bad idea for something to script as it means you won't have it autotapped on first run. Thanks anyway, though!

PR #186, opened 2016-07-01
Remove aliases.

MikeMcQuaid:
These were used for transitional purposes and are no longer needed.

CC @UniqMartin


UniqMartin:
:+1:

PR #196, opened 2016-08-19
brew services related fix

xu-cheng:
* check whether brew services is installed. Otherwise, `brew bundle dump`
  will install brew services.
* remove duplicate codes on checking services started.
* avoid opportunistic regex, e.g. we should not match `started` in formula or
  plist file name.


MikeMcQuaid:
👍

PR #204, opened 2016-09-28
[Fix #188] Bundle check should verify services have been started

### LINKED ISSUE #188: bundle check should verify services have been started
If `restart_service: true` or `start_service: true` are set then `brew bundle check` should ensure that all the services have been started by `brew services`.

voanhduy1512:


lib/bundle/commands/check.rb
```diff
@@ -52,5 +52,13 @@ def self.any_apps_to_install?
       current_apps = Bundle::MacAppStoreDumper.app_ids
       (requested_apps - current_apps).any?
     end
+
+    def self.any_formulae_to_start?
+      @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      @dsl.entries.select { |e| e.type == :brew }.any? do |e|
+        formula = Bundle::BrewInstaller.new(e.name, e.options)
+        !(formula.restart_service? && Bundle::BrewServices.started?(e.name))
```
MikeMcQuaid:
```ruby
needs_to_start = formula.start_service? || formula.restart_service?
next unless needs_to_start
!Bundle::BrewServices.started?(e.name)
```

lib/bundle/commands/check.rb
```diff
@@ -10,7 +10,7 @@ def self.reset!
     end
 
     def self.run
-      if any_taps_to_tap? || any_casks_to_install? || any_apps_to_install? || any_formulae_to_install?
+      if any_taps_to_tap? || any_casks_to_install? || any_apps_to_install? || any_formulae_to_install? || any_formulae_to_start?
         puts "brew bundle can't satisfy your Brewfile's dependencies."
         puts "Install missing dependencies with `brew bundle install`."
```
MikeMcQuaid: `Satisfy missing dependencies with ...` as this no longer just does an install.

lib/bundle/commands/check.rb
```diff
@@ -10,7 +10,7 @@ def self.reset!
     end
 
     def self.run
-      if any_taps_to_tap? || any_casks_to_install? || any_apps_to_install? || any_formulae_to_install?
+      if any_taps_to_tap? || any_casks_to_install? || any_apps_to_install? || any_formulae_to_install? || any_formulae_to_start?
```
MikeMcQuaid: Split this onto multiple lines so it doesn't take more than 80 characters.

lib/bundle/commands/check.rb
voanhduy1512: Right now Formula doesn't have option `start_service` in dsl. Do you want me to add it too?

voanhduy1512: Added. Can you take a look at it

lib/bundle/commands/check.rb
```diff
@@ -52,5 +56,13 @@ def self.any_apps_to_install?
       current_apps = Bundle::MacAppStoreDumper.app_ids
       (requested_apps - current_apps).any?
     end
+
+    def self.any_formulae_to_start?
+      @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      @dsl.entries.select { |e| e.type == :brew }.any? do |e|
+        formula = Bundle::BrewInstaller.new(e.name, e.options)
+        !((formula.restart_service? || formula.service_needs_started?) && Bundle::BrewServices.started?(e.name))
```
MikeMcQuaid: The logic here is pretty hard to follow. Could you split `formula.restart_service? || formula.service_needs_started?` into a separate variable, `next` if it's `false`y and then have just `!Bundle::BrewServices.started?(e.name)` as the last statement? Thanks!

voanhduy1512: Sorry, i thought i've done it from your first comment

lib/bundle/brew_installer.rb
```diff
@@ -31,6 +32,10 @@ def install_change_state!
       end
     end
 
+    def service_needs_started?
```
MikeMcQuaid: @voanhduy1512 Sorry: absolutely last thing: can you renamed this to `start_service?` Thanks!

lib/bundle/brew_installer.rb
```diff
@@ -31,6 +32,10 @@ def install_change_state!
       end
     end
 
+    def start_service?
+      @start_service
+    end
+
     def restart_service?
       # Restart if `restart_service: :always`, or if the formula was installed or upgraded
       @restart_service && (@restart_service.to_s != 'changed' || changed?)
```
MikeMcQuaid: Did some local testing and this behaviour is not correct for this PR; it'll only return `true` if `:restart_service => true` and it's been upgraded. This method should mirror `start_service?` above and be just `@restart_service` and a new method should be added called `restart_service_needed?` which uses the `changed` logic. Make sense? Sorry; I know I said the last change was the last one but I tested this locally and it doesn't work properly for `restart` otherwise.

voanhduy1512: Ok, let see if i understand you correctly.

The `restart_service?` now use both for installing and checking. But now the logic of them is different (with this PR). So we return the `restart_service?` to it old behaviour, which is return `@restart_service` directly, and add a new `restart_service_needed?` to determine if we need to restart the service or not when install or upgrade it.

voanhduy1512:
Fixed everything, can you take a look at it. Thank you

MikeMcQuaid:

Issue #157, opened 2016-03-07
Deterministically create `bundle dump`


demosdemon:
Dump files should have a deterministic output. At least, it would be nice so I can keep them versioned without lines jumping around all over the place randomly.

I haven't looked at the code base in a while, but if I recall correctly, there's a dependency tree traversal when outputting. The nodes should probably be sorted lexicographically before descending further into the tree.

I'll remind myself to look at this later and drum up a PR if no one else has when I get around to it.

mikemcquaid:
I think this would be cool but we don't really accept feature requests and a PR would be welcome.

PR #206, opened 2016-09-30
Sort dependencies by name during tsort

josephfrazier:
This should help make the dump more deterministic, but I'm not sure if it's a complete fix for https://github.com/Homebrew/homebrew-bundle/issues/157

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -106,7 +106,7 @@ module Bundle
       include TSort
       alias_method :tsort_each_node, :each_key
       def tsort_each_child(node, &block)
-        fetch(node).each(&block)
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Can you provide some sample before/after output? Thanks!

josephfrazier:
Sure, here's a gist showing the before/after diff on my machine: https://gist.github.com/josephfrazier/d3b962333fd13255adc51df68aabcd55/revisions

MikeMcQuaid:
Seems no worse than before so 👍

MikeMcQuaid:
And: thanks for the PR 😍

josephfrazier:
Cheers for the merge!

PR #210, opened 2016-10-02
brew-bundle: use Homebrew help method.

MikeMcQuaid:
Use Homebrew’s new method of printing out `--help`.


MikeMcQuaid:
Don't merge this until we've tagged a new Homebrew/brew version.

PR #215, opened 2016-10-15
Implement dependency checking during cleanup

### LINKED ISSUE #143: Don't cleanup dependencies of whitelisted formulae
Is it possible to prevent `brew bundle cleanup` from uninstalling dependencies of the whitelisted formulae? This would facilitate a nice use-case: keep a .Bundle file of just the formula that you actually care about (without their dependencies explicitly), and still be able to run `brew bundle clean` without breaking them all.

mivok:
This change grabs the dependencies of all formulae mentioned in the
Brewfile, and adds them to the list of formulae to keep during cleanup.

When looking at the formula list, I saw dependencies listed for the formulae I have installed, so it looks like dependencies are now being tracked as suggested in #143. Let me know if this is the wrong approach.

Fixes #143

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -59,6 +59,11 @@ module Bundle::Commands
       kept_formulae.map! { |f| Bundle::BrewDumper.formula_aliases[f] || f }
       current_formulae = Bundle::BrewDumper.formulae
       all_deps = Hash[current_formulae.map { |f| [f[:name], f[:dependencies]] }]
+      all_reqs = Hash[current_formulae.map { |f|
+        [f[:name], f[:requirements].map { |r| r["default_formula"] }.select {
+          |r| ! r.nil? } ]
+      }]
+      all_deps.merge!(all_reqs) { |k, deps, reqs| deps + reqs }
       dependencies = {}
       kept_formulae.each { |f| dependencies[f] = all_deps[f] }
       # Work out nested dependencies
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -58,6 +58,20 @@ module Bundle::Commands
       kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
       kept_formulae.map! { |f| Bundle::BrewDumper.formula_aliases[f] || f }
       current_formulae = Bundle::BrewDumper.formulae
+      all_deps = Hash[current_formulae.map { |f| [f[:name], f[:dependencies]] }]
+      dependencies = {}
+      kept_formulae.each { |f| dependencies[f] = all_deps[f] }
+      # Work out nested dependencies
+      old_dep_count = 0
+      while old_dep_count != dependencies.count do
+        old_dep_count = dependencies.count
+        nested_deps = {}
+        dependencies.values.flatten.each { |f| nested_deps[f] = all_deps[f] }
+        dependencies.merge!(nested_deps)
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Nice work here. This ties in a bit with what @alyssais is doing in https://github.com/Homebrew/brew/pull/1082 so interested in her thoughts here.

alyssais:
It's a shame we don't have access to the formula instances, because it means logic for e.g. finding recursive dependencies has to be duplicated… (nice approach there @mivok btw!)

MikeMcQuaid:
@alyssais That's something that could be provided in the JSON output? To be more specific: was also wondering if your changes will end up using the Tab when going through this path.

mivok:
Thanks!

So, it looks like we might have access to the formula instances after all, I can do `Formulary.find_with_priority("formulaname")` to grab it, and then have access to recursive_dependencies and recursive_requirements (shamelessly discovered by digging through the code for `brew deps`).

I have a replacement version of the code that uses those methods and also handles requirements, which is missing from this code. I'm working on tests and will resubmit.

MikeMcQuaid:
I would avoid using the Formula instances for now; they aren't considered publicly accessible APIs for use like this.

mivok:
OK, I'll leave the implementation of dependency checking as is.

I realized earlier that the existing code in this PR doesn't deal with requirements, which causes issues if the requirements are satisfied by installing a formula. While testing on my system, I had the cleanup command try to uninstall perl that was needed by vim (among other things). I'll probably add another commit which will prevent uninstalling any formulae that were installed to satisfy requirements.

mivok:
I've added two more commits. The first deals with formulae that were installed as requirements, as mentioned above. The second deals with another issue I found where dependencies for formulae specified with their full name (e.g. homebrew/x11/foo) were not resolved, causing their dependencies to be marked for removal.

alyssais:
> To be more specific: was also wondering if your changes will end up using the Tab when going through this path.

@MikeMcQuaid yes, with my changes `brew uninstall x` first checks to see if dependencies are listed in the tab. Only if they're not does it fall back to checking the dependencies of the current version of the formula.

MikeMcQuaid:
@alyssais Thanks! Additionally, does dumping the `dependencies` JSON end up using the stuff from the Tab ever?

alyssais:
@MikeMcQuaid nope. That currently happens [here](https://github.com/alyssais/brew/blob/7eaa7bb3e2d799832f27e2a6ff2a18d0f09b7976/Library/Homebrew/keg.rb#L90-L123).

MikeMcQuaid:
@alyssais Feature request: use that code to speed up generating the JSON and add recursive dependencies to the JSON output. Thoughts? Would that be something you could or would be interested in tackling?

alyssais:
> @alyssais Feature request: use that code to speed up generating the JSON and add recursive dependencies to the JSON output. Thoughts? Would that be something you could or would be interested in tackling?

@MikeMcQuaid happy to try, although it still feels like a lot of effort to be going to to avoid accessing the `Formula` instances.

MikeMcQuaid:
@alyssais If you could that'd be great ❤️. Alternatively, if you feel like just making this use `Formula` instances more widely that'd work too.

MikeMcQuaid:
Thanks for the PR but passing on this in favour of the simpler https://github.com/Homebrew/homebrew-bundle/pull/263.

mivok:
No problem. The solution in #263 makes a lot of sense. I'm looking forward to trying it out. Thanks!

PR #219, opened 2016-10-19
Fix deprecations.

### LINKED ISSUE #218: Warning: Calling puts_columns is deprecated
```bash
Warning: Calling puts_columns is deprecated!
Use puts Formatter.columns instead.
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/commands/cleanup.rb:18:in `run'
Please report this to the homebrew/bundle tap!
```

I'm seeing that when using `brew bundle`.

MikeMcQuaid:
Fix the deprecations of `puts_columns` and `MultiFormatter[]`.

Fixes #218.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -13,7 +13,7 @@ $:.unshift(STUB_PATH)
 
 Dir.glob("#{PROJECT_ROOT}/lib/**/*.rb").each { |f| require f }
 
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
+SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new [
   SimpleCov::Formatter::Codecov,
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -37,6 +37,11 @@ def require(path)
   end
 end
 
+class Formatter
```
reitermarkus: It's a `module`.

MikeMcQuaid: Works fine either way; just for stubbing.

reitermarkus: Yeah, just wanted it to be consistent with what's in core. Out of curiosity, why exactly is this being stubbed?

MikeMcQuaid: The test suite doesn't include Homebrew code.

PR #221, opened 2016-10-20
Work in progress - basic linkapps support

sachse:
I still need to wrap my head around your test setup to add some specs, but I figured I might as well open this PR now to see if there's any feedback.


MikeMcQuaid:
Sorry, very likely to pass on this as the intention is to remove `linkapps` support from Homebrew at some point in the future.

sachse:
Oh damn, I see. Are you planning some kind of replacement feature that can be used with formulas that build an .app bundle? I know `linkapps` isn't perfect but the functionality it provides is useful.

MikeMcQuaid:
We are not, sorry. Things that create .app bundles should be ideally installed by Homebrew Cask which handles these cases far better.

PR #231, opened 2016-11-02
bundle: fix brew cask installation detection.

### LINKED ISSUE #226: Bundle outputs Cask silliness.
No `cask`s or `caskroom` tap in my `Brewfile`, yet `brew bundle cleanup -f` repeatedly provides:

```
~> brew bundle cleanup -f
Error: No available Cask for ==> Tapping caskroom/cask
Uninstalled 2 casks
Untapping caskroom/cask... (3,529 files, 3.5M)
Untapped 0 formulae
```

And it keeps happening. Running `brew bundle cleanup -f` seconds after the first `brew bundle cleanup -f` produces the same output; no other `brew` commands or anything done in the terminal between.

Homebrew-related env:

```
HOMEBREW_CACHE=/usr/local/var/homebrewcache
HOMEBREW_LOGS=/usr/local/var/homebrewcache/logs
HOMEBREW_DEVELOPER=1
HOMEBREW_SANDBOX=1
HOMEBREW_NO_AUTO_UPDATE=1
HOMEBREW_NO_ANALYTICS=1
```

Taps:

```
domt4/autoupdate
domt4/chromium
domt4/crypto
homebrew/bundle
homebrew/command-not-found
homebrew/core
homebrew/dev-tools
homebrew/devel-only
homebrew/dupes
homebrew/gui
homebrew/services
homebrew/versions
```

```
cat Brewfile | grep cask
$?
1
```

MikeMcQuaid:
Fix the check now that `brew-cask` has moved into Homebrew/brew.

CC @DomT4 to check this fixes your issue.

Fixes #226.

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -27,7 +27,8 @@ module Bundle
 
   def cask_installed?
     @cask ||= begin
-      which("brew-cask") || which("brew-cask.rb")
+      File.directory?("#{HOMEBREW_PREFIX}/Caskroom") ||
+        File.directory?("#{HOMEBREW_REPOSITORY}/Library/Taps/caskroom")
     end
   end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -37,6 +37,9 @@ module Kernel
   end
 end
 
+HOMEBREW_PREFIX = Pathname.new("/usr/local")
+HOMEBREW_REPOSITORY = Pathname.new("/usr/local/Homebrew")
+
 module Formatter
   module_function
 


DomT4:
👍 resolves the issue as long as I delete the `Caskroom` directory that had for some reason stuck around in my `HOMEBREW_PREFIX` despite it being untapped a couple weeks back.

DomT4:
_(Potentially because `brew` no longer has ownership rights over `/usr/local` itself and would need elevated permissions to remove the directory? But unsure why that would fail silently if that was the case)_

MikeMcQuaid:
@DomT4 Is there another check you think might be better e.g. checking for it being tapped?

DomT4:
```ruby
Pathname.new(Tap::TAP_DIRECTORY/"caskroom").directory?
```
Maybe? Or something more simple like:
```ruby
Tap.names.map.include?("caskroom/cask")
```

DomT4:
Don't need the `map` there actually, thinking about it.

PR #234, opened 2016-11-03
Do not auto install cask tap for `brew bundle dump`

xu-cheng:
Since Homebrew began to bundle Homebrew Cask lib directly,
`which("brew-cask.rb")` will always return true. This makes
`brew bundle dump` to think Homebrew Cask is always installed, and
generates the incorrect Brewfile as result.

----

I am still quiet busy recently. Here is a quick bug fix I just found out. ;)

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -27,7 +27,7 @@ module Bundle
 
   def cask_installed?
     @cask ||= begin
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Duplicate of https://github.com/Homebrew/homebrew-bundle/pull/231 but thanks for the PR, @xu-cheng!

Issue #233, opened 2016-11-02
Should mas test if it is signed in?


ffittschen:
Hi!

I recently added the Mac App Store apps to my `Brewfile` using `mas`, and received some errors, that I'm not signed in with `mas`. I use the `Brewfile` mostly on fresh installs of macOS, so `mas` will be either installed explicitly before calling it later in the `Brewfile`, or it will be installed implicitly using the check in `Bundle::MacAppStoreInstaller.install()`.

Either way, it will result in the errors, because between the installation of `mas` and the execution, I'm not able to sign in to the Mac App Store, since the dsl only allows `tap`, `brew`,  `cask` or `mas`.
The only way to prevent this from happening is to manually sign in using the `App Store.app` before I start `brew bundle`.

I would propose to add another check to `Bundle::MacAppStoreInstaller.install()` if `mas` is signed in or not, because I guess the dsl is designed to not execute arbitrary commands.

MikeMcQuaid:
Can I see some example output? It sounds like erroring-out if you try to execute a DSL which requires `mas` is a good idea. Similarly, dumping should just ignore `mas` if it's not signed-in.

ffittschen:
Thanks for the quick reply!

Here is a minimal example with only `homebrew` and no formulas installed:
```
$ which mas
$ cat Brewfile 
mas '1Password', id: 443987910
$ brew bundle
Warning: Not signed in
Failed in installing 1Password

Success: 0 Fail: 1
``` 

My real `Brewfile` should install all apps, so it looks similar to this:
```
# Taps
tap 'caskroom/cask'
tap 'homebrew/bundle'
tap 'homebrew/services'

# Install fish shell
brew 'fish'

# Install binaries
brew 'mas'
brew 'rbenv'

# Development
brew 'carthage'

# Apps
cask 'alfred'
cask 'slack'
cask 'spotify'
cask 'sublime-text'
# omitted more Apps

# Quicklook
cask 'quicklook-json'
# omitted more Quicklook casks

# Install Mac App Store apps
mas '1Password', id: 443987910
mas 'Xcode', id: 497799835
# omitted more Mac App Store apps
```

In this example I would get `Success: 12 Fail: 2`, because I had no chance to sign in into the Mac App Store, so the two calls to `mas` will fail.

MikeMcQuaid:
What do you suggest happen in that case?

ffittschen:
[...]

PR #235, opened 2016-11-04
Add check to ensure mas is signed in (#233)

ffittschen:
This PR would resolve the issue #233.

When the user is not signed in to the Mac App Store, it will open the `App Store.app` and prompt for logging in instead of failing to install all apps specified for `mas` in the `Brewfile`.

Clarification of the chosen command:
`mas signin` requires to provide an Apple ID, which is not possible to provide at the time of the call. So I opted for `mas signin --dialog ''`. The `--dialog` option opens the login dialog of the `App Store.app` and passes the string after the argument to the email field of the dialog (therefore we pass an empty string)

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -25,6 +25,10 @@ module Bundle
     end
   end
 
+  def mas_signedin?
+    Bundle.system "mas", "account"
+  end
+
   def cask_installed?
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Missing a little coverage but approach looks good 👍

MikeMcQuaid:

PR #244, opened 2016-11-23
Add `--no-upgrade` option.

### LINKED ISSUE #229: Allow only installing missing formulae without upgrading existing ones
I'm a big fan of having a Brewfile in my home directory (and keeping it in version control) and using `brew bundle --global` to ensure I have all the utilities and apps I need. However, because my Brewfile lists formulae I use for development such as `node` and `go`, I don't want these to be auto-upgraded when I run `brew bundle` to fill in the blanks, i.e. install the new additions I made to my Brewfile. I only ever want formulae to be upgraded when I issue an explicit `brew upgrade`.

Now, I know about pinning and yes, I could easily `brew pin node go` to prevent them from being auto-upgraded. But, on principle, I kind of don't want `brew bundle --global` to ever upgrade *any* of the formulae. That means I would have to pin them all, which would make manual upgrades more clumsy since I'd have to go through an unpin-upgrade-repin dance.

I also know about `homebrew/versions`, and while I see it very useful for project-specific Brewfiles shared between a team of people, I feel it's an overkill for my own personal needs, where I simply want to install the latest version of e.g. `node` when I set up a fresh computer, but have control over when it's upgraded.

I've created [a workaround for myself](https://github.com/mislav/dotfiles/commit/7a05ec98779b25981c91526f19782aaaf274c183) that temporarily pins everything before `brew bundle`, and restores previously pinned on completion, but that makes me wish that `brew bundle` had some kind of `--no-upgrade` option.

Does this feature make sense for anyone else, or was `brew bundle` envisioned with a different set of principles in mind?

MikeMcQuaid:
As requested for people who want to avoid upgrading.

CC @mislav @jeremy

Fixes #229.

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -27,6 +27,7 @@ module Bundle
     def install_change_state!
       return false unless resolve_conflicts!
       if installed?
+        return true if ARGV.include?("--no-upgrade")
         upgrade!
       else
         install!
@@ -61,7 +62,9 @@ module Bundle
     end
 
     def self.formula_installed_and_up_to_date?(formula)
-      formula_installed?(formula) && !formula_upgradable?(formula)
+      return false unless formula_installed?(formula)
+      return true if ARGV.include?("--no-upgrade")
+      !formula_upgradable?(formula)
     end
 
     def self.formula_in_array?(formula, array)
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -1,7 +1,7 @@
 #:  * `bundle` <command>:
 #:    Bundler for non-Ruby dependencies from Homebrew.
 #:
-#:    `brew bundle` [-v|--verbose] [--file=<path>|--global]:
+#:    `brew bundle` [-v|--verbose] [--no-upgrade] [--file=<path>|--global]:
 #:    Install or upgrade all dependencies in a Brewfile.
 #:
 #:    `brew bundle dump` [--force] [--file=<path>|--global]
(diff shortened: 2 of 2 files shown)


lib/bundle/brew_installer.rb
```diff
@@ -61,7 +62,9 @@ def service_change_state!
     end
 
     def self.formula_installed_and_up_to_date?(formula)
-      formula_installed?(formula) && !formula_upgradable?(formula)
+      return false unless formula_installed?(formula)
+      return true if ARGV.include?("--no-upgrade")
```
mislav:
```rb
formula_installed?(formula) && (
  ARGV.include?("--no-upgrade") || !formula_upgradable?(formula)
)
```

MikeMcQuaid: I prefer the multiple return approach; I think it's easier to parse quickly.

javierjulio:
@MikeMcQuaid this is great, thanks! Just like @mislav I needed the same thing as most dependencies I have are important and I'd like to upgrade explicitly.

It is a little alarming now when using the `--no-upgrade` flag and seeing "Succeeded in installing..." in the output for each brew call when tap, cask and mas show "Skipping install of..." when the `-v` is given. Would you accept a PR to include this for brew installer?

MikeMcQuaid:
@javierjulio We based the output here on that of Bundler which behaves similarly. I'm not entirely sure what you're asking?

mislav:
@MikeMcQuaid Bundler says `Using foo 1.0.0` when not installing/upgrading, and `Installing foo …` when it's actually installing. I think there is value in output accurately describing what's being done.

MikeMcQuaid:
TIL, thanks @mislav. @javierjulio Yes, we'd accept a PR to say "Using" instead of "Installing" in that case.

Issue #175, opened 2016-05-02
Add 'context'/group type option


alias1:
It would be awesome if we could add a 'context' to the Brewfile, so a bunch of packages could be installed or not based on a flag. I'm thinking along the lines of [bundler groups](http://bundler.io/v1.12/groups.html).

```
brew "everywhere-tool" # goes into :default group

brew "different-style-personal-tool", :group => :personal
brew "different-style-2", :group => [:personal, :work]

group :personal do
  brew "personal-tool"
end

group :work do
  brew "work-tool"
end
```

```
brew bundle install --without personal
brew bundle install --only work # maybe
```

mikemcquaid:
Seems reasonable but we don't really accept feature requests. If you could try and submit a PR I'll review it and this seems like a feature we'll accept 👍

alias1:
Hokai, will see what I can do when I find some spare time :)

PR #248, opened 2016-11-27
Add support for bundle install --without=<groups>

umireon:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Modeled after Ruby bundler and implements #175 partially.
Having `--with` is also possible, but I believe that is too complex for Homebrew.

## Synopsis
In Brewfile:

```ruby
cask_args appdir: "/Applications"

tap "caskroom/cask"
brew "ruby187"
cask "vagrant"
mas 'CotEditor', id: 1024640650

group "development" do
  tap "brew/versions"
  cask "google-chrome"
  brew "vim"
end

group "xcode" do
  mas 'Xcode', id: 497799835
end
```

Everything but not in `development` and `xcode` group can be installed:

```
brew install --without=development,xcode
```

### DIFF
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -1,7 +1,7 @@
 #:  * `bundle` <command>:
 #:    Bundler for non-Ruby dependencies from Homebrew.
 #:
-#:    `brew bundle` [-v|--verbose] [--no-upgrade] [--file=<path>|--global]:
+#:    `brew bundle` [-v|--verbose] [--no-upgrade] [--file=<path>|--global] [--without=<groups>]:
 #:    Install or upgrade all dependencies in a Brewfile.
 #:
 #:    `brew bundle dump` [--force] [--file=<path>|--global]
@@ -28,6 +28,8 @@
 #:    `--file=-` to output to console).
 #:
 #:    If `--global` is passed, set Brewfile path to `$HOME/.Brewfile`.
+#:
+#:    If `--without=<groups>` is passed, skip installation in specified groups.
 #:
 #:    If `-h` or `--help` are passed, print this help message and exit.
 
--- a/lib/bundle/dsl.rb
+++ b/lib/bundle/dsl.rb
@@ -17,6 +17,8 @@ module Bundle
       @entries = []
       @cask_arguments = {}
 
+      @groups = %w[default]
+
       begin
         process
       # Want to catch all exceptions for e.g. syntax errors.
@@ -34,7 +36,12 @@ module Bundle
       success = 0
       failure = 0
 
-      @entries.each do |entry|
+      without_groups = ARGV.values(:without).to_a
+      selected_entries = @entries.select do |entry|
+        (without_groups & entry.options[:groups]).empty?
+      end
+
+      selected_entries.each do |entry|
         arg = [entry.name]
         verb = "installing"
         cls = case entry.type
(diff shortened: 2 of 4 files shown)


MikeMcQuaid:
This is cool but I'm trying to understand what the use-case is. For bundler it's e.g. production deployment vs. development but I'm not sure that makes sense here. Can you elaborate a bit? Thanks!

umireon:
Sorry for late reply.

In my case, I don't want to use TeX and utilities sometimes because they are very huge and take times to build.

```ruby
brew 'vim'
cask 'iterm2'
# continues...

group 'tex' do
  cask 'mactex'
  brew 'gnuplot', args: ['with-qt5', 'with-tex', 'with-cairo']
  brew 'ghostscript'
  # ... continues
end
```

Or not to install various browsers.
```ruby
cask 'google-chrome'

group 'additinal_browsers' do
  cask 'firefox'
  cask 'caskroom/versions/firefoxdeveloperedition'
  cask 'caskroom/versions/firefoxnightly'
  cask 'caskroom/versions/google-chrome-canary'
end
```

MikeMcQuaid:
@umireon Ok but why just not add these to your `Brewfile`?

umireon:
I am sorry that was unclear but I mean skipping unwanted 20 minutes installation time and 3 GB downloading by `brew bundle --without=tex`.

MikeMcQuaid:
Sure, I understand that but: why add them to your `Brewfile` in the first place? Is the `Brewfile` controlled by someone else?

umireon:
I don't want to install tex environment into VM but keep my `Brewfile` one file.

MikeMcQuaid:
Ok, so you're personally wanting to use the same `Brewfile` once for multiple deployment environments, I see now. I'm not strongly opposed to this but I'd like to leave it for a while and see if anyone else 👍 or 👎 this feature. Thanks!

neckhair:
Just a little input. I'd rather have something like

```ruby
host 'Macbook' do
  cask 'firefox'
end

host 'MyVM' do
  cask 'tex'
end
```

This would be much more convenient when you provision multiple Macs. You don't have to know anything about your Brewfile when doing a `brew bundle`.

MikeMcQuaid:
@neckhair That approach does not scale, I'm afraid.

MikeMcQuaid:
Passing on this due to no other 👍s. Thanks anyway @umireon and sorry.

Gaelan:
I'd like this because I want to share coding tools between my personal and work machines, but don't want to have personal software (games, etc) on my work machine.

MikeMcQuaid:
@Gaelan You could do this with two `Brewfile`s.

arunvelsriram:
@MikeMcQuaid Two `Brewfile`s for two machines. This will scale poorly if we have to manage more machines for ex. personal, work and work laptop for client-related work.

Also, there will be a lot of common packages which we have two duplicate in all `Brewfile`s. So I feel this feature of grouping the packages is quite useful.

MikeMcQuaid:
@arunvelsriram You can still do that with multiple `Brewfile`s.

PR #261, opened 2017-02-26
Add support for --global BREWFILE environment variable

### LINKED ISSUE #199: $BREWFILE environment variable
How about a `BREWFILE` environment variable as an alternative to `--file=`?

danielbayley:
Resolves #199.

### DIFF
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -28,6 +28,7 @@
 #:    `--file=-` to output to console).
 #:
 #:    If `--global` is passed, set Brewfile path to `$HOME/.Brewfile`.
+#:    A `$BREWFILE` environment variable can be set to override this file path.
 #:
 #:    If `-h` or `--help` are passed, print this help message and exit.
 
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -44,7 +44,7 @@ module Bundle
 
   def brewfile
     if ARGV.include?("--global")
-      file = Pathname.new("#{ENV["HOME"]}/.Brewfile")
+      file = Pathname.new(ENV["BREWFILE"] || "#{ENV["HOME"]}/.Brewfile")
     else
       filename = ARGV.value("file")
       filename = "/dev/stdin" if filename == "-"
--- a/lib/bundle/dumper.rb
+++ b/lib/bundle/dumper.rb
@@ -7,7 +7,7 @@ module Bundle
 
     def dump_brewfile
       if ARGV.include?("--global")
-        file = Pathname.new("#{ENV["HOME"]}/.Brewfile")
+        file = Pathname.new(ENV["BREWFILE"] || "#{ENV["HOME"]}/.Brewfile")
       else
         filename = ARGV.value("file")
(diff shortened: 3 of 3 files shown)


MikeMcQuaid:
Can you explain the use-case for this?

danielbayley:
The same use case for `--global` (#60), only without the implication of having to further clutter the `HOME` directory, which some people (😉) find annoying… when tools hard code that path.

MikeMcQuaid:
Where do you personally want to store this instead?

danielbayley:
Personally, my `Brewfile` lives in my [local] `$DOTFILES` repo, which I put in `~/Library/Application Support/dotfiles` (macOS user)… I could go into the whole [_`XDG` spec_](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) thing, but I'm sure you would have likely come across issues about it before…

MikeMcQuaid:
Ok. Will leave this open for a while to see if anyone else wants it as I'd rather wait and see if this would be used by more than just you.

> Personally, my Brewfile lives in my [local] $DOTFILES repo

Mine too, `~/.Brewfile` is just a symlink.

> I could go into the whole XDG spec thing, but I'm sure you would have likely come across issues about it before…

Do you set those variables yourself? Supporting them is an argument I'm more convinced by.

danielbayley:
> Do you set those variables yourself? Supporting them is an argument I'm more convinced by.

I do have 
~~~ sh
setenv () { launchctl setenv $@ && export $1=${@:2}}

setenv XDG_DATA_HOME "$HOME/Library/Application Support"
setenv XDG_CONFIG_HOME $XDG_DATA_HOME
setenv XDG_CACHE_HOME ~/Library/Caches
~~~
in my `$ZDOTDIR/.zshenv`

Personally, I prefer to use environment variables that I can keep all in one place than relying on symlinks everywhere if I can…

MikeMcQuaid:
@danielbayley Would you consider changing this PR to use the correct one of those instead? Thanks!

danielbayley:
So you mean something like `$XDG_CONFIG_HOME/.Brewfile`? That's still semi-hardcoding the path and forces you to have `.Brewfile` [hidden] in the root of wherever that is (so in this case that would force it into `~/Library/Application Support/.Brewfile` which is even less “dotfiles” friendly…

Just `BREWFILE` I think is a lot more flexible and makes more sense.

I came across a similar issue here also https://github.com/MikeMcQuaid/strap/issues/21, and [this project](https://github.com/rcmdnk/homebrew-file), which appears to introduce `$HOMEBREW_BREWFILE` 😕.

MikeMcQuaid:
> So you mean something like $XDG_CONFIG_HOME/.Brewfile?

Yes.

> Just BREWFILE I think is a lot more flexible and makes more sense.

I'm not convinced the flexibility is a good thing, I'm afraid.


> I came across a similar issue here also MikeMcQuaid/strap#21, and this project, which appears to introduce $HOMEBREW_BREWFILE 😕.

Perhaps that project is a better fit for your needs?

MikeMcQuaid:
Passing on this for now but others please chime in if you desperately want this feature and come across this closed PR.

PR #262, opened 2017-03-01
Allow passing boolean flag args like --force to cask install

maxfierke:
Sometimes you just want to bulldoze over existing packages 😛

### DIFF
--- a/lib/bundle/cask_installer.rb
+++ b/lib/bundle/cask_installer.rb
@@ -8,7 +8,15 @@ module Bundle
         return :skipped
       end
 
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
How does this work currently? Does it need added to every line of the `Brewfile`? I'm wondering if either a global DSL and/or a `brew bundle --force` would be better? Thoughts?

MikeMcQuaid:
I get it now! Seems reasonable! Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #264, opened 2017-04-05
Make `brew bundle cleanup` preserve dependencies.

MikeMcQuaid:
Always ensure that dependencies for specified formulae are kept on any cleanup operation. To make these more accurate use `runtime_dependencies` from the keg when available as these are based on the actual dependencies that were used by this actual install.


mivok:
So I'm not sure why right now, but this doesn't seem to be catching all dependencies. If you install graphviz, which depends on gd, which in turn depends on fontconfig, then run brew bundle cleanup (with graphviz in your Brewfile, but not gd or fontconfig), it wants to uninstall fontconfig (and fails if you add --force, because gd is installed) . Interestingly, fontconfig _does_ appear to be listed in the runtime_dependencies of gd.

MikeMcQuaid:
> So I'm not sure why right now, but this doesn't seem to be catching all dependencies.

Yeh, I relooked at the implementation and it'll only do a single level. I'll take another look, thanks.

MikeMcQuaid:
@mivok Should be addressed by https://github.com/Homebrew/homebrew-bundle/pull/266

mivok:
Yup, `brew bundle cleanup` runs clean for me with #266 applied. Thanks!

MikeMcQuaid:
@mivok Thanks for testing! Merged.

PR #278, opened 2017-07-09
CaskDumper uses cask full-name

wendorf:
Commands like `cleanup` and `check` would provide incorrect information
if a cask from a tap other than caskroom/cask was installed. While the
Brewfile might include the tap name (e.g. "caskroom/version/foo"),
CaskDumper would report the installed cask as "foo". Now that CaskDumper
provides the full name of the cask, comparisons between a Brewfile and
the installed state correctly consider the tap the cask is from.

This requires homebrew/brew#2878, which adds the flag, to be merged first.

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -8,7 +8,7 @@ module Bundle
 
     def casks
       @casks ||= if Bundle.cask_installed?
-        `brew cask list -1 2>/dev/null`.split("\n").map { |cask| cask.chomp " (!)" }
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Nice work! Once it's merged this looks good to go 👍

commitay:
@MikeMcQuaid Gentle ping on this, https://github.com/Homebrew/brew/pull/2878 has been merged.

MikeMcQuaid:
Thanks @wendorf!

PR #279, opened 2017-07-13
Check command will check all services that might need starting, instead of just the first

wendorf:
If two formulae have services that need to start and the first is
already started, the previous code would exit early from the method and
not check the second service. By switching to `next`, we properly
compare each service until one is found that needs to be started.

Note: The second commit is mostly for style's sake, and can be removed if requested. `Bundle::BrewInstaller#resolve_conflicts!` also has the potentially-confusing return from a block, but an early-exit is desirable there, so I've left it.

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -74,9 +74,7 @@ module Bundle
       @formula_info_name ||= {}
       @formula_info_name[name] ||= begin
         require "formula"
-        formula = Formula[name]
-        return {} unless formula
-        formula_inspector formula.to_hash
+        formula_inspector Formula[name].to_hash
       end
     end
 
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Nice catch! Need to add some additional tests to get the coverage up but otherwise 👍

wendorf:
When I tried to add the tests, I looked for a case in which the formula would be falsey and thus trigger the other branch... and it turns out it's not possible (unless I contrive a situation via stubs). Formulary will raise an exception instead of returning `nil`, so I just removed the branching.

Before this PR, the method didn't handle the exception either, so this feels correct to me.

MikeMcQuaid:
> When I tried to add the tests, I looked for a case in which the formula would be falsey and thus trigger the other branch... and it turns out it's not possible (unless I contrive a situation via stubs). Formulary will raise an exception instead of returning nil, so I just removed the branching.

This is definitely the way to go; thanks for taking a smart approach here rather than mindlessly stubbing. 

Thanks for your contribution to Homebrew Bundle, @wendorf! Without people like you submitting PRs we couldn't run this project. You rock!

PR #280, opened 2017-08-01
bundle check: give a rundown of what dependencies aren't satisfied

jeremy:
Useful to get a quick sense of "how unsatisfied" they are and what will
happen when we run `brew bundle install`.

Also useful for understanding when `--no-upgrade` says deps aren't
satisfied even when all formulae and casks are installed, such as
services that aren't started.

### DIFF
--- a/lib/bundle/commands/check.rb
+++ b/lib/bundle/commands/check.rb
@@ -13,12 +13,17 @@ module Bundle
       end
 
       def run
-        if any_taps_to_tap? ||
-           any_casks_to_install? ||
-           any_apps_to_install? ||
-           any_formulae_to_install? ||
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Hey @jeremy! The problem is that checking all these values (as opposed to the first one to return `true` is slower and we want `brew bundle check` to check as little as possible and be as quick as possible. I'd be 👍 with this functionality were it hidden behind a `--verbose` flag. Also, check out `brew style homebrew/bundle` which will check your `rubocop` style for you. Thanks!

jeremy:
Good call. A quiet/silent flag would be nice, too, so we don't have to `>/dev/null` when doing `brew bundle check || …`

Issue #282, opened 2017-08-12
brew-bundle: git-based time check can result in ruby exception if repo not found


JCount:
The time check at https://github.com/Homebrew/homebrew-bundle/blob/master/cmd/brew-bundle.rb#L38 can cause a ruby exception if a git repository cannot be immediately found during the semi-naive check.

See https://github.com/Homebrew/brew/issues/3045 for the original issue

Relevant code block:
```
# Homebrew version check
# commit cf71e30180d44219836ef129d5e5f00325210dfb
MIN_HOMEBREW_COMMIT_DATE = Time.parse "Wed Aug 17 11:07:17 2016 +0100"
HOMEBREW_REPOSITORY.cd do
  if MIN_HOMEBREW_COMMIT_DATE > Time.parse(`git show -s --format=%cD`)
    odie "Your Homebrew is outdated. Please run `brew update`."
  end
end
```

That `homebrew/bundle` was not tapped prior to the `brew bundle --global` call, leading to the code at https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.rb#L108-L116 being executed, were all likely contributing factors. As may have been the non-standard install location too.

The time check seems mainly aimed at verifying that the version of the vendored ruby is at least 2.0.0. At this point is there a potentially more robust way of merely introspecting the ruby's version at runtime? Or, would specifying the target git repository more explicitly, along with some exception handling code, make more sense?

PR #283, opened 2017-08-12
brew-bundle: improve HOMEBREW_VERSION check.

MikeMcQuaid:
Improve the `HOMEBREW_VERSION` check to be based on the global constant rather than the time of the HOMEBREW_REPOSITORY's last commit.

CC @JCount

Fixes https://github.com/Homebrew/brew/issues/3045
Fixes https://github.com/Homebrew/homebrew-bundle/issues/282


JCount:
Thanks @MikeMcQuaid for the quick fix! 😄

PR #293, opened 2017-10-06
Allow automatically linking formulae.

MikeMcQuaid:
For versioned formulae e.g. `mysql@5.6` it may be desirable to automatically link formulae after that are installed by Homebrew/bundle if you are only going to have one version of MySQL on a system and you wish it to be in your `PATH` by default.


ilovezfs:
Will `link: false` work for unlinking non-keg-only stuff?

MikeMcQuaid:
> Will link: false work for unlinking non-keg-only stuff?

Yep.

PR #297, opened 2017-10-27
add formulae descriptions (if available) when dumping Brewfile

phallstrom:
This patch makes it so that when one runs `brew bundle dump` it will
append the formulae's descriptions as comments to the generated
Brewfile.

Comments are padded to align with the longest "brew line."

For example:

    ...
    tap "homebrew/bundle"
    brew "libyaml"                       # YAML Parser
    brew "consul", restart_service: true # Tool for service discovery, monitoring and configuration
    brew "eot-utils"
    brew "pick"                          # Utility to choose one option from a set of choices
    ...

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -24,12 +24,16 @@ module Bundle
       requested_formula = formulae.select do |f|
         f[:installed_on_request?] || !f[:installed_as_dependency?]
       end
+      longest_brewline = 0
       requested_formula.map do |f|
         brewline = "brew \"#{f[:full_name]}\""
         args = f[:args].map { |arg| "\"#{arg}\"" }.sort.join(", ")
         brewline += ", args: [#{args}]" unless f[:args].empty?
         brewline += ", restart_service: true" if BrewServices.started?(f[:full_name])
-        brewline
+        longest_brewline = [longest_brewline, brewline.length].max
+        [brewline, f[:desc]]
+      end.map do |brewline, desc|
+        desc.to_s.empty? ? brewline : format("%-#{longest_brewline}s # %s", brewline, desc)
       end.join("\n")
     end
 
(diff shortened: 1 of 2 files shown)


phallstrom:
The specs are passing. Rubocop is annoyed. One I could agree with the other are line lengths due to the JSON test setup (one of those isn't even in my patch so maybe you don't care too much about that). Let me know if you want the others fixed up.

```
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb ==
C: 35:  7: Avoid multi-line chains of blocks.
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/spec/brew_dumper_spec.rb ==
C:  6:  1: Block has too many lines. [566/144]
C: 81:  3: Block has too many lines. [196/144]
C:252:  7: Redundant curly braces around a hash parameter.
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/spec/brew_installer_spec.rb ==
C:  5:  1: Block has too many lines. [242/144]
```

MikeMcQuaid:
@phallstrom Yeh, please fix them all up, thanks.

phallstrom:
@MikeMcQuaid I didn't look too closely at the rubocop errors, but the 'too many lines' ones are complaining about Rspecs' "describe do" block. Doesn't seem right to break those up merely to satisfy rubocop. But I also see you aren't excluding any of them with inline disabling of rubocop. Makes me think this rubocop thing is relatively new.  Please advise...

MikeMcQuaid:
@phallstrom As I think about this more I think it should probably be an optional parameter to `brew dump`. What do you think?

MikeMcQuaid:
@phallstrom Any thoughts?

phallstrom:
Sorry @MikeMcQuaid, I was out of town. I'll admit to being new to using this tool, but kind of feel it should be the default.  I'm basing that off the assumption that if you're already setup with it you're going to manage the file yourself. If you're not, you probably want to dump it to get started and if you're like me have a million things you can't remember what they are. Not sure I'd look to see if there was a flag for it.  But you could certainly convince me that's the right way :)

MikeMcQuaid:
@phallstrom I think it's better to not change the format of everyone's Brewfile's and e.g. Gemfiles and similar do not have this behaviour.

PR #305, opened 2017-11-18
Do not uninstall dependencies for custom formulae

### LINKED ISSUE #301: `brew bundle cleanup` wants to clean up dependencies of formulae from custom taps
If `vitorgalvao/tiny-scripts/cask-repair` is in your Brewfile, `hr` will be installed because it is a dependency. However, if you then `brew bundle cleanup --force`, `hr` will be uninstalled.

This is because `Bundle::Commands::Cleanup#formulae_to_install` constructs `kept_formulae` and `current_formulae` differently. `kept_formulae[i][:name]` is the formula's `full_name`, but `current_formulae[i][:name]` is the "short" `name`. When the two names are compared in `#recursive_dependencies`'s `current_formulae.find { |f| f[:name] == name }`, the names do not match and the dependency is not found.

wendorf:
When searching `current_formulae` for a formula, it was not found if the
formula was from a custom tap and had a `full_name` different from its
`name`. This caused all of its dependencies to be missed.

Fixes #301

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -79,7 +79,7 @@ module Bundle
 
         formulae_names.each do |name|
           next if @checked_formulae_names.include?(name)
-          formula = current_formulae.find { |f| f[:name] == name }
+          formula = current_formulae.find { |f| f[:full_name] == name }
           next unless formula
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Great work again @wendorf, you rock!

PR #306, opened 2017-11-19
Only clean up build dependencies for bottles

wendorf:
If a formula has not been poured from a bottle, its build dependency will be used for upgrades and should be kept around.

There's an argument for getting rid of build dependencies during a cleanup (they're not needed for runtime), but it was unexpected for me. Upgrading a formula built from source is very likely, so it seems like a good idea to keep build dependencies around.

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -82,6 +82,10 @@ module Bundle
           formula = current_formulae.find { |f| f[:full_name] == name }
           next unless formula
           f_deps = formula[:dependencies]
+          unless formula[:poured_from_bottle?]
+            f_deps += formula[:build_dependencies]
+            f_deps.uniq!
+          end
           next unless f_deps
           next if f_deps.empty?
           @checked_formulae_names << name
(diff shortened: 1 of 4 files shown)


lib/bundle/commands/cleanup.rb
```diff
@@ -82,6 +82,9 @@ def recursive_dependencies(current_formulae, formulae_names, top_level = true)
           formula = current_formulae.find { |f| f[:full_name] == name }
           next unless formula
           f_deps = formula[:dependencies]
+          unless formula[:poured_from_bottle?]
+            f_deps = (f_deps + formula[:build_dependencies]).uniq
```
MikeMcQuaid:
```ruby
f_deps += formula[:build_dependencies]
f_deps.uniq!
```
How about that?

MikeMcQuaid:
This makes a lot of sense to me 👍. One comment; let me know your thoughts (and update if you agree) and I'll merge tomorrow. Thanks again, more fantastic work @wendorf! I hope to see more of you!

wendorf:
Your comment makes sense—just pushed the change now. Thanks for taking a look!

MikeMcQuaid:
Thanks again @wendorf, doing great work here!

PR #314, opened 2017-12-26
Add init command

kyoshidajp:
Sometimes, I forget a name `Brewfile` and how to write dependencies.

New `init` command which like a `bundle init` make user can write contents unconsciously.

# Usage

```
brew bundle init
```

Will generate `Brewfile` with commented general dependencies.

# Other

To add this feature, refactored `Bundle.brewfile`.

Up until this PR the method returns dependencies of Brewfile. Now that it returns Brewfile path. And added `Bundle.read_brewfile` returns Brewfile dependencies.


MikeMcQuaid:
Thanks for the PR! I'm not sure I understand what this feature adds over `touch Brewfile` or `brew bundle dump` (which creates the same file without comments, right?). Can you explain? Thanks again!

kyoshidajp:
My bad, I didn't notice that `brew bundle dump` meets this situation.

Thanks.

MikeMcQuaid:
No worries and thanks anyway!

PR #324, opened 2018-02-07
Adds --describe option to brew bundle dump

### LINKED ISSUE #322: Optionally output package description as comment
I'd like to have a Brewfile that has only the formulae that I've intended to install, no dependencies. Since Homebrew doesn't seem to track† if a formula was installed manually or as a dependency, I'm left to manually edit the output of `brew bundle dump`. It's a slog because I've got more than 350 packages installed but I'll deal.

What could really expedite my discovery of what some formulae are is including as a comment on the line or above it in the Brewfile output. That way, I don't have to do `brew info packagename` for every package I don't recognize, which is a lot since my Homebrew installation dates back a few years. Honestly, I've found that every Brewfile I've manually created for the last couple of months – something I've been dropping into every repo I touch at work – has had a comment describing what the package is or does.

I'm willing to put the time into developing this feature. I envision it as an addition [here](https://github.com/Homebrew/homebrew-bundle/blob/master/lib/bundle/brew_dumper.rb#L123)
 and [here](https://github.com/Homebrew/homebrew-bundle/blob/master/lib/bundle/brew_dumper.rb#L32). It would be enabled through a command line option to the `dump` command, probably something like `--comment` or `--describe`. There's some complexity in that it appears that there are not currently arguments passed down into the dumper.

_† One thing I noted is that maybe this assumption is out of date, given what I see [here](https://github.com/Homebrew/homebrew-bundle/blob/master/lib/bundle/brew_dumper.rb#L25): `f[:installed_on_request?] || !f[:installed_as_dependency?]`_

colindean:
This option outputs above each dependency a comment containing a description when the dependency has a description associated with it.

Right now, only Homebrew formulae have descriptions. Casks, taps, and mas dependencies do not have descriptions available.

Fixes #322

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -25,7 +25,9 @@ module Bundle
         f[:installed_on_request?] || !f[:installed_as_dependency?]
       end
       requested_formula.map do |f|
-        brewline = "brew \"#{f[:full_name]}\""
+        brewline = ""
+        brewline += "# #{f[:desc]}\n" if ARGV.include?("--describe")
+        brewline += "brew \"#{f[:full_name]}\""
         args = f[:args].map { |arg| "\"#{arg}\"" }.sort.join(", ")
         brewline += ", args: [#{args}]" unless f[:args].empty?
         brewline += ", restart_service: true" if BrewServices.started?(f[:full_name])
@@ -121,6 +123,7 @@ module Bundle
 
       {
         name: f["name"],
+        desc: f["desc"],
         oldname: f["oldname"],
         full_name: f["full_name"],
         aliases: f["aliases"],
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -26,7 +26,7 @@ module Bundle
       end
       requested_formula.map do |f|
         brewline = ""
-        brewline += "# #{f[:desc]}\n" if ARGV.include?("--describe")
+        brewline += "# #{f[:desc]}\n" if ARGV.include?("--describe") && f[:desc]
         brewline += "brew \"#{f[:full_name]}\""
         args = f[:args].map { |arg| "\"#{arg}\"" }.sort.join(", ")
         brewline += ", args: [#{args}]" unless f[:args].empty?
--- a/spec/brew_dumper_spec.rb
+++ b/spec/brew_dumper_spec.rb
@@ -59,6 +59,7 @@ describe Bundle::BrewDumper do
       expect(subject.formulae).to contain_exactly(
         name: "foo",
         full_name: "homebrew/tap/foo",
+        desc: "",
         oldname: nil,
         aliases: [],
         args: [],
@@ -183,6 +184,7 @@ describe Bundle::BrewDumper do
         {
           name: "foo",
           full_name: "homebrew/tap/foo",
+          desc: "",
           oldname: nil,
           aliases: [],
           args: [],
@@ -202,6 +204,7 @@ describe Bundle::BrewDumper do
         },
         name: "bar",
         full_name: "bar",
+        desc: "",
         oldname: nil,
         aliases: [],
         args: ["with-a", "with-b"],
--- a/spec/brew_dumper_spec.rb
+++ b/spec/brew_dumper_spec.rb
@@ -482,7 +482,7 @@ describe Bundle::BrewDumper do
 
   context "when --describe is not set" do
     before do
-      ARGV.delete("--describe") if ARGV.include?("--describe")
+      stub_const("ARGV", [])
       Bundle::BrewDumper.reset!
       allow(Bundle::BrewDumper).to receive(:formulae_info).and_return [
         {
@@ -518,6 +518,7 @@ describe Bundle::BrewDumper do
         },
       ]
     end
+
     it "does not output a comment with dependency description" do
       expect(Bundle::BrewDumper.dump).not_to include("#")
     end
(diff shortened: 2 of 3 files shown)


colindean:
Working on a test for this.

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
     end
   end
 
+  context "when --describe is not set" do
+    before do
+      ARGV.delete("--describe") if ARGV.include?("--describe")
```
MikeMcQuaid: Use `stub_const("ARGV", [])` instead.

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
     end
   end
 
+  context "when --describe is not set" do
```
MikeMcQuaid: I wonder if this case is needed just to test `#` isn't added?

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
...
+      ]
+    end
+    it "does not output a comment with dependency description" do
+      expect(Bundle::BrewDumper.dump).not_to include("#")
+    end
+  end
+
+  context "when --describe is set" do
+    before do
+      ARGV << "--describe" unless ARGV.include?("--describe")
```
MikeMcQuaid: Use `stub_const("ARGV", ["--describe"])` instead

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
...
+          build_dependencies: [],
+          requirements: [],
+          conflicts_with: [],
+          pinned?: false,
+          outdated?: false,
+        },
+      ]
+    end
+    after do
+      ARGV.delete("--describe") if ARGV.include?("--describe")
```
MikeMcQuaid: Can remove this with the `stub` addition.

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
...
+          optional_dependencies: [],
+          build_dependencies: [],
+          requirements: [],
+          conflicts_with: [],
+          pinned?: false,
+          outdated?: false,
+        },
+      ]
+    end
+    it "does not output a comment with dependency description" do
```
MikeMcQuaid: Add a newline before each `it`

spec/brew_dumper_spec.rb
```diff
@@ -477,6 +480,99 @@
...
+          conflicts_with: [],
+          pinned?: false,
+          outdated?: false,
+        },
+      ]
+    end
+    after do
+      ARGV.delete("--describe") if ARGV.include?("--describe")
+    end
+    it "outputs a comment on the line before a dependency with a description" do
```
MikeMcQuaid: Add a newline before each `it`

lib/bundle/brew_dumper.rb
```diff
@@ -25,7 +25,9 @@ def dump
         f[:installed_on_request?] || !f[:installed_as_dependency?]
       end
       requested_formula.map do |f|
-        brewline = "brew \"#{f[:full_name]}\""
+        brewline = ""
+        brewline += "# #{f[:desc]}\n" if ARGV.include?("--describe") && f[:desc]
```
MikeMcQuaid: Do you think it's preferably to display this comment above vs. on the same line to the right of the name? Don't feel strongly but figured it's worth thinking about.

MikeMcQuaid:
Thanks for your work here, @colindean!

lib/bundle/brew_dumper.rb
colindean: Above felt better for reading.

```
# this is some description
brew 'asdf'
# a tool for describing
brew 'sdfgsdfsdfdadfs'
# yet another great tool everyone installs
brew 'wertwert'
```

versus

```
brew 'asdf' # this is some description
brew 'sdfgsdfsdfdadfs' # a tool for describing
brew 'wertwert' # yet another great tool everyone installs
```

The latter is harder to skim by a lot, even when syntax highlighting is present.

spec/brew_dumper_spec.rb
colindean: Eh, it was a TTD test that I left in. It's ensuring that there is no description when the user hasn't requested it. I don't see the harm in leaving it in other than that it extends tests by about 0.7 seconds (`time rspec spec/brew_dumper_spec.rb`).

colindean:
Easy comments addressed. Let me know what you'd like for that one test.

lib/bundle/brew_dumper.rb
MikeMcQuaid: The only thought reading that is you may want an additional newline after every described line to improve the grouping?

spec/brew_dumper_spec.rb
MikeMcQuaid: Ok, cool 👍

lib/bundle/brew_dumper.rb
colindean: I'd thought about that. I played with it a little. I can't decide on _where_ I want to put the newlines. I can insert this into `brew_dumper.rb` on line 35 as of the latest commit and I get reasonable output.

```ruby
brewline += "\n" if ARGV.include?("--describe") && f[:desc]
```

If I omit the `&& f[:desc]`, then _all_ brew lines get an extra newline, regardless of if they have a  description. This looks more consistent.

Honestly, the inconsistency is what makes me not want to have the extra newline.

colindean:
Fixed the style error.

MikeMcQuaid:
Thanks again @colindean!

colindean:
🎉

PR #325, opened 2018-02-07
Sets minimum rspec version to 3.5.0

colindean:
I encountered an error saying that `filter_run_when_matching` was missing on the RSpec config object when running specs for #324. It turns out that this method was added in rspec 3.5.0 and I had an older version of RSpec installed and that older version met the requirement in the Gemfile of "any version!"

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,44 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    codecov (0.1.10)
+      json
+      simplecov
+      url
+    diff-lcs (1.3)
+    docile (1.1.5)
+    json (1.8.3)
+    rake (10.4.2)
+    rspec (3.7.0)
+      rspec-core (~> 3.7.0)
+      rspec-expectations (~> 3.7.0)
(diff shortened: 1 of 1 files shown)


colindean:
This is the commit in which `filter_run_when_matching` is added, which was added in rspec 3.5.0: https://github.com/rspec/rspec-core/commit/7a238306fdcc693e16dcd430191f881c8c26db30

MikeMcQuaid:
@colindean What do you think about just checking in `Gemfile.lock` instead? If you're 👍: mind updating this PR to do that? Thanks!

colindean:
If that’s the best practice for ensuring a the minimum dependency version, sure. 

When I _add_ the Gemfile.lock, should remove the version specifier in the Gemfile or not?

MikeMcQuaid:
> should remove the version specifier in the Gemfile or not?

Yes, please, thanks!

colindean:
Updated!

MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @colindean!

colindean:
🎉

PR #326, opened 2018-02-07
Adds a `list` command that shows all dependencies in the Brewfile

colindean:
By default, only brew-type dependencies are output because only brew commands can really use the output of this by default. See the usage in cmd/brew-bundle.rb for the options that choose which types to output. They do compound, so --taps --casks shows both of those.

This came out of [suggestion from @ilovezfs](https://github.com/Homebrew/homebrew-bundle/issues/322#issuecomment-363389093) on #322.


colindean:
This needs tests. Out of steam for the night.

colindean:
Test failure was dip in coverage.

colindean:
Tests say that this is ready and I think it's ready so… ready!

lib/bundle/installer.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Installer
```
MikeMcQuaid: Why the new class?

MikeMcQuaid:
Nice work so far!

lib/bundle/installer.rb
colindean: I extracted it from [here](https://github.com/Homebrew/homebrew-bundle/pull/326/commits/fbe0b9eaec85f19a78f05a69234ee324ed9640bc#diff-d485b64f0c3589a2a3afb078040dde93L35). I must have forgotten to add it on that commit. Ugh. Got it here, though.

Moving it to a separate class seemed appropriate given the existence of other tool classes like this: Dumper and XDumper were a model, where X is one of the types. There was XInstaller but not an Installer. Conforming to that setup made the installer parts fall in line and, most importantly, removed installation from the responsibility of the DSL parser. I felt that the DSL parser should just turn the DSL into some model that can be acted on later: the entries.

MikeMcQuaid: Ok, seems reasonable in this case. In future: can you split similar refactorings into a separate PR? It makes it harder to review when in an unrelated one. Thanks!

cmd/brew-bundle.rb
```diff
@@ -16,6 +16,10 @@
 #:    `brew bundle exec` [command]
 #:    Run an external command in an isolated build environment.
 #:
+#:    `brew bundle list` [--all|--brews|--casks|--taps|--mas] [--file=<path>|--global]
```
MikeMcQuaid: Can you add some docs into the `README.md` too?

colindean: Done in 8130afb

lib/bundle/installer.rb
colindean: Will do. It was the first thing I did wording toward this and then it ended up not being 100% necessary.

MikeMcQuaid:
Thanks again @colindean, nice work!

colindean:
🎉

PR #327, opened 2018-02-11
Display brew bundle check failures instead of silently failing

### LINKED ISSUE #320: More meaningful `brew bundle check` outputs
When I run `brew bundle check --verbose`, the output just says something failed without any detailed information.
```
$ brew bundle check --verbose
Kernel.exit
Error: Kernel.exit
```
This even happens to newly dumped Brewfile.
```
⋊> brew bundle dump --verbose
⋊> brew bundle check --verbose
Kernel.exit
Error: Kernel.exit
```

Anyone has ideas help to track where `check` goes wrong?

p.s. I have tried `brew bundle install` and `brew bundle cleanup`, both return success. Yet `check` still fails.

colindean:
Changes the output to this:

```
↳ Cask asdfasdf needs to be installed or updated.
↳ Formula fdsafda needs to be installed or updated.
↳ App Slack needs to be installed or updated.
```

Formatting suggestions welcome, including recommendations on enhancing this with some color.

Fixes #320.


dalemyers:
This is a great step forward and make my life significantly easier. Thanks for putting in the work on this! 

However, since no form of lockfile is supported, I think it would be more useful if it differentiated between formulae which need to be installed, and those which need to be upgraded. Even more useful would be to have the ability to see the version differences. Currently, if a formulae has a new version published, the check fails. Not particularly deterministic. Seeing the versions would allow us to ignore these changes without having to depend on a more involved system such as vagrant.

dalemyers:
> you will have everyone who runs brew bundle on a random selection of versions
That's only a problem for some scenarios. 

If the output says "Package X is out of date (installed: 1.0.0, current: 1.0.1)" that's useful for at a glance decisions of whether updating is required or not. Furthermore, it can be used in scripts. It may be that there are 20 formulae mentioned in the `Brewfile`, but we only care about 2 of them staying up to date. On a "server", `brew bundle check` fails if there is a change. By being able to ignore it for those 18 other packages, we can filter out noise and only pay attention to what we care about. As it is, Homebrew doesn't give a choice.

MikeMcQuaid:
> That's only a problem for some scenarios.

I'm afraid I fundamentally disagree and am not interested with functionality being added to this tool that encourages people to sit on old, unsupported and insecure versions. If you wish to do that: you can create your own tap.

MikeMcQuaid:
@colindean Sorry if I wasn't clearer before but this should only display the output if e.g. `--verbose` is passed. The existing code is intentionally lazy and exits as soon as it finds a single failure so it's as fast as possible for scripts. This will necessitate almost a separate implementation for `--verbose` where it's not using `any?` or `||` but instead (like you've done here) calculates every single failure and outputs them.

MikeMcQuaid:
Thanks for all your work so far here, though! ❤️

colindean:
Ah, I think I understand how my changes altered the behavior. The `any_xxx?` methods were short-circuiting and the `any?` calls within return true upon encountering a single true block return.

I'll have to think about how I can redesign this. Might be a few days before I have time.

colindean:
I made it better.

I need to write tests for the two new flags.

Readme.md
```diff
@@ -70,6 +70,10 @@ You can check there's anything to install/upgrade in the `Brewfile` by running:
 
 This provides a successful exit code if everything is up-to-date so is useful for scripting.
 
+For a list of dependencies that are missing, pass `--verbose`.
+
+To check all dependencies by not exiting on the first missing dependency category, pass `--all`.
```
MikeMcQuaid: I reckon just make this implied by `--verbose` as otherwise just `--all` will produce the same output but take longer to run.

lib/bundle/commands/check.rb
```diff
@@ -5,6 +5,8 @@ module Commands
     module Check
       module_function
 
+      @arrow = "\u21B3".encode("utf-8")
```
MikeMcQuaid: Use an ASCII equivalent for this.

lib/bundle/commands/check.rb
```diff
@@ -14,50 +16,78 @@ def reset!
...
         end
       end
 
-      def any_casks_to_install?
-        @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
+      def casks_to_install
         requested_casks = @dsl.entries.select { |e| e.type == :cask }.map(&:name)
-        requested_casks.any? do |c|
-          !Bundle::CaskInstaller.cask_installed_and_up_to_date?(c)
+        actionable = requested_casks.reject do |c|
```
MikeMcQuaid: Probably want to pass through `enumerator` and use it here (and below) so you can still get the `any?` short circuit behaviour

lib/bundle/commands/check.rb
```diff
@@ -14,50 +16,78 @@ def reset!
...
+        checks = {
+          taps_to_tap: "Taps",
+          casks_to_install: "Casks",
+          apps_to_install: "Apps",
+          formulae_to_install: "Formulae",
+        }
+        completed_checks = []
+        errors = []
+        enumerator = exit_on_first_error? ? :any? : :each
+        work_to_be_done = checks.keys.send(enumerator) do |check_method|
```
MikeMcQuaid: Maybe make `check_method_names = checks.keys` above here to be a little more explicit/readable?

MikeMcQuaid:
@colindean Any news on this one? Thanks again!

colindean:
I've not had a chance. I might next week.

lib/bundle/commands/check.rb
colindean: Ohh, you'd like to still see the short circuit _within each category_, not just at the category level. I think I understand now.

colindean: I started passing the enumerator through, but it got ugly fast. The check methods would return either a boolean or an array: `any?` exists on arrays but not on booleans. I'll have to refactor the bodies of the checks to separate the errors from the return value like I did for the `work_to_be_done` loop.

Maybe tomorrow night?

MikeMcQuaid: @colindean Exactly, thanks! Basically: as soon as anything is missing and we're not `--verbose`: stop checking.

colindean: I'm getting into some major rework on this but it's making the code a lot more readable.

colindean:
I'm almost there. I've got one test failure that I've not really looked into yet:

```
Failures:

  1) Bundle::Commands::Check when service is not started should not raises error by default
     Failure/Error: expect { do_check }.not_to raise_error

       expected no Exception, got #<SystemExit: exit> with backtrace:
         # ./lib/bundle/commands/check.rb:59:in `exit'
         # ./lib/bundle/commands/check.rb:59:in `run'
         # ./spec/check_command_spec.rb:7:in `do_check'
         # ./spec/check_command_spec.rb:90:in `block (4 levels) in <top (required)>'
         # ./spec/check_command_spec.rb:90:in `block (3 levels) in <top (required)>'
         # ./spec/spec_helper.rb:29:in `block (3 levels) in <top (required)>'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `block in with_clean_env'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:540:in `with_env'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `with_clean_env'
         # ./spec/spec_helper.rb:29:in `block (2 levels) in <top (required)>'
     # ./spec/check_command_spec.rb:90:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:29:in `block (3 levels) in <top (required)>'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `block in with_clean_env'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:540:in `with_env'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `with_clean_env'
     # ./spec/spec_helper.rb:29:in `block (2 levels) in <top (required)>'
```

I've also got a lot of style cleaning to do.

colindean:
Style cleaning addressed…

…but the test still fails.

I'll look at it over the weekend.

colindean:
I cannot for the life of me figure out how [this test](https://github.com/Homebrew/homebrew-bundle/pull/327/files#diff-fb45c9ca25640e16644dde1ec796a93dR88) is failing.

I can't figure out how to get rspec to show me the stdout output so I changed all of that to just write to a file within the errors printing part of `Bundle::Commands::Check#run`

    here!brew bundle can't satisfy your Brewfile's dependencies.Entry(cask abc {:args=>{}})Taps were checked.Casks were checked.Apps were checked.Formulae were checked.→ Cask abc needs to be installed or updated.Satisfy missing dependencies with `brew bundle install`.

How is a _cask_ getting added in there? [This line](https://github.com/Homebrew/homebrew-bundle/pull/327/files#diff-fb45c9ca25640e16644dde1ec796a93dR80) is explicitly telling the cask dumper that there aren't any casks to be returned.

[This](https://github.com/Homebrew/homebrew-bundle/pull/327/files#diff-fb45c9ca25640e16644dde1ec796a93dR43) is the only place that a cask 'abc' is getting added at all. I'm under the impression that [this `reset!`](https://github.com/Homebrew/homebrew-bundle/pull/327/files#diff-fb45c9ca25640e16644dde1ec796a93dR11) is functioning correctly at all times. If put that `reset!` into the failing test, the test still fails.

I'm grasping at straws debugging here and could use some advice how to debug this more efficiently. My next step is trying to remember how to use `ruby-debug`, which I've only ever used once…

MikeMcQuaid:
@colindean I'll take a look when I get a chance. Thanks for digging in already 🙇

lib/bundle/commands/check.rb
```diff
@@ -14,50 +16,66 @@ def reset!
...
+          apps_to_install: "Apps",
+          formulae_to_install: "Formulae",
+        }
+        check_method_names = checks.keys
+
+        completed_checks = []
+        errors = []
+        enumerator = exit_on_first_error? ? :any? : :each
+
+        work_to_be_done = check_method_names.send(enumerator) do |check_method|
```
MikeMcQuaid: When this is `.send(:each)` it's not going to return `any_errors` but it'll just be `check_method_names`. That's the cause of the test failure.

MikeMcQuaid:
@colindean FYI this is what I used for debugging this:

```patch
diff --git a/lib/bundle/commands/check.rb b/lib/bundle/commands/check.rb
index 4d574fc..5fdec62 100644
--- a/lib/bundle/commands/check.rb
+++ b/lib/bundle/commands/check.rb
@@ -37,7 +37,7 @@ module Bundle
 
         completed_checks = []
         errors = []
-        enumerator = exit_on_first_error? ? :any? : :each
+        enumerator = exit_on_first_error? ? :any? : :all?
 
         work_to_be_done = check_method_names.send(enumerator) do |check_method|
           check_errors = send(check_method)
@@ -54,6 +54,8 @@ module Bundle
           unchecked_checks.each { |unchecked| puts "#{checks[unchecked]} were not checked." }
           if output_errors?
             errors.each { |package| puts "#{@arrow} #{package}" }
+            p work_to_be_done
+            p any_formulae_to_start?
           end
           puts "Satisfy missing dependencies with `brew bundle install`."
           exit 1
diff --git a/spec/check_command_spec.rb b/spec/check_command_spec.rb
index 3045d32..3957df0 100644
--- a/spec/check_command_spec.rb
+++ b/spec/check_command_spec.rb
@@ -85,7 +85,7 @@ describe Bundle::Commands::Check do
       allow(Bundle::BrewServices).to receive(:started?).with("def").and_return(false)
     end
 
-    it "should not raises error by default" do
+    it "should not raises error by default", :focus do
       allow_any_instance_of(Pathname).to receive(:read).and_return("brew 'abc'")
       expect { do_check }.not_to raise_error
     end
```

Will wait until tests are ✅ (including 100% coverage) and then will do a style review pass (just to prepare you: as there's some style changes I think should be made but there's not point getting into that before this is otherwise working).

Thanks again!

colindean:
I'm revisiting this. It took me a bit tonight to remember where I'd left off but I think I've got it back in mind for some time this weekend.

colindean:
I got it to pass by changing the `:each` to a `:map` but that I had to do `work_to_be_done = work_to_be_done.any? if work_to_be_done.class == Array` thereafter to make it all pass tells me that there is a poor design decision somewhere. 

I need to remove my debugging code and then I'll push it up as-is for some discussion on how I could make it better. I'll do that tomorrow at the earliest!

colindean:
Existing tests are passing and I'm fully rebased. I'm looking at the coverage report now.

colindean:
Coverage drop is from the newly refactored `exit_early_check` in `BrewChecker`, `CaskChecker`, and `MacAppStoreChecker`. I think it should be reasonably easy to contrive a test for that!

colindean:
┏(･o･)┛♪  ｔｅｓｔｓ　ａｒｅ　ｐａｓｓｉｎｇ　ｎｏｗ  ♪┗ (･o･) ┓
ƪ(˘⌣˘)┐ ａ　ｖｉｃｔｏｒｉｏｕｓ　ｈａｉｋｕ ┌(˘⌣˘)ʃ
〜(￣▽￣〜) ｎｏｗ　ｆｏｒ　ｓｔｙｌｅ　ｗｏｒｋ (〜￣▽￣)〜

lib/bundle/brew_checker.rb
```diff
@@ -0,0 +1,39 @@
...
+      if work_to_be_done
+        Bundle::Checker.action_required_for(last_checked)
+      else
+        Bundle::Checker::NO_ACTION
+      end
+    end
+
+    def full_check(formulae)
+      actionable = formulae.reject { |f| ok? f }
+      actionable.map { |entry| "Formula #{entry} needs to be installed or updated." }
```
MikeMcQuaid: Would be nice to chain:
```ruby
formulae.reject { |f| ok? f }
        .map ...
```

lib/bundle/brew_checker.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def find_actionable(entries)
+      requested_formulae = entries.select { |e| e.type == :brew }.map(&:name)
```
MikeMcQuaid: `.map` on a new line

lib/bundle/checker.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Checker
+    module_function
+
+    NO_ACTION = [].freeze
```
MikeMcQuaid: Not sure I understand why this is an empty array rather than e.g. `nil` or `false`?

lib/bundle/checker.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def action_required_for(formula)
+      [formula]
```
MikeMcQuaid: How is this an action?

lib/bundle/commands/check.rb
```diff
@@ -5,6 +5,8 @@ module Commands
     module Check
       module_function
 
+      @arrow = "→"
```
MikeMcQuaid: A frozen constant might be nicer?

lib/bundle/commands/check.rb
```diff
@@ -14,54 +16,75 @@ def reset!
         Bundle::BrewServices.reset!
       end
 
+      def exit_on_first_error?
+        Bundle::Checker.exit_on_first_error?
+      end
+
+      def output_errors?
+        ARGV.include?("--verbose")
```
MikeMcQuaid: Feels weird to have `exit_on_first_error?` call a method elsewhere and just be the inverse of this?

lib/bundle/commands/check.rb
```diff
@@ -14,54 +16,75 @@ def reset!
...
+          errors.concat(check_errors) if any_errors
+          any_errors
+        end
+
+        work_to_be_done = work_to_be_done.any? if work_to_be_done.class == Array
+
+        if work_to_be_done
           puts "brew bundle can't satisfy your Brewfile's dependencies."
+          unchecked_checks = (checks.keys - completed_checks)
+          completed_checks.each { |checked| puts "#{checks[checked]} were checked." }
```
MikeMcQuaid: Let's not output these in non-verbose mode.

lib/bundle/commands/check.rb
```diff
@@ -14,54 +16,75 @@ def reset!
...
+          unchecked_checks.each { |unchecked| puts "#{checks[unchecked]} were not checked." }
+          if output_errors?
+            errors.each { |package| puts "#{@arrow} #{package}" }
+          end
           puts "Satisfy missing dependencies with `brew bundle install`."
           exit 1
+        elsif any_formulae_to_start?
+          puts "brew bundle can't satisfy your Brewfile's dependencies."
+          puts "At least one brew formula must be started."
```
MikeMcQuaid: I'd probably rather the existing wording be changed to be more ambiguous and encapsulate this.

lib/bundle/mac_app_store_checker.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Bundle
+  module MacAppStoreChecker
```
MikeMcQuaid: These modules all look very similar. Would it be possible to DRY them up at all? I'm happy to give it a go myself if that'd be useful.

lib/bundle/brew_checker.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Bundle
+  module BrewChecker
+    module_function
+
+    def ok?(formula)
```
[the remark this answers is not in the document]
colindean: ✅

lib/bundle/commands/check.rb
colindean: ✅

lib/bundle/checker.rb
colindean: It's convenient for the return of all of the `find_actionable` methods to be a list, that way the list can convey what was there or that there was nothing using the same type, enabling use cases for both the early exit and the full check. Full check benefits more from having the check result be an array since it won't stop on the first missing package, which could instead be represented as a simple string.

Having this be written out enables a more readable

```ruby
def exit_early_check(formulae)
     work_to_be_done = …
     if work_to_be_done
        Bundle::Checker.action_required_for(last_checked)
      else
        Bundle::Checker::NO_ACTION
      end
end
```

which would otherwise be a rather confusing

```ruby
def exit_early_check(formulae)
     work_to_be_done = …
     if work_to_be_done
        [last_checked]
      else
        []
      end
end
```

which is what I had before my brain said, "I can't remember the significance of that".

lib/bundle/checker.rb
colindean: See https://github.com/Homebrew/homebrew-bundle/pull/327/review_comment/create#discussion_r209120013

lib/bundle/commands/check.rb
colindean: It's more explicit that way, should another option (e.g. `--exit-on-error` or `--output-all-errors`) or other logic come into play. It separates the desire from the business logic.

The next step would be making `"--verbose"` a constant, haha.

lib/bundle/commands/check.rb
colindean: ✅

lib/bundle/commands/check.rb
colindean: What would you like to see here? My goal in keeping the checks separate was to isolate `any_formulae_to_start?` from the others because its structure was significantly different and didn't really fit the mold I'd poured for the others.

I could move `"brew bundle can't satisfy your Brewfile's dependencies."` to a constant to keep that message the same across both…?

lib/bundle/mac_app_store_checker.rb
colindean: It's very likely they can be DRYed up. BrewChecker and CaskChecker are virtually the same, with only arg names and the actionable package string as differences. MacAppStoreChecker's `exit_early_check` is also the same as the others, but its `full_check` is necessarily different because of the appid/name pair instead of a string.

```ruby
def full_check(app_ids)
  actionable = app_ids.reject { |id, _name| installed_and_up_to_date? id }
  actionable.map { |_id, name| "App #{name} needs to be installed or updated." }
end
```

If the checkers were classes instead of modules, the DRYed up version could have GenericChecker or something and inherit the functionality. I kept them as modules because that was the pattern established elsewhere throughout the code, IIRC (that decision was six+ months ago?)

lib/bundle/brew_checker.rb
colindean: ✅

lib/bundle/brew_checker.rb
colindean: ✅

colindean:
A test run is taking ~80-90 seconds now…

lib/bundle/checker.rb
MikeMcQuaid: That link points to this PR?

lib/bundle/commands/check.rb
MikeMcQuaid: I think just using `ARGV.include?` everywhere is a bit more obvious. If we introduce another option I think refactoring can be done there otherwise: YAGNI 😉

lib/bundle/commands/check.rb
MikeMcQuaid: I think if the messaging is the same that would be fine. I'd still personally prefer (but really don't feel strongly) that the services checks were done the same way as the others?

lib/bundle/mac_app_store_checker.rb
MikeMcQuaid: Yeh, I reckon classes instead of modules would make this a bit cleaner 👍

lib/bundle/commands/check.rb
colindean: Can we refactor the services checks to fit let in a follow-up PR?

lib/bundle/commands/check.rb
MikeMcQuaid: Personally I find this less rather than more readable; it's unclear to me that `exit_on_first_error` and `output_errors` are related due to the indirection.

lib/bundle/commands/check.rb
MikeMcQuaid: Sure as long as the messaging remains the same here (i.e. can't satisfy ...)

lib/bundle/mac_app_store_checker.rb
MikeMcQuaid: Ideally part of this PR?

lib/bundle/commands/check.rb
colindean: I moved them together into `Bundle::Checker` and the relationship is more apparent. However, I dislike them there because then calling them is `Bundle.checker.output_errors?` and that's long.

I'm gonna mess around with it a little more _after_ I refactor the brew/cask/mas checkers.

colindean:
I've nearly completed the refactor to classes for the checkers. I'm doing some more cleanup and additional refactoring before tackling the services to start check.

lib/bundle/cask_checker.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Checker
+    class CaskChecker < Bundle::Checker::Base
+      PACKAGE_TYPE = :cask
+      PACKAGE_TYPE_NAME = "Cask"
+
+      def installed_and_up_to_date?(formula)
```
MikeMcQuaid: `cask` as an argument here?

lib/bundle/checker.rb
```diff
@@ -0,0 +1,130 @@
...
+
+module Bundle
+  module Checker
+    class Base
+      PACKAGE_TYPE = :pkg
+      PACKAGE_TYPE_NAME = "Package"
+
+      NO_ACTION = [].freeze
+
+      def action_required_for(formula)
```
MikeMcQuaid: Sorry, I'm probably being stupid but I still don't understand how the action for a formula is the formula object is the formula object in an array.

lib/bundle/checker.rb
```diff
@@ -0,0 +1,130 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Checker
+    class Base
+      PACKAGE_TYPE = :pkg
+      PACKAGE_TYPE_NAME = "Package"
```
MikeMcQuaid: Are these variables ever used? If not they could be removed or made a comment.

lib/bundle/checker.rb
```diff
@@ -0,0 +1,130 @@
...
+
+      work_to_be_done = check_method_names.send(enumerator) do |check_method|
+        check_errors = send(check_method)
+        completed_checks << check_method
+        any_errors = check_errors.any?
+        errors.concat(check_errors) if any_errors
+        any_errors
+      end
+
+      work_to_be_done = work_to_be_done.any? if work_to_be_done.class == Array
```
MikeMcQuaid: Sniffing the class name here feels a bit weird.

lib/bundle/checker.rb
```diff
@@ -0,0 +1,130 @@
...
+
+      unchecked_checks = (check_method_names - completed_checks)
+
+      CheckResult.new work_to_be_done, completed_checks, errors, unchecked_checks
+    end
+
+    def casks_to_install
+      Bundle::Checker::CaskChecker.new.find_actionable @dsl.entries
```
MikeMcQuaid: 👍 to this API now.

lib/bundle/commands/check.rb
```diff
@@ -5,75 +5,38 @@ module Commands
...
+            check_result.errors.each { |package| puts "#{ARROW} #{package}" }
+          end
+          puts "Satisfy missing dependencies with `brew bundle install`."
+          exit 1
+        elsif Bundle::Checker.any_formulae_to_start?
+          puts FAILURE_MESSAGE
+          puts "At least one brew formula must be started."
+          exit 1
+        else
+          puts SUCCESS_MESSAGE
```
MikeMcQuaid: I'd just inline this constant as it's only used here.

lib/bundle/commands/check.rb
```diff
@@ -5,75 +5,38 @@ module Commands
...
+      FAILURE_MESSAGE = "brew bundle can't satisfy your Brewfile's dependencies."
+      SUCCESS_MESSAGE = "The Brewfile's dependencies are satisfied."
 
-      def any_formulae_to_install?
-        @dsl ||= Bundle::Dsl.new(Bundle.brewfile)
-        requested_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
-        requested_formulae.any? do |f|
-          !Bundle::BrewInstaller.formula_installed_and_up_to_date?(f)
-        end
+      def output_errors?
```
MikeMcQuaid: Still think these methods would be better off as direct calls to `ARGV.include`

lib/bundle/commands/check.rb
```diff
@@ -5,75 +5,38 @@ module Commands
...
-          old_name ||= old_names[e.name.split("/").last]
-          next if old_name && Bundle::BrewServices.started?(old_name)
+        if check_result.work_to_be_done
+          puts FAILURE_MESSAGE
 
-          true
+          if output_errors?
+            checks = Bundle::Checker::CHECKS
+            check_result.completed_checks.each { |checked| puts "#{checks[checked]} were checked." }
+            check_result.unchecked_checks.each { |unchecked| puts "#{checks[unchecked]} were not checked." }
```
MikeMcQuaid: Not sure I understand here; in `--verbose` mode why/when would things not be checked?

spec/check_command_spec.rb
```diff
@@ -93,4 +113,79 @@ def do_check
       end
     end
   end
+
+  context "when there are taps to install" do
+    before do
+      allow(ARGV).to receive(:value).and_return(nil)
+      allow_any_instance_of(Pathname).to receive(:read).and_return("")
+      allow(Bundle::Checker).to receive(:taps_to_tap).and_return(["asdf"])
+    end
```
MikeMcQuaid: Empty line after each `end`

lib/bundle/cask_checker.rb
colindean: Whoops, ©️ 🍝 !

✅

lib/bundle/checker.rb
colindean: The feeling I'm going for here would be expressed like this in Scala:

```scala
trait Action
case object NoAction extends Action
case class ActionRequired(formula: Formula) extends Action
```

`NO_ACTION`/`NoAction` is a symbol that nothing needs to be done. Its underlying implementation is just an empty array. It could very well be anything else – `false`, `nil` – but it's convenient for the consumers of the function in which it's used for it to be that empty array.

`action_required_for`/`ActionRequired` is a symbol that something has to be done, a simple wrapper that conveys that further action somehow. It could be a separate class as it would be in a statically typed language but I've chosen to implement it as an array because it's convenient for the consumers of the function in which it's used for it to be that array.

I feel that without this in place, it becomes harder to understand what is expected to be done with the output. It'd be just arrays with no clear _meaning_ of the arrays.

lib/bundle/checker.rb
colindean: They are not used, mostly there for catching poor inheritance. I guess I'll make them a comment.

lib/bundle/checker.rb
colindean: Yeah… I did that a few days ago. It got the job done and I feel weird about it. I'd like to revisit it at some point. It's there because `any?` returns boolean and `map` returns an array, and the consumers of that value expect a boolean.

Maybe I should be using `find` instead of `any?` for the enumerator when exiting early? That'd keep the operation exiting at the first it finds, right? If I've always got an array, then I can drop the `if wtbd.class == Array` check.

Thoughts?

lib/bundle/checker.rb
colindean: 🎉

lib/bundle/commands/check.rb
colindean: It felt weird having the `FAILURE_MESSAGE` as a constant and not the success message. I'll change it back…

lib/bundle/commands/check.rb
colindean: That's fair, those two lines used to be exercised in exit early mode. If they're only output in full check mode, then theoretically _everything_ should be checked.

I can remove them if you'd like. I think I'd prefer to keep them there in case something happens (but I'll admit I don't know what) but if you want to invoke YAGNI, go for it.

lib/bundle/commands/check.rb
colindean: I remembered where I've used these that really came in handy: the tests. They already existed before I involved them in tests; they were just hella convenient when I needed them.

```ruby
allow(Bundle::Commands::Check).to receive(:exit_on_first_error?).and_return(true)
```

I kinda remember having some trouble setting up tests with `--verbose` enabled. I remember spending a couple hours on it. Then I realized that I could inject the above and it worked.

If I `:focus` on the test on line 153 of `check_command_spec.rb` and comment out

```ruby
allow(ARGV).to receive(:include?).and_return(true)
```

which seems like it actually _should_ enable verbose mode, too, but it's necessary to get a brewfile loaded. Without it, the test cannot load the brewfile:

```
Run options: include {:focus=>true}
F

Failures:

  1) Bundle::Commands::Check when verbose mode is not enabled stops checking after the first missing formula
     Failure/Error: expect { do_check }.to raise_error(SystemExit)

       expected SystemExit, got #<NoMethodError: undefined method `value' for ["--pattern", "spec/**{,/*/**}/*_spec.rb"]:Array
       Did you mean?  values_at> with backtrace:
         # ./lib/bundle/bundle.rb:51:in `brewfile'
         # ./lib/bundle/checker.rb:67:in `check'
         # ./lib/bundle/commands/check.rb:20:in `run'
         # ./spec/check_command_spec.rb:7:in `do_check'
         # ./spec/check_command_spec.rb:161:in `block (4 levels) in <top (required)>'
         # ./spec/check_command_spec.rb:161:in `block (3 levels) in <top (required)>'
         # ./spec/spec_helper.rb:40:in `block (3 levels) in <top (required)>'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `block in with_clean_env'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:540:in `with_env'
         # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `with_clean_env'
         # ./spec/spec_helper.rb:40:in `block (2 levels) in <top (required)>'
     # ./spec/check_command_spec.rb:161:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:40:in `block (3 levels) in <top (required)>'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `block in with_clean_env'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:540:in `with_env'
     # /Users/colin/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.1/lib/bundler.rb:297:in `with_clean_env'
     # ./spec/spec_helper.rb:40:in `block (2 levels) in <top (required)>'

Finished in 0.053 seconds (files took 0.28185 seconds to load)
1 example, 1 failure
```

So, at this point, if you can suggest an improvement to the test setup that correctly enables verbose mode, I'll be able to remove those two methods. Otherwise, they basically have to stay in order to setup the test harness.

spec/check_command_spec.rb
colindean: ✅

lib/bundle/checker.rb
colindean: Ah, no, because `find` returns the first result, not an array. But, what if I put in some array flexory?

```patch
diff --git a/lib/bundle/checker.rb b/lib/bundle/checker.rb
index 344f3cd..c4fc30a 100644
--- a/lib/bundle/checker.rb
+++ b/lib/bundle/checker.rb
@@ -70,7 +70,7 @@ module Bundle

       completed_checks = []
       errors = []
-      enumerator = exit_on_first_error ? :any? : :map
+      enumerator = exit_on_first_error ? :find : :map

       work_to_be_done = check_method_names.send(enumerator) do |check_method|
         check_errors = send(check_method)
@@ -80,7 +80,7 @@ module Bundle
         any_errors
       end

-      work_to_be_done = work_to_be_done.any? if work_to_be_done.class == Array
+      work_to_be_done = [work_to_be_done].flatten.any?

       unchecked_checks = (check_method_names - completed_checks)

```

I could put that mechanism into another method, perhaps called `ensure_array` or something like that. 

This way is a little nicer to read but creating an array just to flatten it vs checking types ¯\\\_(ツ)_/¯

MikeMcQuaid: `Array(work_to_be_done).flatten.any?` will ensure it's an array without creating a new one.

lib/bundle/commands/check.rb
MikeMcQuaid: > if you want to invoke YAGNI, go for it.

Yes, sorry, thanks 👍

spec/check_command_spec.rb
MikeMcQuaid: Thanks!

lib/bundle/commands/check.rb
```diff
@@ -5,75 +5,37 @@ module Commands
...
+            check_result.errors.each { |package| puts "#{ARROW} #{package}" }
+          end
+          puts "Satisfy missing dependencies with `brew bundle install`."
+          exit 1
+        elsif Bundle::Checker.any_formulae_to_start?
+          puts FAILURE_MESSAGE
+          puts "At least one brew formula must be started."
```
MikeMcQuaid: `At least one service still needs to be started` or something like that, maybe? As-is it feels a bit like "just start any formula('s service) and this will work"

MikeMcQuaid:
> but it's convenient for the consumers of the function in which it's used for it to be that empty array.

I think that's the bit that smells a bit weird to me; it would be nicer to have it be a `nil` or just truthiness check and have the consumer call `Array()` or `to_a` on it instead.

> I feel that without this in place, it becomes harder to understand what is expected to be done with the output. It'd be just arrays with no clear meaning of the arrays.

I would say that relying on basic types (Array/Hash/Boolean/nil) is a bit more Ruby-like. Perhaps a Hash could be used if you want to express emptiness vs. non-emptiness with some sort of meaning? It seems strange for anything relating to an "action" to be something that's not an "action" but something for which an action must be performed (and what action, when and by whom?).

lib/bundle/checker.rb
colindean: ✅

lib/bundle/commands/check.rb
colindean: ✅

colindean:
I'm learning all things about `Array()` today. Imagine my surprise when I learn that `Array(nil)` returns `[]`! Combined with `Array.find` returning the element _or_ `nil`, the result basically gets to what I want: the same data structure coming out of `full_check` and `exit_early_check`. The `NO_ACTION` and `action_required_for` were a kinda nice DSL (at least, to me) for expressing the intent but if I can explicitly say, "Yo, it's always an array," I get something I can handle safely without resorting to type checking or truthy/falsey that feels… _uncomfortable_ for me nowadays.

I do however need to adjust the test now because I need a way to verify that `exit_early_check` is getting called. Since there's not a branch inside it anymore, I can check that it's called and rely on the behavior of `Array.find` to know that it'll work as expected.

colindean:
I'm not quite sure why Travis isn't able to find Bundle::Checker:Base. 

I'm a little behind on Ruby versions:

```
colin@ayla $ ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
```

Travis:

```
$ ruby --version
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
```

So I installed 2.3.3 and it works fine… everything's committed. I must be starting at this for too long?

colindean:
I can only really see that maybe the spec_helper is loading parts of `lib` out of order ⁉️

colindean:
I think the only remaining task is to move the service checker to the `Bundle::Checker::Base` infrastructure, right? GitHub doesn't make it very easy to track TODOs from reviews…

colindean:
Assuming that I have no other style errors, I think this is ready for another review.

One thing I note that that in moving the services check into the list of checks in 9a55b2a, the separate error message is lost.

Should I add something in the error condition that checks if the errors contain a service? I think that doing this will require `Bundle::Checker::Base#find_actionable` and its friends to return more than just a text name of the package. I'll probably have to pass around the `Bundle::Dsl::Entry` instead. That's not a huge change – mostly skipping `Bundle::Checker::Base#format_checkable` and changing the relevant `installed_and_up_to_date?` methods in subclasses of `Bundle::Checker::Base` to handle the `Bundle::Dsl::Entry` instead.

MikeMcQuaid:
What an epic journey! Congratulations @colindean and well done on all the really hard work you've put into this 👏 👏 👏

colindean:
🎉 

Thank you for your attention and review.

I'm glad this feature exists now: my team at work will be making use of it as we onboard folks and I'll be using it myself as I finally get around to actually managing my Homebrew installation across multiple machines using [my dotfiles management system](https://github.com/colindean/hejmo). That's the initiative that started all of this!

MikeMcQuaid:
@colindean Awesome, glad to hear it. Excuse the self-promotion but based on my work on similar stuff in the past https://mikemcquaid.com/2016/06/15/replacing-boxen/ may be of interest.

colindean:
Thanks, I'll have to take a look. I've been wanting to write a talk on home directory management tools…

PR #329, opened 2018-02-18
Adds notes about steps before submitting a PR

colindean:
Running tests is obvious but the style check step is not.

The style check should probably be added to a rake task…

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,8 @@
 * Commit, do not mess with rakefile, version, or history.
   (if you want to have your own version, that is fine but bump version in a commit by itself I can ignore when I pull)
 * Send a pull request. Bonus points for topic branches.
+  * Your branch must pass tests: run `rake` or `rspec spec/{file or functionality you modified}_spec.rb`.
+  * Your branch must pass style checking: run `brew style homebrew/bundle` if your working copy is tapped. If not, run `brew style .` and ignore "Use hash rockets syntax" warnings.
 
 ## Security
 Please report security issues to security@brew.sh.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,19 @@
 ## Contributing
-* Fork the project.
-* Make your feature addition or bug fix.
-* Add tests for it. This is important so I don't break it in a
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@
 * Commit, do not mess with rakefile, version, or history.
   (if you want to have your own version, that is fine but bump version in a commit by itself I can ignore when I pull)
 * Send a pull request. Bonus points for topic branches.
+  * Your branch must pass tests: run `rake` or `rspec spec/{file or functionality you modified}_spec.rb`.
```
MikeMcQuaid: I'd just recommend the `rake` as `rspec spec/` will fail the 100% coverage requirement.

MikeMcQuaid: Also how about: `Your PR must pass all existing tests with 100% code coverage: run rake`

CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@
...
+  * Your branch must pass style checking: run `brew style homebrew/bundle` if your working copy is tapped. If not, run `brew style .` and ignore "Use hash rockets syntax" warnings.
```
MikeMcQuaid: How about `Your PR must pass Rubocop checks: run brew style homebrew/bundle`. I'd suggest omitting the `brew style` one because that will generally be wrong.

colindean:
I reworded a lot, incorporating some things that came to mind throughout the contribution process ("Oh, I didn't know I needed to do that") and incorporated your sentiments in doing so.

I feel like the `CONTRIBUTING` is prescriptive now and will save others' time when the want to contribute. One of my biggest gotchas early on in this contribution sprint was forgetting to run the style checker.

MikeMcQuaid:
Nice work, thanks again @colindean!

MikeMcQuaid:
@colindean If you were interested in tackling https://github.com/homebrew/gsoc#better-support-pinning-to-a-majorminor-version-with-homebrew-bundle-and--versioned-formulae let me know. It would require some Homebrew/brew and Homebrew/homebrew-bundle changes but I think it could be a good new challenge for you if you were interested.

PR #332, opened 2018-04-02
Fix type formulae > formulas

dsalahutdinov:
Hi, minor fix of type using sed
```bash
sed -i '' 's/formulae/formulas/g'
```


MikeMcQuaid:
Formulae is a UK english plural of formula.

dsalahutdinov:
👍

PR #337, opened 2018-05-25
Update Cask references.

reitermarkus:


lib/bundle/bundle.rb
```diff
@@ -33,8 +33,7 @@ def mas_signedin?
 
   def cask_installed?
     @cask_installed ||= begin
-      File.directory?("#{HOMEBREW_PREFIX}/Caskroom") &&
-        File.directory?("#{HOMEBREW_REPOSITORY}/Library/Taps/caskroom")
```
MikeMcQuaid: This shouldn't be changed; we don't want running `brew cask list` to automatically tap this.

MikeMcQuaid:
Thanks again @reitermarkus!

JCount:
👍

PR #360, opened 2018-08-23
Add docs for zap flag

cesarandreu:
I was uncertain what difference the `--zap` flag would have when running `brew bundle cleanup` and had to take a look at the source to find out. Adding a brief explanation helps clarify the difference.


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @cesarandreu!

PR #362, opened 2018-08-27
Extend valid `mas` app name regex

### LINKED ISSUE #359: Pastebin app not in output of "brew bundle"
Here's the output of `mas list` on my machine:

```
Pro: ~ $ mas list
1333542190 1Password 7 (7.0.7)
1278508951 Trello (2.10.7)
425424353 The Unarchiver (4.0.0)
408981434 iMovie (10.1.9)
1055511498 Day One (2.7.4)
904280696 Things3 (3.6.1)
946399090 Telegram Desktop (1.3.10)
1384080005 Tweetbot (3.1)
407963104 Pixelmator (3.7.3)
896450579 Textual (6.0.12)
1176895641 Spark (2.0.9)
937984704 Amphetamine (3.0.2)
603637384 Name Mangler 3 (3.4.1)
417375580 BetterSnapTool (1.8)
944924917 Pastebin It! (1.0)
```

Compare to the output of `brew bundle list --mas --global`:

```
1Password 7
Amphetamine
BetterSnapTool
Day One
iMovie
Name Mangler 3
Pixelmator
Spark
Telegram Desktop
Textual
The Unarchiver
Things3
Trello
Tweetbot
```

They're different in that the "Pastebin It!" app isn't included in the output of `brew bundle` even though it's installed. `brew bundle` must be getting hung up somehow, perhaps due to the exclamation mark?

jacobbednarz:
#349 introduced a basic regex for determining if the app names provided
via `mas` were valid and would discard any that failed to meet this
requirement. Apple are [quite generous](https://developer.apple.com/app-store/review/guidelines/#accurate-metadata) when it comes to what you can
name an application and in essence, the only restriction they place on
it is that it cannot infringe on a trademark and must be under 30
characters long.

This means that there are real applications that fail to meet the
initial regex of `\w\s` due to using non-alphanumeric characters.
"Pastebin It!" is an example of this.

To account for this, I've extended the regex to accept any character
for the `name` portion of the regex. This might look fragile (and I
would agree!) however it is the safest way to ensure that we don't
silently drop applications from being included in the output. Should
Apple tighten this restriction in the future, we can firm it up with a
better regex.

Fixes #359

### DIFF
--- a/spec/mac_app_store_dumper_spec.rb
+++ b/spec/mac_app_store_dumper_spec.rb
@@ -59,6 +59,18 @@ describe Bundle::MacAppStoreDumper do
         409201541 Pages (7.1)
         123456789 123AppNameWithNumbers (1.0)
         409203825 Numbers (5.1)
+        944924917 Pastebin It! (1.0)
+        123456789 My (cool) app (1.0)
+        987654321 an-app-i-use (2.1)
+        123457867 App name with many spaces (1.0)
+        893489734 my,comma,app (2.2)
+        832423434 another_app_name (1.0)
+        543213432 My App? (1.0)
+        688963445 app;with;semicolons (1.0)
+        123345384 my 😊 app (2.0)
+        896732467 你好 (1.1)
+        634324555 مرحبا (1.0)
+        234324325 áéíóú (1.0)
       HEREDOC
     end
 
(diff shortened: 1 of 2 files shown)


spec/mac_app_store_dumper_spec.rb
```diff
@@ -59,6 +59,14 @@
         409201541 Pages (7.1)
         123456789 123AppNameWithNumbers (1.0)
         409203825 Numbers (5.1)
+        944924917 Pastebin It! (1.0)
+        123456789 My (cool) app (1.0)
+        987654321 an-app-i-use (2.1)
+        123457867 App name with many spaces (1.0)
+        893489734 my,comma,app (2.2)
```
jacobbednarz: These are just some (far-fetched) examples of what we should be expecting to accept. Do let me know if you can think of any others we'd like to add here.

colindean: I think I read in the conversation  on #359 that there is a possibility for emoji in a name. If that's true, it'd be nice to have a test or two for that, with different ranges of emoji characters and maybe non-Latin characters, too..

jacobbednarz: Thanks @colindean! I've pushed up a change includes an emoji and some non-latin characters. Let the encoding issues begin 😛 !

spec/mac_app_store_dumper_spec.rb
```diff
@@ -59,6 +59,18 @@
...
+        832423434 another_app_name (1.0)
+        543213432 My App? (1.0)
+        688963445 app;with;semicolons (1.0)
+        123345384 my 😊 app (2.0)
+        896732467 你好 (1.1)
+        634324555 مرحبا (1.0)
```
jacobbednarz: this looks different to what I have locally, maybe something for @MikeMcQuaid to investigate internally 😄

MikeMcQuaid:
@jacobbednarz @colindean would either of you be interested in being maintainers of this repository (or Homebrew itself)? Neither of you have emails on your GitHub profile! Email me at my email on mine.

Issue #358, opened 2018-08-20
Feature request: tap pinning


rlex:
Will be nice to have ability to use "brew tap-pin user/tap" from bundler.

Maybe support something like 
```
tap "homebrew/cask", tap_pin: true
```
In Brewfile?

rlex:
Might try but unlikely, i'm not so familiar with homebrew internals.

MikeMcQuaid:
No-one is to begin with. Give it a go!

PR #363, opened 2018-08-28
Add tap pinning

mikesplain:
Adds tap pinning. 

Fixes https://github.com/Homebrew/homebrew-bundle/issues/358

It's been awhile since I worked with Ruby, so I'm open to any changes!


lib/bundle/tap_installer.rb
```diff
@@ -4,27 +4,49 @@ module Bundle
...
+
+    def check_pinning(name, options)
+      pin = options[:pin]
+      currently_pinned = pinned_installed_taps.include? name
+      if pin && !currently_pinned
+        puts "Pinning #{name} tap." if ARGV.verbose?
+        return :failed unless Bundle.system "brew", "tap-pin", name
+        pinned_installed_taps << name
+      elsif currently_pinned && !pin
+        puts "Unpining #{name} tap." if ARGV.verbose?
```
colindean: `s/Unpining/Unpinning`

spec/tap_installer_spec.rb
```diff
@@ -13,21 +17,36 @@ def do_install(clone_target = nil)
     end
   end
 
+  context ".pinned_installed_taps" do
+    it "calls Homebrew" do
+      Bundle::TapInstaller.pinned_installed_taps
```
colindean: I can tell that you're following suit from `context ".installed_taps"` above, but I feel like this should have some kind of explicit expectation (as should the one above). I don't feel strongly enough about it to block on that feeling, though.

lib/bundle/tap_installer.rb
mikesplain: Ahh yeah good eye! Thanks!

mikesplain:
Thanks @colindean!

Readme.md
```diff
@@ -32,6 +32,7 @@ Add your dependencies in your `Brewfile`:
 ```ruby
 cask_args appdir: "/Applications"
 tap "homebrew/cask"
+tap "homebrew/cask", pin: true
```
MikeMcQuaid: @mikesplain Sorry for pedantry but I think it's worth adding a tap here that would make sense to pin instead (as pinning a cask tap won't really do anything)? Perhaps just the `telemachus` one below?

lib/bundle/dsl.rb
```diff
@@ -59,11 +59,22 @@ def mas(name, options = {})
       @entries << Entry.new(:mac_app_store, name, id: id)
     end
 
-    def tap(name, clone_target = nil)
```
MikeMcQuaid: I wonder if it's possible to do `name, clone_target = nil, pin: false` or similar to avoid some of the logic below. @reitermarkus, any thoughts?

lib/bundle/tap_dumper.rb
```diff
@@ -20,12 +20,17 @@ def taps
...
+        "tap \"#{tap["name"]}\"#{remote}#{pinned}"
       end.sort.join("\n")
     end
 
     def tap_names
       taps.map { |tap| tap["name"] }
     end
+
+    def pinned_tap_names
+      taps.map { |tap| (tap["pinned"] == true) ? tap["name"] : nil }.compact
```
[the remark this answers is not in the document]
mikesplain: Doh! Thanks for that!

mikesplain:
Thanks @MikeMcQuaid and @colindean! I made the changes above. I'll take any advice on the remaining conversations, I don't feel strongly just didn't want to break anything and keep things simple for bundle users 😄.

lib/bundle/dsl.rb
reitermarkus: I think `name, clone_target = nil, pin: false` should work fine.

spec/tap_installer_spec.rb
```diff
@@ -3,8 +3,12 @@
 require "spec_helper"
 
 describe Bundle::TapInstaller do
-  def do_install(clone_target = nil)
-    Bundle::TapInstaller.install("phinze/cask", clone_target)
+  def do_install(options = {})
+    Bundle::TapInstaller.install("phinze/cask", options)
```
reitermarkus: Can we change `phinze/cask` to something else here?

Readme.md
```diff
@@ -33,6 +33,7 @@ Add your dependencies in your `Brewfile`:
 cask_args appdir: "/Applications"
 tap "homebrew/cask"
 tap "telemachus/brew", "https://telemachus@bitbucket.org/telemachus/brew.git"
+tap "telemachus/brew", pin: true
```
MikeMcQuaid: Append `pin: true` to the line above instead, thanks.

spec/tap_installer_spec.rb
MikeMcQuaid: `homebrew/cask` I guess, thanks.

lib/bundle/dsl.rb
MikeMcQuaid: Yeh, that'd be good thanks @reitermarkus, @mikesplain!

spec/dsl_spec.rb
```diff
@@ -12,6 +12,7 @@
       cask_args appdir: '/Applications'
       tap 'homebrew/cask'
       tap 'telemachus/brew', 'https://telemachus@bitbucket.org/telemachus/brew.git'
+      tap 'telemachus/brew', pin: true
```
MikeMcQuaid: Append `pin: true` to the line above instead, thanks.

lib/bundle/dsl.rb
mikesplain: Ahh yes didn't think of something so simple. . .  Thanks!

mikesplain:
Changes made, let me know if you see anything else!

PR #371, opened 2018-10-01
upgrade the json gem from 1.8.3 to 1.8.6 to get ruby >=2.4 support

phallstrom:
Version 1.8.3 of the json gem isn't compatable with ruby 2.4 and later.
see: https://github.com/flori/json/issues/303

Specifically require the json gem to ensure we fetch a compatible
version.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source "https://rubygems.org"
 
 group :test do
+  gem "json", "~> 1.8.6"
   gem "codecov"
   gem "rake"
   gem "rspec"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ GEM
       url
     diff-lcs (1.3)
     docile (1.1.5)
-    json (1.8.3)
+    json (1.8.6)
     rake (10.4.2)
     rspec (3.7.0)
       rspec-core (~> 3.7.0)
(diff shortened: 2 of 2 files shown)


colindean:
What problem did you encounter in `brew bundle` that led you to submit this?

phallstrom:
I was going to resurrect my work from https://github.com/Homebrew/homebrew-bundle/pull/297 and was unable to install/run tests locally on a fresh copy of master.

colindean:
Some time has passed since then. `brew bundle dump --describe` does what #297 would do now. Try it out!

phallstrom:
Well, look at that :)  In that case... consider https://github.com/Homebrew/homebrew-bundle/pull/372

colindean:
This is now obsolete, #372 documented `--describe`, which was the intent of #297, and this PR was driving toward that while building brew on Ruby 2.4.

phallstrom:
Not sure this is obsolete... I still can't `bundle install` if I check out this repo.

colindean:
@phallstrom Let's handle that as a separate bug issue.

phallstrom:
@colindean separate issue? Not sure I understand.. this PR does nothing but update the gem version which is necessary to install it with ruby >= 2.4

MikeMcQuaid:
`brew bundle` does not use Ruby 2.4 or `bundle install` at runtime. Adjust your local ruby version accordingly if you're having trouble running `rake` with this repository for development.

phallstrom:
:facepalm:  I just went to verify this and I don't know how, but the version I was checking against was really really old.  Sorry for the noise.

PR #374, opened 2018-10-02
mas: better handle signin failure

### LINKED ISSUE #370: mas exception
bundle crashes when trying to install mas packages, below is an example for Aphetamine app:
```
$ brew bundle --global --verbose
Not signed in
Error: Not signed in
Not signed in to Mac App Store.
==> Signing in to Apple ID:
2018-09-30 07:41:45.129 mas[51954:6450920] -[__NSXPCInterfaceProxy_ISAccountService signInWithContext:replyBlock:]: unrecognized selector sent to instance 0x7fbf3d51dab0
2018-09-30 07:41:45.138 mas[51954:6450920] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSXPCInterfaceProxy_ISAccountService signInWithContext:replyBlock:]: unrecognized selector sent to instance 0x7fbf3d51dab0'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff48c8b43d __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff74b98720 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff48d08255 -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x00007fff48c2aad0 ___forwarding___ + 1486
	4   CoreFoundation                      0x00007fff48c2a478 _CF_forwarding_prep_0 + 120
	5   mas                                 0x000000010f3b7e53 $SSo14ISStoreAccountC3masE6signIn8username8password12systemDialogABSSSg_AHSbtKFZTf4nnnd_n + 947
	6   mas                                 0x000000010f3cdf3a $S3mas13SignInCommandV3runyAA6ResultOyytAA8MASErrorOGAA0bC7OptionsVFTf4nd_n + 410
	7   mas                                 0x000000010f3a8211 $S3mas14CommandWrapperVyACyxGqd__c11ClientErrorQyd__RszAA0B8ProtocolRd__7Options_AEQYd__AFRSlu33_E513F8A29DAA43E418079999EEBED6A7LlfcAA6ResultOyytAA010CommandantE0OyxGGAA14ArgumentParserCcfU_AA8MASErrorO_AA06SignInB0VTg5Tf4nd_n + 465
	8   mas                                 0x000000010f396cbb $S3mas14CommandWrapperVyACyxGqd__c11ClientErrorQyd__RszAA0B8ProtocolRd__7Options_AEQYd__AFRSlu33_E513F8A29DAA43E418079999EEBED6A7LlfcAA6ResultOyytAA010CommandantE0OyxGGAA14ArgumentParserCcfU_AA8MASErrorO_AA07AccountB0VTG5TATm + 11
	9   mas                                 0x000000010f396c40 $S3mas14CommandWrapperVyACyxGqd__c11ClientErrorQyd__RszAA0B8ProtocolRd__7Options_AEQYd__AFRSlu33_E513F8A29DAA43E418079999EEBED6A7LlfcAA6ResultOyytAA010CommandantE0OyxGGAA14ArgumentParserCcfU_AA8MASErrorO_AA06SignInB0VTG5TA + 16
	10  mas                                 0x000000010f3a6def $S3mas15CommandRegistryC4main9arguments11defaultVerb12errorHandlers5NeverOSaySSG_SSyxXEtFAA8MASErrorO_Tg507$S3mas8K43OIgg_ACIegn_TR010$S3masyAA8B6OXEfU_Tf3npf_nTf1nncn_nTf4nxn_n + 703
	11  mas                                 0x000000010f3940ca main + 5946
	12  libdyld.dylib                       0x00007fff75c66085 start + 1
	13  ???                                 0x0000000000000004 0x0 + 4
)
libc++abi.dylib: terminating with uncaught exception of type NSException
Not signed in
Error: Not signed in
Error: Unable to install Amphetamine app. mas not signed in to Mac App Store.
```
mas works outside bundle:
```
$ mas install 937984704
Warning: Amphetamine is already installed
```

MikeMcQuaid:
- see if app can be skipped because it doesn’t need upgrade
- don’t try to sign in on Mojave (it doesn’t work)
- only fail if app actually needs installed/upgraded
- don’t print `mas account` output; just use exit code
- add `MacOS` stub and cleanup existing stub code

Fixes #370

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -28,7 +28,7 @@ module Bundle
   end
 
   def mas_signedin?
-    Bundle.system "mas", "account"
+    Kernel.system "mas account &>/dev/null"
   end
 
   def cask_installed?
--- a/spec/stub/formulary.rb
+++ b/spec/stub/formulary.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class Formulary
+  def self.factory(name)
+    Formula.new(name)
+  end
+end
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "utils/formatter"
+
 module Bundle
   module Commands
     # TODO: refactor into multiple modules
@@ -42,8 +44,6 @@ module Bundle
             puts cleanup
           end
         else
-          require "utils/formatter"
-
           if casks.any?
             puts "Would uninstall casks:"
             puts Formatter.columns casks
--- a/lib/bundle/commands/exec.rb
+++ b/lib/bundle/commands/exec.rb
@@ -2,6 +2,7 @@
 
 require "extend/ENV"
 require "formula"
+require "formulary"
 require "utils"
 
 module Bundle
--- a/lib/bundle/mac_app_store_installer.rb
+++ b/lib/bundle/mac_app_store_installer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "os"
+
 module Bundle
   module MacAppStoreInstaller
     module_function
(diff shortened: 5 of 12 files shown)


lib/bundle/bundle.rb
```diff
@@ -28,7 +28,7 @@ def mas_installed?
   end
 
   def mas_signedin?
-    Bundle.system "mas", "account"
+    Kernel.system "mas account &>/dev/null"
```
colindean: I'm OK with this but a feel a little weird about it. We've defined our own `system()` just above, should this instead use that and that method have a default parameter that shortcuts like `if ARGV.verbose?`

spec/stub/formulary.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
colindean: All of these stubs are available when run within a real `brew` context, right?

colindean:
Ha, you merged just as I hit "submit review" 👍

spec/stub/formulary.rb
MikeMcQuaid: Yep.

MikeMcQuaid:
> Ha, you merged just as I hit "submit review" 👍

sorry!

lib/bundle/bundle.rb
MikeMcQuaid: @colindean I tried to refactor accordingly but given the behaviour we want here (don't ever print any output regardless) it felt better to just use this. If there was another 1-2 uses in this form I'd be 👍 on adding a separately method.

PR #376, opened 2018-10-02
Update/cleanup all gems

MikeMcQuaid:
Also, remove unnecessary `spec_helper` code after gem updates.


MikeMcQuaid:
Have added @dependabot to this repository for doing this automatically in future.

PR #377, opened 2018-10-03
Update bundle help text

EricFromCanada:
Fix a few inconsistencies and append colons to command specs.

### DIFF
--- a/cmd/brew-bundle.rb
+++ b/cmd/brew-bundle.rb
@@ -1,43 +1,45 @@
 #:  * `bundle` <command>:
 #:    Bundler for non-Ruby dependencies from Homebrew.
 #:
-#:    `bundle` [`install`] [`-v`|`--verbose`] [`--no-upgrade`] [`--file=`<path>|`--global`]:
+#:    `brew bundle` [`install`] [`-v`|`--verbose`] [`--no-upgrade`] [`--file=`<path>|`--global`]:
 #:    Install or upgrade all dependencies in a Brewfile.
 #:
-#:    `brew bundle dump` [`--force`] [`--describe`] [`--file=`<path>|`--global`]
+#:    `brew bundle dump` [`--force`] [`--describe`] [`--file=`<path>|`--global`]:
 #:    Write all installed casks/formulae/taps into a Brewfile.
 #:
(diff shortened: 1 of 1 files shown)


colindean:
I'm not sure how I feel about the colons but I'm cool with all of the other changes.

EricFromCanada:
This PR relates to Homebrew/brew#5027.

jacobbednarz:
I'd be happy waiting to see what `Homebrew/brew` does and if it's accepted there, merge it in here.

cmd/brew-bundle.rb
```diff
@@ -1,43 +1,45 @@
...
-#:    If `--no-upgrade` is passed, don't run `brew upgrade` outdated dependencies.
+#:    If `--no-upgrade` is passed, don't run `brew upgrade` on outdated dependencies.
 #:    Note they may still be upgraded by `brew install` if needed.
 #:
 #:    If `--force` is passed, uninstall dependencies or overwrite an existing Brewfile.
 #:
 #:    If `--zap` is passed, casks will be removed using the `zap` command instead of `uninstall`.
 #:
-#:    If `--file=<path>` is passed, the Brewfile path is set accordingly
+#:    If `--file=`<path> is passed, the Brewfile path is set accordingly.
```
jacobbednarz: I think this should stay as `--file=<path>`

EricFromCanada: This change is because when `brew man` is run, `ronn`'s `--markdown` option converts all angle bracketed `<words>` to `<var>words</var>`, which `man.rb` then converts to backticks. As it's already surrounded by backticks in the source, this causes a backticked word to appear inside a backticked word, resulting in two backticks next to each other, which appear as two literal backticks in the man page's HTML version. There is currently an ugly hack in `man.rb` that prevents this from happening, but the original markup above results in "path" appearing as regular text, instead of monospaced.

jacobbednarz: Thanks for the explanation @EricFromCanada!

MikeMcQuaid:
Let's hold off merging this until https://github.com/Homebrew/brew/pull/5027 👍

MikeMcQuaid:
Thanks again @EricFromCanada!

PR #410, opened 2018-12-13
dumper: fix dumping cask requirements.

MikeMcQuaid:
Ensure that we strip the tap section so they are dumped properly.

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -39,7 +39,9 @@ module Bundle
     end
 
     def cask_requirements
-      formulae.map { |f| f[:requirements].map { |req| req["cask"] } }.flatten.compact.uniq
+      formulae.flat_map do |f|
+        f[:requirements].map { |req| req["cask"]&.split("/")&.last }
+      end.compact.uniq
     end
 
     def formula_names


lib/bundle/dumper.rb
```diff
@@ -15,7 +15,8 @@ def can_write_to_brewfile?(brewfile_path)
     def build_brewfile
       content = []
       content << TapDumper.dump
-      casks_required_by_formulae = BrewDumper.cask_requirements
+      casks_required_by_formulae =
+        BrewDumper.cask_requirements.map {|cask| cask.split("/").last}
```
colindean: This is the only usage of `BrewDumper#cask_requirements` outside of tests.
```
$ rg cask_requirements
spec/bundle/brew_dumper_spec.rb
403:      expect(dumper.cask_requirements).to eq %w[bar]

lib/bundle/dumper.rb
18:      casks_required_by_formulae = BrewDumper.cask_requirements

lib/bundle/brew_dumper.rb
41:    def cask_requirements
```
 This transformation _could_ be inside of that method instead of here. I'd say that a pair of methods, `cask_requirements` and `cask_requirements_with_taps` would suffice, but YAGNI leads me away from that.

I think I'd like to see a test for this, too. I think that would adequately document the expected behavior of the method. It could be around `spec/bundle/brew_dumper_spec.rb:403` or thereabouts.

MikeMcQuaid: > This transformation could be inside of that method instead of here.

Done 👍

PR #429, opened 2019-01-16
Explicitly install bundler and gems before using them

colindean:
Should enable #427 and #428 to proceed

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,8 +11,9 @@ jobs:
         # https://github.com/Homebrew/homebrew-bundle/pull/384
         # brew style homebrew/bundle
         # Use local config until then
-        rubocop --force-exclusion --parallel --except FormulaAudit,FormulaAuditStrict,NewFormulaAudit --format simple
+        gem install bundler
         bundle install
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -11,8 +11,9 @@ jobs:
         # https://github.com/Homebrew/homebrew-bundle/pull/384
         # brew style homebrew/bundle
         # Use local config until then
-        rubocop --force-exclusion --parallel --except FormulaAudit,FormulaAuditStrict,NewFormulaAudit --format simple
+        gem install bundler:1.16.6
```
MikeMcQuaid: Why not use the latest version?

colindean: [Tried](https://github.com/Homebrew/homebrew-bundle/pull/429/commits/f19402e486f597ae045d73485985e5ad3c002d1f#diff-fec826feae04e51c0d94076385408bdcR14), didn't work. Easier to set version explicitly and revisit when I had a larger swath of time available. I was starting a job interview in less than two minutes from when I noticed the CI failures and even used the first minute of the job interview to explain what I was doing!

PR #437, opened 2019-01-18
Synchronizes Rubocop configuration with homebrew/brew

colindean:
And fix the one violation still present


MikeMcQuaid:
Seems fine, thanks. I'm wondering if we need a better solution though given how old 
https://github.com/Homebrew/homebrew-bundle/issues/388 is getting.

PR #445, opened 2019-02-05
install: Ensure all formulae are linked

sjackman:
Link formulae that are unlinked and not keg only.

Fixes https://github.com/Homebrew/homebrew-bundle/issues/468

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -73,11 +73,15 @@ module Bundle
     def link_change_state!
       case @link
       when true
-        puts "Force linking #{@name} formula." if ARGV.verbose?
-        Bundle.system("brew", "link", "--force", @name)
+        unless linked_and_keg_only?
+          puts "Force linking #{@name} formula." if ARGV.verbose?
+          Bundle.system("brew", "link", "--force", @name)
+        end
       when false
-        puts "Unlinking #{@name} formula." if ARGV.verbose?
-        Bundle.system("brew", "unlink", @name)
+        unless unlinked_and_not_keg_only?
+          puts "Unlinking #{@name} formula." if ARGV.verbose?
+          Bundle.system("brew", "unlink", @name)
+        end
       when nil
         if unlinked_and_not_keg_only?
           puts "Linking #{@name} formula." if ARGV.verbose?
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -79,9 +79,12 @@ module Bundle
         puts "Unlinking #{@name} formula." if ARGV.verbose?
         Bundle.system("brew", "unlink", @name)
       when nil
-        if unlinked?
+        if unlinked_and_not_keg_only?
           puts "Linking #{@name} formula." if ARGV.verbose?
           Bundle.system("brew", "link", @name)
+        elsif linked_and_keg_only?
+          puts "Unlinking #{@name} formula." if ARGV.verbose?
+          Bundle.system("brew", "unlink", @name)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


lib/bundle/brew_installer.rb
```diff
@@ -75,6 +75,11 @@ def link_change_state!
       when false
         puts "Unlinking #{@name} formula." if ARGV.verbose?
```
MikeMcQuaid: Maybe this warrants a `linked?` check or similar (like below)

lib/bundle/brew_installer.rb
```diff
@@ -75,6 +75,11 @@ def link_change_state!
...
         Bundle.system("brew", "unlink", @name)
+      when nil
+        if unlinked?
```
MikeMcQuaid: Does this verify keg-only state?

vraravam:
@sjackman - could you please get the review comments fixed?

lib/bundle/brew_installer.rb
sjackman: Done.

lib/bundle/brew_installer.rb
sjackman: Yes.
https://github.com/Homebrew/homebrew-bundle/blob/d6e194364227dfdc11dff12d86d1dfee7b065f5f/lib/bundle/brew_dumper.rb#L119-L126

sjackman: I renamed this predicate `unlinked?` to `unlinked_and_not_keg_only?`.

lib/bundle/brew_installer.rb
sjackman: It's not entirely straight forward with the data that's currently stored. `f[:link?]` stores either `true`, `false`, or `nil`.
- `true` indicates it's linked and keg-only.
- `false` indicates it's not linked and not keg-only.
- `nil` indicates its linkage is typical, either (linked and not keg-only) or (not linked and keg only).

Simply whether the formula is linked or not is not explicitly stored by `Bundle::BrewDumper`.

sjackman:
> Looks good but can you write tests?

How do I run the tests in `spec/bundle/brew_installer_spec.rb`?

sjackman:
`bundle exec rake` gave these errors with Ruby 2.5.1.
```
❯❯❯ bundle exec rake
/Users/sjackman/.homebrew/Cellar/ruby/2.5.1/lib/ruby/2.5.0/uri.rb:106:in `require': /Users/sjackman/.homebrew/Cellar/ruby/2.5.1/lib/ruby/2.5.0/uri/generic.rb:212: syntax error, unexpected '.' (SyntaxError)
      @scheme&.freeze
               ^
/Users/sjackman/.homebrew/Cellar/ruby/2.5.1/lib/ruby/2.5.0/uri/generic.rb:335: syntax error, unexpected '.'
      @scheme = v&.downcase
                   ^
/Users/sjackman/.homebrew/Cellar/ruby/2.5.1/lib/ruby/2.5.0/uri/generic.rb:1320: syntax error, unexpected '.'
      if path&.empty?
               ^
```
Is that expected?

lib/bundle/brew_installer.rb
MikeMcQuaid: @sjackman Could consider extending Homebrew/brew to provide the information here?

MikeMcQuaid:
@sjackman I have no idea what's going on there. I'd suggest running with the normal Homebrew or system version of Ruby.

lib/bundle/brew_installer.rb
sjackman: Never mind. I figured it out. Fixed in 0e0ba74ec025bb5b9e7b82fcfd15c7a312e2b16d

sjackman:
I have the tests working with homebrewed Ruby 2.6.1p33. I'm not sure how much of the following was necessary, but it was sufficient.

```sh
brew upgrade ruby
$(brew --prefix ruby)/bin/gem install bundler -v1.17.3
$(brew --prefix ruby)/bin/gem install bundler
$(brew --prefix ruby)/bin/bundle install
$(brew --prefix ruby)/bin/bundle exec rake
```

vraravam:
ping

sjackman:
I'm defending my PhD thesis next week. Should have time after that to work on this PR again.

MikeMcQuaid:
> I'm defending my PhD thesis next week. Should have time after that to work on this PR again.

@sjackman Good luck 🤞 ❤️

MikeMcQuaid:
@sjackman Gentle bump on this! Shout if I can help at all.

sjackman:
Success!

MikeMcQuaid:
Thanks again @sjackman!

PR #457, opened 2019-03-08
WIP: Prevents exceptions thrown during entry installation

### LINKED ISSUE #456: A failing line in Brewfile should throw an exception, not exit
My new employer seems to block sign-in to the App Store – I cannot for the life of me get it to sign in – so I cannot proceed without editing my Brewfile.

I have had this in my Brewfile for a while:

```
begin
  mas 'Trello', id: 1278508951
rescue
   echo "mas is unavailable, skipping Mac App Store apps"
end
```

I kinda expect it to work: a line shouldn't fail out, it should either continue or throw an exception. Right now, bundle exits when a line – at least a `mas` line – fails.

colindean:
Fixes #456

Still working on the test but my battery is at 3% and I want to get input overnight ;-)

### DIFF
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -4,43 +4,51 @@ module Bundle
   module Installer
     module_function
 
+    InstallerTask = Struct.new :cls, :arg, :verb
+
+    def installer_class_and_args_for_entry(entry)
+      verb = 'Installing'
+      arg = [entry.name]
+      cls = case entry.type
+            when :brew
+              arg << entry.options
+              Bundle::BrewInstaller
+            when :cask
+              arg << entry.options
+              Bundle::CaskInstaller
+            when :mac_app_store
+              arg << entry.options[:id]
+              Bundle::MacAppStoreInstaller
+            when :tap
+              verb = 'Tapping'
+              arg << entry.options
+              Bundle::TapInstaller
+            end
+      InstallerTask.new cls, arg, verb
+    end
+
     def install(entries)
       success = 0
       failure = 0
       errored_entries = {}
 
       entries.each do |entry|
-        arg = [entry.name]
-        verb = "Installing"
-        cls = case entry.type
-        when :brew
-          arg << entry.options
-          Bundle::BrewInstaller
-        when :cask
-          arg << entry.options
-          Bundle::CaskInstaller
-        when :mac_app_store
-          arg << entry.options[:id]
-          Bundle::MacAppStoreInstaller
-        when :tap
-          verb = "Tapping"
-          arg << entry.options
-          Bundle::TapInstaller
-        end
+        task = installer_class_and_args_for_entry(entry)
+        task_header = "#{task.verb} #{entry.name}"
         begin
-          case cls.install(*arg)
+          case task.cls.install(*task.arg)
           when :success
-            puts Formatter.success("#{verb} #{entry.name}")
+            puts Formatter.success(task_header)
             success += 1
           when :skipped
             puts "Using #{entry.name}"
             success += 1
           else
-            puts Formatter.error("#{verb} #{entry.name} has failed!")
+            puts Formatter.error("#{task_header} has failed!")
             failure += 1
           end
         rescue => e
-          puts Formatter.error("#{verb} #{entry.name} raised an exception: #{e}")
+          puts Formatter.error("#{task_header} raised an exception: #{e}")
           errored_entries[entry.name] = e
         end
       end
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -40,7 +40,7 @@ module Bundle
             failure += 1
           end
         rescue => e
-          puts Formatter.error("#{verb} #{entry.name} errored: #{e}")
+          puts Formatter.error("#{verb} #{entry.name} raised an exception: #{e}")
           errored_entries[entry.name] = e
         end
       end
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -52,7 +52,10 @@ module Bundle
       end
 
       unless errored_entries.empty?
-        puts Formatter.error("Homebrew Bundle encountered some errors! #{errored_entries.size} Brewfile #{Bundle::Dsl.pluralize_dependency(errored_entries.size)} failed badly:")
+        error_count = errored_entries.size
+        words = Bundle::Dsl.pluralize_dependency(error_count)
+        error_text = "Homebrew Bundle encountered some errors. #{error_count} Brewfile #{words} failed badly:"
+        puts Formatter.error(error_text)
         errored_entries.each do |entry, error|
           puts Formatter.error("\t#{entry}\t => \t#{error}")
         end
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -7,6 +7,7 @@ module Bundle
     def install(entries)
       success = 0
       failure = 0
+      errored_entries = {}
 
       entries.each do |entry|
         arg = [entry.name]
@@ -26,16 +27,21 @@ module Bundle
           arg << entry.options
           Bundle::TapInstaller
         end
-        case cls.install(*arg)
-        when :success
-          puts Formatter.success("#{verb} #{entry.name}")
-          success += 1
-        when :skipped
-          puts "Using #{entry.name}"
-          success += 1
-        else
-          puts Formatter.error("#{verb} #{entry.name} has failed!")
-          failure += 1
+        begin
+          case cls.install(*arg)
+          when :success
+            puts Formatter.success("#{verb} #{entry.name}")
+            success += 1
+          when :skipped
+            puts "Using #{entry.name}"
+            success += 1
+          else
+            puts Formatter.error("#{verb} #{entry.name} has failed!")
+            failure += 1
+          end
+        rescue => e
+          puts Formatter.error("#{verb} #{entry.name} errored: #{e}")
+          errored_entries[entry.name] = e
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/bundle/installer.rb
```diff
@@ -26,16 +27,21 @@ def install(entries)
...
+            success += 1
+          when :skipped
+            puts "Using #{entry.name}"
+            success += 1
+          else
+            puts Formatter.error("#{verb} #{entry.name} has failed!")
+            failure += 1
+          end
+        rescue => e
+          puts Formatter.error("#{verb} #{entry.name} errored: #{e}")
```
MikeMcQuaid: `raised an exception` rather than `error`ed? I'm not sure the failed/errored distinction is obvious.

lib/bundle/installer.rb
```diff
@@ -26,16 +27,21 @@ def install(entries)
...
-        when :success
-          puts Formatter.success("#{verb} #{entry.name}")
-          success += 1
-        when :skipped
-          puts "Using #{entry.name}"
-          success += 1
-        else
-          puts Formatter.error("#{verb} #{entry.name} has failed!")
-          failure += 1
+        begin
```
MikeMcQuaid: Could this perhaps be split into a separate method to encapsulate the exception handling in there?

lib/bundle/installer.rb
```diff
@@ -45,7 +51,14 @@ def install(entries)
         puts Formatter.error("Homebrew Bundle failed! #{failure} Brewfile #{Bundle::Dsl.pluralize_dependency(failure)} failed to install.")
       end
 
-      failure.zero?
+      unless errored_entries.empty?
```
MikeMcQuaid: `if .any?` or better still do an early return?

lib/bundle/installer.rb
```diff
@@ -45,7 +51,14 @@ def install(entries)
...
+        puts Formatter.error("Homebrew Bundle encountered some errors! #{errored_entries.size} Brewfile #{Bundle::Dsl.pluralize_dependency(errored_entries.size)} failed badly:")
```
MikeMcQuaid: This is a very long line (both in terms of the code and the output). Could both be tweaked?

spec/bundle/commands/install_command_spec.rb
```diff
@@ -14,6 +14,16 @@
     end
   end
 
+  context "when an installer raises an error" do
+    it "does not bubble the error to the top" do
```
MikeMcQuaid: `it rescues the error`

colindean:
Might take a look at this this weekend.

lib/bundle/installer.rb
colindean: I think so. I'm thinking I'm going to refactor the whole big method.

colindean:
Big refactor of `Bundle::Installer` is up for review. The test is still failing but it should be a little more diagnosable now. The coverage will be under because of two impossible-to-reach sections that I will probably eliminate after I'm more confident that they are unreachable once I've got the test passing!

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+        skipped = end_state[:skipped] || 0
+        installed = succeeded + skipped
+        puts Formatter.success("Homebrew Bundle complete! #{installed} Brewfile #{Bundle::Dsl.pluralize_dependency(installed)} now installed.")
       else
-        puts Formatter.error("Homebrew Bundle failed! #{failure} Brewfile #{Bundle::Dsl.pluralize_dependency(failure)} failed to install.")
+        puts Formatter.error("Homebrew Bundle failed! #{failures} Brewfile #{Bundle::Dsl.pluralize_dependency(failures)} failed to install.")
+      end
+    end
+
+    def report_errors(errored_entries)
```
MikeMcQuaid: Why separate reporting failures and reporting errors?

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+      words = Bundle::Dsl.pluralize_dependency(error_count)
+      error_text = "Homebrew Bundle encountered some errors. #{error_count} Brewfile #{words} failed badly:"
+      puts Formatter.error(error_text)
+      errored_entries.each do |entry, error|
+        puts Formatter.error("\t#{entry}\t => \t#{error}")
       end
+    end
 
-      failure.zero?
+    InstallerTask = Struct.new :cls, :arg, :verb, :entry do
```
MikeMcQuaid: I don't understand what's going on here.

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
-          success += 1
+      end_state = install_entries(entries)
+
+      failures = end_state[:failed]
+      report_success_and_failure(end_state)
+
+      errored_entries = end_state[:errored]
+      report_errors(errored_entries) if errored_entries.any?
+
+      failures.zero? && errored_entries.empty?
```
MikeMcQuaid: Why make an (internal) API where you're checking for `zero?` for one and `empty?` for the other?

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+      when :skipped
+        puts "Using #{task.entry.name}"
+      when :failed
+        puts Formatter.error("#{task.header} has failed!")
+      else
+        puts Formatter.error("#{task.header} in unknown state: #{result}")
+      end
+    end
+
+    def report_error(task, error)
```
MikeMcQuaid: Why separate this from `report_result`?

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+      puts Formatter.error("#{task.header} raised an exception: #{error}")
+    end
+
+    def execute_task(task)
+      begin
+        result = task.cls.install(*task.arg)
+        report_result(task, result)
+        { result => 1 }
```
MikeMcQuaid: It's unclear what `1` means in this context.

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+        { errored: { task.entry.name => error } }
+      end
+    end
+
+    def install_entries(entries)
+      initial_state = {
+        success: 0,
+        skipped: 0,
+        failed: 0,
+        errored: {}
```
MikeMcQuaid: It's unclear why `error` is not also an integer in this context.

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
+        result = execute_task(task)
+        combine_states(state, result)
+      end
+    end
+
+    def combine_states(state, result)
+      state.merge(result) do |_, old, new|
+        case old
+        when Integer # for the counts
+          old + new
```
MikeMcQuaid: This is confusing.

lib/bundle/installer.rb
```diff
@@ -5,47 +5,121 @@ module Installer
...
       end
+    end
 
-      if failure.zero?
-        puts Formatter.success("Homebrew Bundle complete! #{success} Brewfile #{Bundle::Dsl.pluralize_dependency(success)} now installed.")
+    def report_success_and_failure(end_state)
+      failures = end_state[:failed]
+      if failures.zero?
+        succeeded = end_state[:success] || 0
+        skipped = end_state[:skipped] || 0
```
MikeMcQuaid: When would these be `nil`?

spec/bundle/commands/install_command_spec.rb
```diff
@@ -14,6 +14,16 @@
     end
   end
 
+  context "when an installer raises an error" do
+    it "rescues the error" do
+      allow(ARGV).to receive(:value).and_return(nil)
+      allow(Bundle::MacAppStoreInstaller).to receive(:install).and_throw(RuntimeError)
```
MikeMcQuaid: This doesn't seem to describe the issue that's mentioned in https://github.com/Homebrew/homebrew-bundle/issues/456. `mas` will only throw Ruby exceptions on invalid format in the `Brewfile`; I'm not sure I understand how catching and reporting this helps?

MikeMcQuaid:
Revisiting https://github.com/Homebrew/homebrew-bundle/issues/456; thanks for the PR but let's discuss more in there before you go further on here.

lib/bundle/installer.rb
colindean: Struct is a convenient way to create a basic class and give it simple functionality. This is equivalent to

```ruby
class InstallerTask
  attr_reader :cls, :arg, :verb, :entry

  def initialize cls, arg, verb, entry
    @cls = cls
    @arg = arg
    @verb = verb
    @entry = entry
  end

  #… other methods inside the block passed to Struct.new

end
```

I chose to go this route because keeping that info in a hash was getting verbose.

lib/bundle/installer.rb
colindean: Until now, the successes, skipped, and failures have only been tracked as a count instead of the formula name and maybe options. Errors need that metadata in order to be displayed later.

I'd welcome keeping more information about successes, skipped, and failure even if that information isn't used.

Maybe there's insufficient dissimilarity between failures and errors that I'm missing in how I've approached the problem?

lib/bundle/installer.rb
colindean: Until now, the successes, skipped, and failures have only been tracked as a count instead of the formula name and maybe options. Errors need that metadata in order to be displayed later.

It's not very clean, but the goal here was to keep the existing container of status information. If I switch to tracking the successes, skipped, and failures by name, this would be cleaner.

lib/bundle/installer.rb
colindean: It's a result of mixing value types in the status tracking hash. I dislike it but it felt like a minimum viable change.

If I move away from that Int|Hash hash, this will go away.

lib/bundle/installer.rb
colindean: I saw it in some tests but it might be no longer possible after I fixed something else. Both guards could probably be removed.

lib/bundle/installer.rb
colindean: The return of this is getting merged into the status hash. It's an unclear vestige that will go away if I move away from the status Int|Hash hash as it currently exists on lines 70-75.

lib/bundle/installer.rb
colindean: The error has additional metadata. If I merged the two methods, it'd need to look like:

```ruby
def report_result(task, result, message=nil)
```

and be invoked with

```
report_result(task, result) # when result in :success, :skipped, :failed
report_result(task, :error, error.to_s) # when there was an exception
```

That felt and feels… weird. It's a display method, so passing the string to be displayed would be OK. 

Maybe it'll be better if I refactor a bit to eliminate the Int|Hash status hash.

PR #490, opened 2019-05-17
Replace whitelist for allow-list

leonardobazico:


README.md
```diff
@@ -67,7 +67,7 @@ The `--no-restart` option will prevent `restart_service` from being added to ``b
 
 ### Cleanup
 
-You can also use `Brewfile` as a whitelist. It's useful for maintainers/testers who regularly install lots of formulae. To uninstall all Homebrew formulae not listed in `Brewfile`:
+You can also use `Brewfile` as a allow-list. It's useful for maintainers/testers who regularly install lots of formulae. To uninstall all Homebrew formulae not listed in `Brewfile`:
```
MikeMcQuaid: I'm ok with changing `whitelist` but this naming is clunky (and should be `an allow-list`). Any other suggestions?

leonardobazico: Maybe we can use trust-list or safe-list.

colindean: I decided to educate myself on the best practice of the proposed alternatives.

* https://english.stackexchange.com/questions/51088/alternative-term-to-blacklist-and-whitelist
* https://ux.stackexchange.com/questions/14812/alternative-term-to-blacklist-and-whitelist
* http://garysaid.com/are-the-terms-whitelist-and-blacklist-racist/
* https://github.com/hapijs/joi/pull/1636 which resulted in https://github.com/hapijs/joi/pull/1662

If there is to be a change to this, perhaps let's use terminology that leaves no question as to what the function is. Maybe "exclusive" is better in context.

```suggestion
You can also use `Brewfile` to list the packages that should be installed exclusively, removing any package not present or dependent. This workflow is useful for maintainers or testers who regularly install lots of formulae. To uninstall all Homebrew formulae not listed in `Brewfile`:
```

MikeMcQuaid: Thanks @colindean! What about `to list the only packages`?

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @leonardobazico!

PR #494, opened 2019-05-23
.github/: defer to Homebrew/.github repo.

MikeMcQuaid:
This allows maintaining more settings in one place.


jacobbednarz:
If we needed to, we also could use something like https://github.com/probot/probot-config to have a parent config and override the local parts that needed some customisation.

MikeMcQuaid:
@jacobbednarz Oooh, good idea.

Issue #368, opened 2018-09-14
Adding an ENV var for the `--global` flag


aramprice:
I'm curious if there is any interest in being able to trigger the `--global` behavior using and ENV var. Something like `HOMEBREW_BUNDLE_GLOBAL=1`.

colindean:
Maybe? What about an environment variable that lets the user specify which file to use, and the user can choose to use the global path or some other one?

MikeMcQuaid:
I don't think so personally; it's not clear what should happen when you have that variable set and you specify a `--file` flag.

aramprice:
We already have this lack of clarity when both `--global` and `--file` are specified. As the [code is today](https://github.com/Homebrew/homebrew-bundle/blob/bc97597d437f32712791d5b68d558da1798d34e1/lib/bundle/bundle.rb#L48) `--global` wins. Maybe `--global` should be removed in favor of `--file`?

I'm interested in being able to specify a `Brewfile` via an ENV var. If the project is interested in a PR for this perhaps we can work out how best to make this possible.

MikeMcQuaid:
@aramprice We won't be removing `--global` as it's widely used.



> If the project is interested in a PR for this perhaps we can work out how best to make this possible.

That sounds reasonable, particularly if you address the conflicts when multiple things (e.g. `--global` and `--file`) are specified at once and error out appropriately.

aramprice:
Just letting you know that I'm chipping away at this, slowly. I have some travel coming up in October perhaps by mid-October.

Issue #365, opened 2018-09-06
Feature Request: subcommands to add/remove packages


tycobbb:
It would be nice for `brew bundle` to include subcommands to install/uninstall packages while keeping corresponding Brewfile up-to-date. Something similar to the add/remove functionality in bundler or npm. 

Is this feature significant enough to be considered for inclusion? If it's not high-enough priority, would a PR be accepted?

### Motivation
While `brew bundle` is great, keeping a Brewfile updated independent of the normal brew installation procedure is tedious/error-prone. It is especially painful when trying to maintain a consistent environment across multiple machines. 

A personal use-case: when testing new tools I tend to install and then quickly remove them once I decide they don't meet my needs. I'm inconsistent remembering, once I decide to keep them, to also update my global `.Brewfile`. The environments on my machines tend to slowly diverge.

### Proposal
Add two subcommands to `brew bundle`:

```sh
brew bundle add <package-name>
brew bundle remove <package-name>
```

Any `brew bundle` options would be applied to map to the correct Brewfile. The rest of the options could be passed through to the root install/uninstall commands.

#### Alternatives
Add npm-like `--save=<brewfile>` and `--save-global` options to `brew install`
- `+` Surfaces `brew bundle` functionality to more users. It's great, but hard-to-discover right now
- `-` Can't think of an unsurprising way for `brew uninstall` to clean up the correct Brewfile
- `-` Top-level commands now require a plugin, inverts control

Use `install/uninstall` instead of `add/remove` to maintain core `brew` jargon
- `-` Install already has a different meaning for `brew bundle`
- `-` Doesn't signal the Brewfile mutation as well as `add/remove`

colindean:
I would very much like this feature.

MikeMcQuaid:
> Is this feature significant enough to be considered for inclusion? If it's not high-enough priority, would a PR be accepted?

Given @colindean is 👍: yeh, I think it'd be suitable to at least take a look at a potential implementation. Personally I don't see it being much easier than adding the line manually to the `Brewfile` but I guess I'm not the target audience for this feature and if `bundler` and `npm` have something similar that's a decent argument for it.

Bear in mind you'll need to figure out a syntax to be able to add `brew`/`tap`/etc. lines and ideally specify options as well.

colindean:
I think it could work explicitly and implicitly.

## Explicit operation

    brew bundle add packagename
    brew bundle add packagename --with-only-the-best-options
    brew bundle add cask packagename
[...]

PR #502, opened 2019-06-11
Allow `Brewfile` location to be specified in ENV

aramprice:
Adds support for `HOMEBREW_BUNDLE_FILE` as a way to specify the location
of the `Brewfile` to be used by `brew bundle` commands.

Specifying `HOMEBREW_BUNDLE_FILE` when `--global` or `--file` are also
specified results in an error.

Existing behavior wherein `--global` overrides `--file` with no error is
preserved. This could be revisited in the future, the code in
`brewfile.rb` would likely need further refactoring.

* This addresses an issue initially raised in https://github.com/Homebrew/homebrew-bundle/issues/368.

### DIFF
--- a/lib/bundle/brewfile.rb
+++ b/lib/bundle/brewfile.rb
@@ -5,15 +5,15 @@ module Bundle
     module_function
 
     def path(dash_writes_to_stdout: false)
-      env_bundle_file = ENV.fetch("HOMEBREW_BUNDLE_FILE", "")
+      env_bundle_file = ENV["HOMEBREW_BUNDLE_FILE"]
 
       filename =
           if ARGV.include?("--global")
-            raise "'HOMEBREW_BUNDLE_FILE' can not be specified with '--global'" unless env_bundle_file.empty?
+            raise "'HOMEBREW_BUNDLE_FILE' can not be specified with '--global'" if env_bundle_file.present?
             "#{ENV["HOME"]}/.Brewfile"
           elsif ARGV.include?("--file")
             handle_file_value(ARGV.value("file"), dash_writes_to_stdout)
-          elsif !env_bundle_file.empty?
+          elsif env_bundle_file.present?
             env_bundle_file
           else
             "Brewfile"
--- a/lib/bundle/brewfile.rb
+++ b/lib/bundle/brewfile.rb
@@ -5,26 +5,37 @@ module Bundle
     module_function
 
     def path(dash_writes_to_stdout: false)
-      if ARGV.include?("--global")
-        Pathname.new("#{ENV["HOME"]}/.Brewfile")
-      else
-        filename = ARGV.value("file")
-        if filename == "-"
-          filename = if dash_writes_to_stdout
-            "/dev/stdout"
+      env_bundle_file = ENV.fetch("HOMEBREW_BUNDLE_FILE", "")
+
+      filename =
+          if ARGV.include?("--global")
+            raise "'HOMEBREW_BUNDLE_FILE' can not be specified with '--global'" unless env_bundle_file.empty?
+            "#{ENV["HOME"]}/.Brewfile"
+          elsif ARGV.include?("--file")
+            handle_file_value(ARGV.value("file"), dash_writes_to_stdout)
+          elsif !env_bundle_file.empty?
+            env_bundle_file
           else
-            "/dev/stdin"
+            "Brewfile"
           end
-        end
-        filename ||= "Brewfile"
-        Pathname.new(filename).expand_path(Dir.pwd)
-      end
+
+      Pathname.new(filename).expand_path(Dir.pwd)
     end
 
     def read
       Brewfile.path.read
     rescue Errno::ENOENT
       raise "No Brewfile found"
+    end
+
+    def handle_file_value(filename, dash_writes_to_stdout)
+      if filename != "-"
+        filename
+      elsif dash_writes_to_stdout
+        "/dev/stdout"
+      else
+        "/dev/stdin"
+      end
     end
   end
 end
--- a/spec/bundle/brewfile_spec.rb
+++ b/spec/bundle/brewfile_spec.rb
@@ -4,7 +4,7 @@ require "spec_helper"
 
 describe Bundle::Brewfile do
   describe "path" do
-    let(:env_bundle_file_value) { "" }
+    let(:env_bundle_file_value) { nil }
     let(:has_file) { false }
     let(:file_value) { "" }
     let(:has_global) { false }
@@ -14,7 +14,15 @@ describe Bundle::Brewfile do
       allow(ARGV).to receive(:value).with("file").and_return(file_value)
       allow(ARGV).to receive(:include?).with("--global").and_return(has_global)
 
-      allow(ENV).to receive(:fetch).with("HOMEBREW_BUNDLE_FILE", "").and_return(env_bundle_file_value)
+      original_method = ENV.method(:[])
+      allow(ENV).to receive(:[]) do |env_string|
+        case env_string
+        when "HOMEBREW_BUNDLE_FILE"
+          env_bundle_file_value
+        else
+          original_method.call(env_string)
+        end
+      end
     end
 
     context "when `--file` is passed" do
@@ -35,6 +43,14 @@ describe Bundle::Brewfile do
             expect(described_class.path).to eq(expected_pathname)
           end
         end
+
+        context "and HOMEBREW_BUNDLE_FILE is `` (empty)" do
+          let(:env_bundle_file_value) { "" }
+
+          it "returns the value specified by `--file` path" do
+            expect(described_class.path).to eq(expected_pathname)
+          end
+        end
       end
 
       context "with an absolute path" do
@@ -47,6 +63,14 @@ describe Bundle::Brewfile do
 
         context "and HOMEBREW_BUNDLE_FILE is set" do
           let(:env_bundle_file_value) { "/path/to/Brewfile" }
+
+          it "returns the value specified by `--file` path" do
+            expect(described_class.path).to eq(expected_pathname)
+          end
+        end
+
+        context "and HOMEBREW_BUNDLE_FILE is `` (empty)" do
+          let(:env_bundle_file_value) { "" }
 
           it "returns the value specified by `--file` path" do
             expect(described_class.path).to eq(expected_pathname)
@@ -71,6 +95,14 @@ describe Bundle::Brewfile do
           end
         end
 
+        context "and HOMEBREW_BUNDLE_FILE is `` (empty)" do
+          let(:env_bundle_file_value) { "" }
+
+          it "returns the value specified by `--file` path" do
+            expect(described_class.path).to eq(expected_pathname)
+          end
+        end
+
         context "when `dash_writes_to_stdout` is true" do
           let(:expected_pathname) { Pathname.new("/dev/stdout") }
 
@@ -87,14 +119,40 @@ describe Bundle::Brewfile do
             expect(described_class.path(dash_writes_to_stdout: true)).to eq(expected_pathname)
           end
         end
+
+        context "when `dash_writes_to_stdout` is true and HOMEBREW_BUNDLE_FILE is `` (empty)" do
+          let(:expected_pathname) { Pathname.new("/dev/stdout") }
+          let(:env_bundle_file_value) { "" }
+
+          it "returns the value specified by `--file` path" do
+            expect(described_class.path(dash_writes_to_stdout: true)).to eq(expected_pathname)
+          end
+        end
       end
     end
 
     context "when `--global` is passed" do
       let(:has_global) { true }
+      let(:expected_pathname) { Pathname.new("#{ENV["HOME"]}/.Brewfile") }
 
       it "returns the expected path" do
-        expect(described_class.path).to eq(Pathname.new("#{ENV["HOME"]}/.Brewfile"))
+        expect(described_class.path).to eq(expected_pathname)
+      end
+
+      context "and HOMEBREW_BUNDLE_FILE is set" do
+        let(:env_bundle_file_value) { "/path/to/Brewfile" }
+
+        it "returns the value specified by `--file` path" do
+          expect { described_class.path }.to raise_error(RuntimeError)
+        end
+      end
+
+      context "and HOMEBREW_BUNDLE_FILE is `` (empty)" do
+        let(:env_bundle_file_value) { "" }
+
+        it "returns the value specified by `--file` path" do
+          expect(described_class.path).to eq(expected_pathname)
+        end
       end
     end
 
@@ -107,6 +165,14 @@ describe Bundle::Brewfile do
 
       context "that is `` (empty)" do
         let(:env_bundle_file_value) { "" }
+
+        it "defaults to `${PWD}/Brewfile`" do
+          expect(described_class.path).to eq(Pathname.new("Brewfile").expand_path(Dir.pwd))
+        end
+      end
+
+      context "that is `nil`" do
+        let(:env_bundle_file_value) { nil }
 
         it "defaults to `${PWD}/Brewfile`" do
           expect(described_class.path).to eq(Pathname.new("Brewfile").expand_path(Dir.pwd))
--- a/spec/bundle/brewfile_spec.rb
+++ b/spec/bundle/brewfile_spec.rb
@@ -4,49 +4,113 @@ require "spec_helper"
 
 describe Bundle::Brewfile do
   describe "path" do
+    let(:env_bundle_file_value) { "" }
+    let(:has_file) { false }
+    let(:file_value) { "" }
+    let(:has_global) { false }
+
+    before do
+      allow(ARGV).to receive(:include?).with("--file").and_return(has_file)
+      allow(ARGV).to receive(:value).with("file").and_return(file_value)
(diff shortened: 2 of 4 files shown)


lib/bundle/brewfile.rb
```diff
@@ -5,19 +5,28 @@ module Brewfile
...
-      if ARGV.include?("--global")
-        Pathname.new("#{ENV["HOME"]}/.Brewfile")
-      else
-        filename = ARGV.value("file")
-        if filename == "-"
-          filename = if dash_writes_to_stdout
-            "/dev/stdout"
-          else
-            "/dev/stdin"
+      bundle_file = "#{ENV["HOMEBREW_BUNDLE_FILE"]}"
```
colindean:
```suggestion
      bundle_file = ENV["HOMEBREW_BUNDLE_FILE"]
```

I think I'd prefer this to be named `env_bundle_file` or the like to make its origin clear since it is not mutated later.

colindean:
Thinking about how I'd use this _reasonably˚_ myself, I see this pattern:

* `HOMEBREW_BUNDLE_FILE` is set to `~/my_dotfiles/Brewfile.work` on my work machine.
* `HOMEBREW_BUNDLE_FILE` is set to `~/my_dotfiles/Brewfile.home` on my home machine.
* Project has a `Brewfile` in the root of the repo ➡️ Use automatic `./Brewfile`
* Project has a `Brewfile` but it's not in the root directory of the repo ➡️ specify `--file path/to/Brewfile`
* ❌ because then I can't use `--file` to specify when I'm using the envvar.

This feels like a reasonable pattern for me, esp. if we were to implement #365 adding a CLI for Brewfile management. As written, this usage pattern would break because of `--file` and envvar together erroring instead of the CLI argument taking precedence.

I think it makes sense for the envvar and `--global` to blow up when used together. _I_ don't see a usage pattern where someone would _want_ to use both, which would be maintaining _two_ active "master" Brewfiles per machine. 

_˚ I could get real creative but 90% of our users won't._

MikeMcQuaid:
Great work here @aramprice and a great idea!

> * ❌ because then I can't use `--file` to specify when I'm using the envvar.

Agreed that it'd be nice to handle the `--file` case. I'm also on the fence about `--global`.

lib/bundle/brewfile.rb
aramprice: I like the rename to `env_bundle_file`.

I added the interpolation to prevent having to distinguish between `nil` vs `''` (empty string). That being said I didn't look into whether `ENV['FOO']` would return `nil` in cases where `FOO` was not set.

aramprice:
I think relaxing what I have to allow `HOMEBREW_BUNDLE_FILE` to co-exist with `--file`, where `--file` wins is reasonable.

I suspect this will be a more invasive change. Would all you prefer that I PR some structural refactors to and then PR a feature change?

I'm also curious about the build failure. Looking at the logs it seems to be rubocop issues but when I run `brew style homebrew/bundle` there are no warnings (see below). Is there an update to the style checker that I need to do?

```
→ brew style homebrew/bundle

20 files inspected, no offenses detected

10:14:32 laptop in ~/workspace/homebrew-bundle
```

lib/bundle/brewfile.rb
aramprice: I could implement this as `ENV.fetch("HOMEBREW_BUNDLE_FILE", "")` which might be more clear. It would also make it easier to mock, as well as test for `nil` vs `""` (empty string).

aramprice:
Reminder that currently if both `--global` and `--file` are specified the `--global` flag wins. This is the existing behavior.

lib/bundle/brewfile.rb
MikeMcQuaid: > I added the interpolation to prevent having to distinguish between nil vs '' (empty string).

MikeMcQuaid: > I added the interpolation to prevent having to distinguish between nil vs '' (empty string).

Can use `.blank?` instead.

> That being said I didn't look into whether `ENV['FOO']` would return `nil` in cases where `FOO` was not set.

It does.

aramprice: Sad days `undefined method `blank?' for "":String`. I'll try `.fetch()`

MikeMcQuaid:
> Would all you prefer that I PR some structural refactors to and then PR a feature change?

Fine either way with me, whatever you'd rather.

> I'm also curious about the build failure. Looking at the logs it seems to be rubocop issues but when I run `brew style homebrew/bundle` there are no warnings (see below). Is there an update to the style checker that I need to do?

That's not quite sufficient. Running `bundle exec rubocop` from the repository root should be.

> Reminder that currently if both `--global` and `--file` are specified the `--global` flag wins. This is the existing behavior.

Perhaps `--file` should or it should error? I'm open to either option if it makes the PR easier/better or the behaviour make more sense.

Thanks again @aramprice!

aramprice:
I tried running `bundle exec rubocop` and got:

```
Inspecting 60 files
CCCCCCCCCCCWCCCCCCCWWCCCC..C..CCCCWCCWCWCWWWCCCWCCWWCCCWCCWC

Offenses:
# ... SNIPPED....
60 files inspected, 2315 offenses detected
```

The CI system is reporting:
```
2019-06-12T17:23:48.2726270Z More information: https://github.com/rubocop-hq/rubocop-rails
2019-06-12T17:23:55.0679970Z == /usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/spec/bundle/brewfile_spec.rb ==
2019-06-12T17:23:55.0680760Z C: 81: 17: Space missing to the left of {.
2019-06-12T17:23:55.0687980Z C: 89: 17: Space missing to the left of {.
2019-06-12T17:23:55.0688340Z 
2019-06-12T17:23:55.0688650Z 21 files inspected, 2 offenses detected
```

but doesn't print out what command is actually running.

I can probably fix errors after they're reported by CI but it would be nice to be able to check locally.

MikeMcQuaid:
https://github.com/Homebrew/homebrew-bundle/blob/master/azure-pipelines.yml details what CI is running.

jacobbednarz:
The logic here sounds reasonable however it did take me a little while to get my head around the precedence stuff. I wonder if it's worth explicitly documenting the order and what overrides what to ensure we maintain it? Ideally some test cases can help be useful but documentation is probably a better first step.

aramprice:
In case folks have troubles in the future with `brew style homebrew/bundle` the trick is to run `brew update-reset`.

aramprice:
@jacobbednarz I think I've made the code more clear at the expense of readability of the PR.

The cascade is as follows:
- `--global` first
  - `--global` + `HOMEBREW_BUNDLE_FILE` results in an error
- `--file` trumps `HOMEBREW_BUNDLE_FILE`; no error
- `HOMEBREW_BUNDLE_FILE` if there are no other flags
- otherwise `${PWD}/Brewfile`

The function `handle_file_value` should probably be private but I wasn't sure how this interacted with `module_function`.

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
 
 module Bundle
   module Brewfile
+    ENV_KEY = "HOMEBREW_BUNDLE_FILE"
     module_function
 
     def path(dash_writes_to_stdout: false)
+      env_bundle_file = ENV.fetch(ENV_KEY, "")
```
colindean: 👍

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
...
-        if filename == "-"
-          filename = if dash_writes_to_stdout
-            "/dev/stdout"
-          else
-            "/dev/stdin"
-          end
-        end
-        filename ||= "Brewfile"
+      elsif ARGV.include?("--file")
+        filename = handle_file_value(ARGV.value("file"), dash_writes_to_stdout)
```
colindean: Looks like you could move the call to `ARGV.value("file")` here into the `handle_file_value` method since that method isn't being invoked elsewhere.

aramprice: That makes sense. I chose to keep `ARGV` access within `#path` so that all access to `ARGV` could be seen at a glance in a single method.

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
 
 module Bundle
   module Brewfile
+    ENV_KEY = "HOMEBREW_BUNDLE_FILE"
```
MikeMcQuaid: Don't think this is worth making a constant for a single use in this file.

lib/bundle/brewfile.rb
MikeMcQuaid: Use `ENV["HOMEBREW_BUNDLE_FILE"]`

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
...
     module_function
 
     def path(dash_writes_to_stdout: false)
+      env_bundle_file = ENV.fetch(ENV_KEY, "")
+
       if ARGV.include?("--global")
+        raise "'#{ENV_KEY}' can not be specified with '--global'" unless env_bundle_file.empty?
```
MikeMcQuaid: Again: use `.blank?` please.

lib/bundle/brewfile.rb
MikeMcQuaid: I'm fine with either approach.

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
...
-            "/dev/stdout"
-          else
-            "/dev/stdin"
-          end
-        end
-        filename ||= "Brewfile"
+      elsif ARGV.include?("--file")
+        filename = handle_file_value(ARGV.value("file"), dash_writes_to_stdout)
         Pathname.new(filename).expand_path(Dir.pwd)
+      elsif !env_bundle_file.empty?
```
MikeMcQuaid: `elsif env_bundle_file.present?`

lib/bundle/brewfile.rb
```diff
@@ -2,22 +2,22 @@
 
 module Bundle
   module Brewfile
+    ENV_KEY = "HOMEBREW_BUNDLE_FILE"
     module_function
 
     def path(dash_writes_to_stdout: false)
+      env_bundle_file = ENV.fetch(ENV_KEY, "")
+
       if ARGV.include?("--global")
```
MikeMcQuaid: Could assign all the results to a variable and then `Pathname.new` that variable to avoid `Pathname.new` being repeated in each case.

lib/bundle/brewfile.rb
aramprice: `.blank?` doesn't work.

```
ENV["HOME"].blank?
NoMethodError: undefined method `blank?' for "/Users/aram":String
	from (irb):2
	from /usr/bin/irb:11:in `<main>'
```

lib/bundle/brewfile.rb
aramprice: This postpones distinguishing between `nil` and `""` (empty string). I'm using `.fetch` to ensure that subsequent access to `env_bundle_file` is guaranteed to be operating on `String`.

lib/bundle/brewfile.rb
aramprice: That makes sense. It looks like the `.expand_path(Dir.pwd)` can safely be added to `Pathname.new("#{ENV["HOME"]}/.Brewfile")`.

lib/bundle/brewfile.rb
aramprice: For one I agree. There are two so maybe its arguable.

lib/bundle/brewfile.rb
MikeMcQuaid: Use `brew irb` instead. It's a ActiveSupport method included by Homebrew.

lib/bundle/brewfile.rb
aramprice:
```
→ bundle exec irb
irb(main):001:0> "".present?
NoMethodError: undefined method `present?' for "":String
# ...
```

jacobbednarz: @aramprice You'll need `brew irb` as @MikeMcQuaid mentioned in https://github.com/Homebrew/homebrew-bundle/pull/502#discussion_r293581893 as it's included by Homebrew, not just bundler like you're attempting here.

aramprice: These methods are present when I run `brew irb` but tests fail when I run `bundle exec rake`. This appears to be what is [run in CI](https://github.com/Homebrew/homebrew-bundle/blob/master/azure-pipelines.yml#L13) so my expectation is that CI will not pass with `.blank?` or `.present?`

lib/bundle/brewfile.rb
aramprice: `.blank?` works when I run `brew irb` but tests fail when I run `bundle exec rake`.

lib/bundle/brewfile.rb
aramprice: See https://github.com/Homebrew/homebrew-bundle/pull/502#discussion_r293596295 regarding problems using `.blank?` or `.present?`. If these can be used I'm happy to change from `ENV.fetch()` to `ENV[]`.

lib/bundle/brewfile.rb
MikeMcQuaid: Yes, it will need mocked in the unit tests.

lib/bundle/brewfile.rb
```diff
@@ -2,29 +2,39 @@
 
 module Bundle
   module Brewfile
+    BUNDLE_FILE_ENV_KEY = "HOMEBREW_BUNDLE_FILE"
```
MikeMcQuaid: Again, please hardcode this: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

lib/bundle/brewfile.rb
```diff
@@ -2,29 +2,39 @@
...
-      else
-        filename = ARGV.value("file")
-        if filename == "-"
-          filename = if dash_writes_to_stdout
-            "/dev/stdout"
+      env_bundle_file = ENV[BUNDLE_FILE_ENV_KEY]
+
+      filename =
+          if ARGV.include?("--global")
+            raise "'#{BUNDLE_FILE_ENV_KEY}' can not be specified with '--global'" unless env_bundle_file.blank?
```
MikeMcQuaid: `if env_bundle_file.present?` to avoid double negative.

lib/bundle/brewfile.rb
```diff
@@ -2,29 +2,39 @@
...
     end
 
     def read
       Brewfile.path.read
     rescue Errno::ENOENT
       raise "No Brewfile found"
     end
+
+    def handle_file_value(filename, dash_writes_to_stdout)
+      if filename == "-"
```
MikeMcQuaid: `if filename != "-"` and then could do `elsif dash_writes_to_stdout` and `else` which would make the function a little more readable.

spec/bundle/brewfile_spec.rb
```diff
@@ -4,49 +4,190 @@
 
 describe Bundle::Brewfile do
   describe "path" do
-    context "when `--file` is passed" do
-      before do
-        allow(ARGV).to receive(:include?).with("--global").and_return(false)
-        allow(ARGV).to receive(:value).with("file").and_return(file_value)
+    # monkeypatch Object so it behaves as if ActiveSupport were present
+    class Object
```
MikeMcQuaid: Either put this in `spec/stub` or use `allow` or `expect`

lib/bundle/brewfile.rb
aramprice: as you like

MikeMcQuaid:
Thanks for all this @aramprice and sorry for all the back and forth. Great work here, you rock!

lib/bundle/brewfile.rb
```diff
@@ -5,26 +5,37 @@ module Brewfile
...
-        if filename == "-"
-          filename = if dash_writes_to_stdout
-            "/dev/stdout"
+      env_bundle_file = ENV["HOMEBREW_BUNDLE_FILE"]
+
+      filename =
+          if ARGV.include?("--global")
+            raise "'HOMEBREW_BUNDLE_FILE' can not be specified with '--global'" if env_bundle_file.present?
+            "#{ENV["HOME"]}/.Brewfile"
+          elsif ARGV.include?("--file")
```
andyli: When using `brew bundle --file=path/to/Brewfile`, `ARGV` will be `["--file=path/to/Brewfile"]`. So, the check of `ARGV.include?("--file")` is wrong.

colindean:
Thank you for your hard work on this, @aramprice! I appreciate you!

lib/bundle/brewfile.rb
aramprice: Thanks for pointing this out. It seems like this might need to be adjusted in the mainline now that the PR has been closed.

danielbayley:
> Great work here @aramprice and a great idea!

@MikeMcQuaid What changed between [then](https://github.com/Homebrew/homebrew-bundle/pull/261) (#261) and now? 😅

Anyway, glad this was finally allowed… nice one @aramprice!

MikeMcQuaid:
@danielbayley 2.5 years and a bunch more people asking for it. I realise it must be frustrating but it's general good product practice to want to hear demand from more users before implementing new features.

danielbayley:
> I realise it must be frustrating

@MikeMcQuaid Not a big deal 😉

Issue #506, opened 2019-06-17
Homebrew Bundle installs fonts that are already installed


leafac:
Create a `Brewfile`:

```ruby
tap "caskroom/fonts"
cask "font-charter"
```

Run Homebrew Bundle twice:

```console
$ brew bundle
Using caskroom/fonts
Installing font-charter
Homebrew Bundle complete! 2 Brewfile dependencies now installed.
$ brew bundle
Using caskroom/fonts
Installing font-charter
Homebrew Bundle complete! 2 Brewfile dependencies now installed.
```

Note how the second time it doesn’t say `Using font-charter`, but `Installing font-charter`, which means it’s installing the cask again.

Nothing breaks by installing a cask again, but if you have many fonts in your `Brewfile`, it adds up and gets quite slow.

MikeMcQuaid:
What does `brew bundle dump` output?

leafac:
Interesting.

Running `brew bundle dump` I get the following `Brewfile`:

```ruby
tap "caskroom/cask"
tap "caskroom/fonts"
tap "homebrew/bundle"
tap "homebrew/cask"
tap "homebrew/core"
brew "bash"
brew "caddy"
brew "cloc"
brew "coreutils"
brew "exim"
brew "ffmpeg"
brew "figlet"
brew "git"
brew "graphviz"
brew "imagemagick"
brew "nmap"
brew "node"
brew "pstree"
brew "rbenv"
brew "rlwrap"
brew "tree"
brew "watch"
brew "wget"
brew "yarn"
brew "youtube-dl"
cask "authy"
cask "backblaze"
cask "cyberduck"
cask "day-o"
cask "detexify"
cask "firefox"
cask "google-chrome"
cask "grandperspective"
cask "inkscape"
cask "intellij-idea"
cask "java"
cask "mactex"
cask "mattermost"
cask "postman"
cask "racket"
cask "screenflow"
cask "shiftit"
cask "skype"
cask "the-unarchiver"
cask "tor-browser"
cask "transmission"
cask "visual-studio-code"
cask "vlc"
cask "xquartz"
cask "zotero"
cask "caskroom/fonts/font-abril-fatface"
cask "caskroom/fonts/font-alegreya"
cask "caskroom/fonts/font-alegreya-sans"
cask "caskroom/fonts/font-andada"
cask "caskroom/fonts/font-andada-sc"
cask "caskroom/fonts/font-andika"
[...]

PR #513, opened 2019-06-18
Add example of Caskroom tap to `Brewfile` in README

leafac:
See https://github.com/Homebrew/homebrew-bundle/issues/506


MikeMcQuaid:
Thanks @leafac!

PR #533, opened 2019-07-25
Linux support

MikeMcQuaid:
Handle and document enough Linux functionality to call it "supported".

### DIFF
--- a/lib/bundle/extend/os/linux/skipper.rb
+++ b/lib/bundle/extend/os/linux/skipper.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Skipper
+    class << self
+      def skip?(entry)
+        return generic_skip?(entry) unless [:cask, :mas].include?(entry.type)
+
+        puts Formatter.warning "Skipping #{entry.name} (on Linux)"
+        true
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -6,8 +6,7 @@ Bundler for non-Ruby dependencies from Homebrew.
 
 ## Requirements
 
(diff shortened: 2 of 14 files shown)


issyl0:
At the moment my approach to this is a load of `if OS.mac?` conditions in Brewfiles, so this looks great! I guess we’ll still have to put Linux-only formulae (alternatives to casks on Linux, ie “linuxbrew/extra/aws-vault”) behind an `if OS.linux?` conditional. But this definitely tidies things up. Thanks for doing this!

issyl0:
I just tested this locally on Linux with a three line test Brewfile with `brew` and `cask` entries, and I get:

```
$ brew bundle --file=BrewfileTest
Error: undefined local variable or method `generic_skip' for Bundle::Skipper:Module
```

lib/bundle/extend/os/linux/skipper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Skipper
+    module_function
+
+    def skip?(entry)
+      return generic_skip unless [:cask, :mas].include?(entry.type)
```
colindean: I think this will address @issyl0's problem:

```suggestion
      return generic_skip? unless [:cask, :mas].include?(entry.type)
```

MikeMcQuaid:
Probably want to add some Linux CI with this before I merge it. Yeh, Linux only formulae will need an if still. Thanks for the suggestion @colindean and testing @issyl0!

lib/bundle/extend/os/linux/skipper.rb
issyl0: Sadly not, this fails too with the same `undefined method` error.

colindean:
:tada: cool!

PR #535, opened 2019-07-30
Fix brew bundle check output for services

### LINKED ISSUE #534: `brew bundle check` output does not show services correctly
Given a brewfile containing a service definition, the output of `brew bundle check` shows an unhelpful struct inspection, rather than the name. For example:

```
→ Service #<Bundle::Dsl::Entry:0x00007fcf5285b7a8> needs to be started.
```

PR incoming...

joehorsnell:
The current `brew bundle check` output for a brewfile that specifies a service
is the somewhat unhelpful default `to_s` of the `Bundle::Dsl::Entry` struct.

```
→ Service #<Bundle::Dsl::Entry:0x00007fcf5285b7a8> needs to be started.
```

This PR adds a trivial `to_s` implementation to improve the output:

```
→ Service def needs to be started.
```

Fixes #534.

### DIFF
--- a/spec/bundle/commands/check_command_spec.rb
+++ b/spec/bundle/commands/check_command_spec.rb
@@ -3,13 +3,17 @@
 require "spec_helper"
 
 describe Bundle::Commands::Check do
-  def do_check
-    Bundle::Commands::Check.run
+  RSpec::Matchers.define_negated_matcher :not_raise_error, :raise_error
+
+  def do_check(expected_error = nil, expected_output = //)
+    args = expected_error ? [:raise_error, expected_error] : %i[not_raise_error]
+    expect {
+      Bundle::Commands::Check.run
+    }.to output(expected_output).to_stdout.and send(*args)
   end
 
   before do
     Bundle::Checker.reset!
-    allow_any_instance_of(IO).to receive(:puts)
   end
 
   context "when dependencies are satisfied" do
@@ -21,7 +25,7 @@ describe Bundle::Commands::Check do
       allow(Bundle::Checker).to receive(:formulae_to_install).and_return(nothing)
       allow(Bundle::Checker).to receive(:apps_to_install).and_return(nothing)
       allow(Bundle::Checker).to receive(:taps_to_tap).and_return(nothing)
-      expect { do_check }.not_to raise_error
+      do_check
     end
   end
 
@@ -30,7 +34,7 @@ describe Bundle::Commands::Check do
       allow(ARGV).to receive(:value).and_return(nil)
       allow_any_instance_of(Pathname).to receive(:read).and_return("")
       allow_any_instance_of(Bundle::Dsl).to receive(:entries).and_return([])
-      expect { do_check }.not_to raise_error
+      do_check
     end
   end
 
@@ -41,7 +45,7 @@ describe Bundle::Commands::Check do
       allow(Bundle::BrewInstaller).to receive(:upgradable_formulae).and_return([])
       allow(ARGV).to receive(:include?).and_return(true)
       allow_any_instance_of(Pathname).to receive(:read).and_return("cask 'abc'")
-      expect { do_check }.to raise_error(SystemExit)
+      do_check(SystemExit)
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/bundle/commands/check_command_spec.rb
```diff
@@ -3,13 +3,17 @@
 require "spec_helper"
 
 describe Bundle::Commands::Check do
-  def do_check
-    Bundle::Commands::Check.run
+  RSpec::Matchers.define_negated_matcher :not_raise_error, :raise_error
+
+  def do_check(expected_error = nil, expected_output = //)
```
jacobbednarz: I don't really like this as we're rolling our own version of RSpec's [`raise_error`](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/raise-error-matcher) matcher which seems unnecessary given the tests were previously using explicit assertions that were more concise than this one. Can you provide a little bit of insight into why you've wrapped this up into it's own method versus leaving the tests as they were?

joehorsnell: Hi @jacobbednarz, thanks for your review.

This is not really rolling our own version of RSpec's `raise_error` - this is using RSpec's built-in support for [defining a negated version](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/define-negated-matcher) of the existing `raise_error` matcher. This is required, to the best of my knowledge, when defining a [compound expectation](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/compound-expectations) where one of the clauses is negated (ie some expectation _and_ to _not_ raise an error) as there is [currently no way of doing that](https://github.com/rspec/rspec-expectations/issues/664).

Back to the why - I wanted to verify the output of the command, but currently [this before each](https://github.com/Homebrew/homebrew-bundle/blob/fae70b8e6252f07e0fec170e88738b5ae175827f/spec/bundle/commands/check_command_spec.rb#L12) block stubs all calls to `puts`, presumably to keep the spec output clear.

Given that RSpec has [built-in support for matching output](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/output-matcher) I thought that was a better approach than stubbing `puts` directly, even if I didn't want to verify _what_ the output was, ie just to prevent it being printed to screen.

I could obviously get rid of the logic in `do_check` and inline the compound expectations back into each spec, but I would argue that it's more concise as it is (certainly fewer characters), at the expense of extra complexity in the `do_check` method.

Personally I would favour the current approach in terms of being more concise, but happy to switch to inlining it if you'd prefer, being simpler at the expense of more verbose.

Anyway, let me know your thoughts.

MikeMcQuaid: We don't need to check for an exception not being raised because the test will fail if the exception is raised.

spec/bundle/dsl_spec.rb
```diff
@@ -3,6 +3,18 @@
 require "spec_helper"
 
 describe Bundle::Dsl do
+  describe described_class::Entry do
+    subject(:entry) { described_class.new(:brew, entry_name) }
+
+    let(:entry_name) { :foo }
+
+    describe "#to_s" do
```
MikeMcQuaid: Unless the coverage check fails without this: please remove this test because it's effectively just duplicating the (simple) implementation and the bug fixed is covered by the integration test above. Thanks!

joehorsnell: Hi @MikeMcQuaid - I will obviously remove it if that's what you want - it's your project! 😄But are you sure?

I know this is a trivial example, but I would argue it's not duplicating the implementation - the implementation is admittedly very simple, but it's a regression spec for other code that depends on that specific behaviour. The other spec might still pass due to unrelated changes.

Also, there are currently no specs for the `Entry` class, so the majority of that diff is just the boilerplate of adding the spec. It reasonable that it might have additional behaviour at some point?

Let me know.

MikeMcQuaid: > But are you sure?

Yes, thanks.

MikeMcQuaid:
Nice work, thanks again @joehorsnell!

PR #546, opened 2019-08-26
Adjust documentation output formatting

EricFromCanada:
Shift whitespace to match other `brew help <cmd>` output and fix up grammar.


MikeMcQuaid:
Nice work, thanks again @EricFromCanada! A possible improvement would be to migrate `brew bundle` and `brew services` to use the CLIParser stuff.

PR #548, opened 2019-08-27
skipper: Specify the type of the formula being skipped on Linux

issyl0:
- The output was unclear in the case that you had a Brewfile with a cask
  and then a standard formula (eg Linux formula) of the same name which
  installs fine.

Before:

```
Using aws-vault
Skipping aws-vault (on Linux)
```

After:

```
Using aws-vault
Skipping cask aws-vault (on Linux)
```


MikeMcQuaid:
Makes sense, thanks @issyl0!

PR #552, opened 2019-09-11
Write Brewfile.lock.json files

MikeMcQuaid:
Starting with a caveat: we will never have a "proper" lockfile in Homebrew/homebrew-bundle that installs all your dependencies at exactly the versions you want because Homebrew doesn't work that way (and never will because e.g. pinning OpenSSL at the same version forever is a bad idea).

I was somewhat inspired by a conversation with `@jnewland` to give this a go. What I'm thinking at a high-level this lockfile does is say "on this given configuration Homebrew/homebrew-bundle installed these versions of packages successfully". If a run is successful we'll dump various data about what succeeded into a lockfile. Within teams people may vary on the OS versions and even OS (macOS vs. Linux) they use so these will be (eventually) used as separate namespaces for system information so the lockfile can store every high-level configuration in which things worked.

`brew bundle` will output a `Brewfile.lock.json` in the same directory as the `Brewfile` if all dependencies are installed successfully. This contains dependency and system status information which can be useful in debugging `brew bundle` failures and replicating a "last known good build" state.

You can opt-out of this behaviour by setting the `HOMEBREW_BUNDLE_NO_LOCK` environment variable or passing the `--no-lock option`.

You may wish to check this file into the same version control system as your `Brewfile` (or ensure your version control system ignores it if you'd prefer to rely on debugging information from a local machine).

For example, for the following `Brewfile`:
```ruby
brew "ack"
cask "fork"
mas "StopTheMadness", id: 1376402589
tap "homebrew/homebrew-bundle"
```

The `Brewfile.lock.json` would be:
```json
{
  "entries": {
    "brew": {
      "ack": {
        "version": "3.2.0",
        "bottle": false
      },
      "readline": {
        "version": "8.0.1",
        "bottle": {
          "cellar": ":any",
          "prefix": "/usr/local",
          "files": {
            "catalina": {
              "url": "https://homebrew.bintray.com/bottles/readline-8.0.1.catalina.bottle.tar.gz",
              "sha256": "ab3c966f4cae7d0f3ecc5688bb989820c3261f5ed547a08c84186ba7f53bdd9c"
            },
            "mojave": {
              "url": "https://homebrew.bintray.com/bottles/readline-8.0.1.mojave.bottle.tar.gz",
              "sha256": "3c754391e9d243835811d128771ca0f1a565024100fd2c2871534353d46aaf0e"
            },
            "high_sierra": {
              "url": "https://homebrew.bintray.com/bottles/readline-8.0.1.high_sierra.bottle.tar.gz",
              "sha256": "ae341a036139a92a47396aabc773ffcf40a17fc388aaadf0147f688c72ece987"
            },
            "sierra": {
              "url": "https://homebrew.bintray.com/bottles/readline-8.0.1.sierra.bottle.tar.gz",
              "sha256": "f234d1ff8148bf08b0ac31e661f2e96b5c6e64df26a45d2392056c9077f964af"
            }
          }
        }
      }
    },
    "cask": {
      "fork": {
        "version": "1.0.72.2"
      }
    },
    "mas": {
      "StopTheMadness": {
        "id": "1376402589",
        "version": "8.6"
      }
    },
    "tap": {
      "homebrew/bundle": {
        "revision": "045be0715c7cd41f0e914c9faf9dd42e7e1fae59"
      }
    }
  },
  "system": {
    "macos": {
      "catalina": {
        "HOMEBREW_VERSION": "2.1.16-39-g5595dea",
        "HOMEBREW_PREFIX": "/usr/local",
        "Homebrew/homebrew-core": "a426105d937988c2d7dbe6b4e6fc1589dfb29e10",
        "CLT": "1100.0.33.8",
        "Xcode": "11.2",
        "macOS": "10.15.1"
      }
    }
  }
}
```


ssmusoke:
Excellent idea

1. How about a timestamp reference in the file to know when the changes were made 
2. How are old copies of the file maintained - last good versions? And how is this configured or changed?

MikeMcQuaid:
> * How about a timestamp reference in the file to know when the changes were made

I'd rather store all state e.g. Xcode versions, homebrew-core revisions rather than the time.

> 2. How are old copies of the file maintained - last good versions? And how is this configured or changed?

The last versions installed on any OS and the last good state per-OS (so overwritten for the same OS but kept for a different OS).

maxfierke:
I'm a fan. One thing that would be useful in addition though: cask args used, like location of application directory, etc.

ssmusoke:
> I'd rather store all state e.g. Xcode versions, homebrew-core revisions rather than the time.

Actually I meant in addition to the state, but to know when the brew update happened to help when troubleshooting 

> The last versions installed on any OS and the last good state per-OS (so overwritten for the same OS but kept for a different OS).

Not to mis-use it, but reminds me of Windows Restore points, keep a number from which a revert can be done

tgamblin:
> I'd rather store all state

Even though core doesn't have options, it seems like you would need to allow people to specify options in addition to versions to capture all state from taps:

```json
  "entries": {
    "brew": {
      "ack": {
        "version": "3.1.1",
        "options": [
            "with-openh264",
            "with-openjpeg"
        ]
      }
    },
    "tap": {
      "varenc/homebrew-ffmpeg": {
        "revision": "25ea527b3c6f99c28b6481fcffed3eeb243d3e30"
      }
    }
```

We do this in [`spack.yaml`](https://spack.readthedocs.io/en/latest/environments.html#spack-yaml), though Spack already has a syntax for specifying and depending on versions/options/compilers/compiler flags/target architecture/etc, so it's a bit more straightforward to work into a file.

jacobbednarz:
I think this proposal itself is an excellent launchpad, thanks for raising it! Something I'd like some clarification on is the purpose and intent of this as you've called out pretty clearly that this functionality isn't something we'd ever want in a "proper" lockfile. I'd like to better understand the place this is coming from design wise before getting too caught up in the implementation itself.

sjackman:
I'm excited by this feature. It would be useful to me to record which formulae and versions were used to satisfy `Brewfile`.

sjackman:
On Linux It would be useful to record the information from these lines of `brew config:
```
OS: CentOS Linux release 7.4.1708 (Core)  (Core)
Host glibc: 2.17
/usr/bin/gcc: 4.8.5
glibc: 2.23
gcc: 5.5.0_4
```
It may not be necessary to record the host's `glibc` and `gcc` version if the brewed `glibc` and `gcc` are installed.

MikeMcQuaid:
> One thing that would be useful in addition though: cask args used, like location of application directory, etc.
> Even though core doesn't have options, it seems like you would need to allow people to specify options in addition to versions to capture all state from taps

Yeh, this already happens it just doesn't show empty/default options.


> I think this proposal itself is an excellent launchpad, thanks for raising it! Something I'd like some clarification on is the purpose and intent of this as you've called out pretty clearly that this functionality isn't something we'd ever want in a "proper" lockfile. I'd like to better understand the place this is coming from design wise before getting too caught up in the implementation itself.

I'm thinking that although this is unlikely to ever support e.g. checking out a specific version of Homebrew/homebrew-core for you that it could be something stored with the project to indicate the last known configuration in which the `brew bundle` run was successful. For example, if a `brew bundle` run starts failing because no-one has run it in a year this format of lock file would be able to figure out the differences as e.g. an Xcode version bump or `wget` being on a newer version. 

> It may not be necessary to record the host's `glibc` and `gcc` version if the brewed `glibc` and `gcc` are installed.

@sjackman If the brewed versions are installed are they always used over the non-brewed ones?

tgamblin:
> Yeh, this already happens it just doesn't show empty/default options.

If the defaults can change over time, it's good to be explicit in the lockfile about all existing options.  Technically you know what is default by looking at the particular brew or cask hash, but if you ever want to reproduce a lockfile using a *different* version of brew, it's easier if you don't to have to check out an old one to determine what the locked configuration should be.

I realize that is not in your list of current envisioned use cases, but even if it's not, it does make the provenance clearer to humans, as well.

MikeMcQuaid:
> If the defaults can change over time, it's good to be explicit in the lockfile about all existing options.

This is accounted for by the tap hashes being stored. The formulae will still change even if the options do not.

issyl0:
On Linux with the example Brewfile:

```shell
$ brew bundle --lock     
Using ack
Skipping cask fork (on Linux)
Skipping mas StopTheMadness (on Linux)
Using homebrew/bundle
Error: No such file or directory - mas
```

That happens because it tries to run `mas list`, and is fixed with:

```diff
diff --git a/lib/bundle/installer.rb b/lib/bundle/installer.rb
index e8c3f87..e4a16f8 100644
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -82,6 +82,7 @@ module Bundle
             options
             { version: @cask_list[entry.name] }
           when :mas
+            if OS.mac?
               @mas_list ||= begin
                 name_id_versions = {}
                 `mas list`.lines.each do |line|
@@ -100,6 +101,7 @@ module Bundle
               options.delete(:id)
               options
               @mas_list[entry.name]
+            end
           when :tap
             options = entry.options
             options.delete(:clone_target) if options[:clone_target].blank?
```

Then, with the right branch checked out of `Homebrew/brew`, it works fine and outputs like this:

```json
{
  "system": {
    "linux": {
      "Ubuntu 18.04.3 LTS (bionic)": {
        "HOMEBREW_VERSION": "2.1.11-52-g09b73aa",
        "HOMEBREW_PREFIX": "/home/linuxbrew/.linuxbrew",
        "Homebrew/linuxbrew-core": "91c73a8c01d062435b76cc4bcc73f37b88efdc12"
      }
    }
  },
  "entries": {
    "brew": {
      "ack": {
        "version": "3.0.2"
      }
    },
    "cask": {
      "fork": {
        "version": null
      }
    },
    "mas": {
      "StopTheMadness": null
    },
    "tap": {
      "homebrew/bundle": {
        "revision": "0c2005eebe2e690db209c419d275fdcc444dc3eb"
      }
    }
  }
}
```

Overall, I think this is a good idea, not least to help resolve version discrepancies between eg team members in a work context, and to have an idea of what state things are in across systems. Thanks, Mike! :star:

MikeMcQuaid:
Thanks @issyl0! The main questions I had for you specifically (and @sjackman I guess!) are:

- what do you think we should do with e.g. `cask` and `mas` in the lock file on Linux?
- is `Ubuntu 18.04.3 LTS (bionic)` a good differentiator between distros? If not, what would be a better one?
- what other information do you think is needed to identify configuration differences (that would matter in 99% of cases) on Linux? For example, on macOS we're storing the Xcode version but not the Ruby version or X11/XQuartz version.

MikeMcQuaid:
> The version of the default compiler (whether brewed or host), the version of `glibc` (ditto), and the version of the Linux kernel.

@sjackman How often do you find failures are down to one of these? I basically only want fairly high signal things at first. For example, having the wrong XQuartz version breaks things on macOS sometimes but very infrequently and is warned by `brew doctor`. Having the wrong Xcode version breaks things very often.

sjackman:
>> The version of the default compiler (whether brewed or host), the version of `glibc` (ditto), and the version of the Linux kernel.

> @sjackman How often do you find failures are down to one of these? I basically only want fairly high signal things at first. For example, having the wrong XQuartz version breaks things on macOS sometimes but very infrequently and is warned by `brew doctor`. Having the wrong Xcode version breaks things very often.

I'd say rarely, because if the versions of these things are too old, then Homebrew on Linux installs its brewed versions. If the versions of these things are too new, it could potentially cause trouble, but in practice, I don't think it's common. Of the three, I'd include the version of the default compiler, whether brewed or host's.

MikeMcQuaid:
> On Linux with the example Brewfile:

@issyl0 Mind checking out the latest version again here? Thanks!

issyl0:
This version falls over if the `Brewfile.lock.json` doesn't exist at the start (it tries to [`lockfile.read`](https://github.com/Homebrew/homebrew-bundle/pull/552/commits/2525ebe5eab2e892cd37644d015ccfa086bada9c#diff-360aecb85c046bd68b6ec1f918c54a0eR9).

```
$ brew bundle --lock
Using ack
Using homebrew/bundle
Error: No such file or directory @ rb_sysopen - /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-bundle/Brewfile.lock.json
```

To continue testing, I made `Brewfile.lock.json` containing `{}`. Then it output:

```
Using ack
Using homebrew/bundle
Homebrew Bundle complete! 2 Brewfile dependencies now installed.
```

and 

```
{
  "system": {
    "linux": {
      "Ubuntu 19.04 (disco)": {
        "HOMEBREW_VERSION": "2.1.11-60-g47769f9",
        "HOMEBREW_PREFIX": "/home/linuxbrew/.linuxbrew",
        "Homebrew/linuxbrew-core": "c590d90da1c412fddd047fbb9fbd946445c4faaf",
        "GCC": "5.5.0"
      }
    }
  },
  "entries": {
    "brew": {
      "ack": {
        "version": "3.1.1"
      }
    },
    "tap": {
      "homebrew/bundle": {
        "revision": "2525ebe5eab2e892cd37644d015ccfa086bada9c"
      }
    }
  }
}
```

There's been a regression in that the `brew bundle` output no longer shows "Skipping cask ... (on Linux)" to the user.

Also, when a fully-populated Brewfile does exist, I found that running `brew bundle --lock` _again_ duplicated all the existing information. Is that intentional? I'd expect nothing to be added if all the versions were exactly the same?

MikeMcQuaid:
@issyl0 Think I've addressed all your comments. Mind trying one last time? Then I'll write tests properly and get this ready to 🚢.

issyl0:
Yep, I did all the same things and it seems to work fine now. :raised_hands:

```
$ brew bundle --lock
Using ack
Skipping cask fork (on Linux)
Skipping mas StopTheMadness (on Linux)
Using homebrew/bundle
```

and the resulting `Brewfile.lock.json`:

```
{
  "entries": {
    "brew": {
      "ack": {
        "version": "3.1.1"
      }
    },
    "tap": {
      "homebrew/bundle": {
        "revision": "87e369a7a46064fab00c66991a85fd8007e36e7a"
      }
    }
  },
  "system": {
    "linux": {
      "Ubuntu 18.04.3 LTS (bionic)": {
        "HOMEBREW_VERSION": "2.1.11-96-g1e725f4",
        "HOMEBREW_PREFIX": "/home/linuxbrew/.linuxbrew",
        "Homebrew/linuxbrew-core": "102f2c1f8abdf6f3ea4ffe0d65f854e7c8c83d22",
        "GCC": "5.5.0"
      }
    }
  }
}
```

lib/bundle/locker.rb
```diff
@@ -0,0 +1,100 @@
...
+
+module Bundle
+  module Locker
+    module_function
+
+    def lock(entries)
+      lockfile = Brewfile.path.dirname/"Brewfile.lock.json"
+
+      lock = if lockfile.exist?
+        JSON.parse(lockfile.read)
```
issyl0: How about a single-line `lock = JSON.parse(lockfile.read) if lockfile.exist?`?

MikeMcQuaid: I think that made more sense before refactoring but: yes, why not.

MikeMcQuaid:
Talked to @issyl0 about this in the Homebrew Slack. Some current thoughts on moving this forward:

- This file should maybe be renamed to `.Brewfile.lock.json` to avoid it being shown by default
- On a failed `brew bundle install` and an existing `.Brewfile.lock.json` being present: output the things in the configuration that have changed
- The file should be generated an existing one is present, if `--lock` is passed or `HOMEBREW_BUNDLE_LOCK` is set.

colindean:
I've been following the conversation as emails came in during the last few days but just now got around to reviewing everything all together.

At first, I didn't really understand the value of this, if anything beyond evidence that "When this file was generated, the bundle install was successful. It installed these versions defined by the formulae at these tap revisions in this environment." It had no other meaning or use beyond proof that "it worked on my machine". It's an installation ticket.

Outputting _something_ on a failure – that configuration changed – seems to be the sole use of this and it's probably a good one. I'm not sure how widely this would be used, though. It seems like the action for the user based on this would be "checkout this revision on this tap and try `bundle install` again" which I'm not sure is an action we want to support, since Homebrew always wants to be at the head of master of each tap in order to use the latest formulae.

I could see this lockfile being used to undo an installation more surgically: uninstall these formulae if they are at this version or newer and use the formulae defined in the tap at this revision to do it. I'm not sure how useful that would be, either. Upgrades would cause drift from the lockfile.

I remain concerned that if it's called a lockfile, users may _implicitly expect_ for it to function like lockfiles in other tools (Gemfile.lock, Cargo.lock, etc.) where the tool, here, brew bundle, is capable of installing the exact dependency versions stated in the lockfile. It seems clear from the second bullet point in the original post that this won't happen. I would thus expect this mismatched expectation to be a cause of friction: user frustration that it doesn't work as expected and developer frustration from pointing out to users that "It doesn't work that way".

Maybe it's worth enabling output of the file and seeing what uses surface from its existence.

lib/bundle/locker.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Locker
+    module_function
+
+    def lock(entries)
```
colindean: I wouldn't mind seeing this broken into several methods.

jhabdas:
> I remain concerned that if it's called a lockfile, users may _implicitly_ expect for it to function like lockfiles in other tools (Gemfile.lock, Cargo.lock, etc.) where the tool, here, brew bundle, is capable of installing the exact dependency versions stated in the lockfile.

The full-stop in front of the filename should mitigate that concern so long as the feature is left undocumented for now and assuming the file is never auto-generated.

My ideal command would be `< ~/.Brewfile.lock jq` to redirect the json blob to `jq`.

brandon-leapyear:
> I remain concerned that if it's called a lockfile, users may implicitly expect for it to function like lockfiles in other tools (Gemfile.lock, Cargo.lock, etc.) where the tool, here, brew bundle, is capable of installing the exact dependency versions stated in the lockfile. It seems clear from the second bullet point in the original post that this won't happen. I would thus expect this mismatched expectation to be a cause of friction: user frustration that it doesn't work as expected and developer frustration from pointing out to users that "It doesn't work that way".

Very much this. Right now, this file is breaking CI. For us, this PR is a breaking change, because we verify that no untracked files have been added when running build, and a new `Brewfile.lock.json` fails this check.

Furthermore, I just checked in a `Brewfile.lock.json`, but running `brew bundle` changes this file, which again files the check that no files have been modified in CI.

I'm going to add `--no-lock` to our `brew` invocation, but I am disappointed that this change was auto-opt-in.

Also confused about this section:

> You may wish to check this file into the same version control system as your `Brewfile`

If the lock file isn't meant to pin versions, then why would anyone want to check it into VCS? The point of checking things like `yarn.lock` to VCS is to get reproducible builds.

MikeMcQuaid:
> Right now, this file is breaking CI. For us, this PR is a breaking change, because we verify that no untracked files have been added when running build, and a new `Brewfile.lock.json` fails this check.

That's a bad assumption in your CI system. Many other tools will write different files between versions. Your organisation makes use of software created by volunteers in your spare time: you're welcome.

> Furthermore, I just checked in a `Brewfile.lock.json`, but running `brew bundle` changes this file

This is intentional behaviour.

> I'm going to add `--no-lock` to our `brew` invocation, but I am disappointed that this change was auto-opt-in.

Setting `HOMEBREW_BUNDLE_NO_LOCK=1` in your environment may be preferable.

This is how all of Homebrew works: it is a rolling release package manager.

> If the lock file isn't meant to pin versions, then why would anyone want to check it into VCS?

To identify the last known-good configuration able to install the `Brewfile`.

> The point of checking things like `yarn.lock` to VCS is to get reproducible builds.

This is something that could be more supported in future by e.g. installing from the bottles.

ydnar:
First, thanks for this PR. We’re excited about a lockfile so we can move closer to reproducible environments in development and CI.

> That's a bad assumption in your CI system. Many other tools will write different files between versions.

It’s trivial to set up a polyglot monorepo with a dozen or more dependency systems without a single changed file in the working tree. In one project, each of our CI job check for a modified tree, and fail the build if any committed file has changed.

I guess if I were to request any change, it’d be to omit the `system` key from the lockfile, and/or emit that on stdout when running `brew bundle`. CocoaPods does something similar, where it writes the version of CocoaPods used to generate the lockfile, when nothing else has changed.

MikeMcQuaid:
> In one project, each of our CI job check for a modified tree, and fail the build if any committed file has changed.

Cool, that's fine but it's also your decision and not really out problem.

> I guess if I were to request any change, it’d be to omit the `system` key from the lockfile, and/or emit that on stdout when running `brew bundle`. CocoaPods does something similar, where it writes the version of CocoaPods used to generate the lockfile, when nothing else has changed.

👍 Would accept a flag and environment variable PR that doesn't add `system`.

PR #569, opened 2019-10-15
Use `ruby-version` instead of `version`

jacobbednarz:
While merging an automated bundle update, I seen an annoated beta message below the files section that pointed out we should be using `ruby-version` not `version` when using the `actions/setup-ruby@v1` container.

See https://github.com/Homebrew/homebrew-bundle/pull/568/files for the message.


MikeMcQuaid:
Thanks @jacobbednarz, good catch!

Issue #480, opened 2019-04-26
`brew bundle cleanup` would incorrectly remove formulae required by casks


xu-cheng:
Steps to reproduce the bug.

```
brew cask install qlcolorcode
brew bundle dump
brew bundle cleanup
```

The last command would print:
```
Would uninstall formulae:
highlight
```

However `highlight` formula is in fact required by the `qlcolorcode` cask.

MikeMcQuaid:
This is a missing feature rather than a bug. We don't handle formula dependencies for casks but this would be a desirable feature to add.

alecclarke:
👋 I'd like to take this one on. Is it still up for grabs?

MikeMcQuaid:
Yes.

PR #571, opened 2019-10-26
Keep cask formula dependencies in "bundle cleanup"

alecclarke:
Fixes https://github.com/Homebrew/homebrew-bundle/issues/480

When running `brew bundle cleanup`, cask formula dependencies weren't being added to the list of kept formulae.

This change checks that for each kept cask, we add any formula dependencies to the list of kept formulae.

### Steps for testing:
1. `brew cask install qlcolorcode`
2. `brew bundle dump`
3. `brew bundle cleanup`

#### Before applying this PR's changes
When running `brew bundle cleanup` the listed formulae to be uninstalled **will** include `highlight` (a dependency of the `qlcolorcode` cask) and `lua` (a dependency of the `highlight` formula)

#### After applying this PR's changes
When running `brew bundle cleanup` the listed formulae to be uninstalled **won't** include `highlight` (a dependency of the `qlcolorcode` cask) or `lua` (a dependency of the `highlight` formula)

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -29,5 +29,14 @@ module Bundle
         (full_name_casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
       ]
     end
+
+    def formula_dependencies(cask_list)
+      cask_info_response = `brew cask info #{cask_list.join(" ")} --json=v1`
+      cask_info = JSON.parse(cask_info_response)
+
+      cask_info.flat_map do |cask|
+        cask.dig("depends_on", "formula")
+      end.compact.uniq
+    end
   end
 end
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -10,6 +10,7 @@ module Bundle
 
       def reset!
         @dsl = nil
+        @kept_casks = nil
         Bundle::CaskDumper.reset!
         Bundle::BrewDumper.reset!
         Bundle::TapDumper.reset!
@@ -68,8 +69,6 @@ module Bundle
       end
 
       def casks_to_uninstall
-        @dsl ||= Bundle::Dsl.new(Brewfile.read)
-        kept_casks = @dsl.entries.select { |e| e.type == :cask }.map(&:name)
         current_casks = Bundle::CaskDumper.casks(full_names_only: true)
         current_casks - kept_casks
       end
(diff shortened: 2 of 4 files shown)


lib/bundle/cask_dumper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "cask/cask_loader"
```
MikeMcQuaid: Would like to avoid including cask code directly here if it can be avoided. Can this not be obtained from the JSON e.g. `brew cask info mongodb --json=v1`?

MikeMcQuaid: Would like to avoid including cask code directly here if it can be avoided. Can this not be obtained from the JSON e.g. `brew cask info mongodb --json=v1`?

lib/bundle/commands/cleanup.rb
```diff
@@ -68,15 +69,15 @@ def run
...
-        kept_casks = @dsl.entries.select { |e| e.type == :cask }.map(&:name)
         current_casks = Bundle::CaskDumper.casks(full_names_only: true)
         current_casks - kept_casks
       end
 
       def formulae_to_uninstall
         @dsl ||= Bundle::Dsl.new(Brewfile.read)
         kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
+        kept_cask_formula_dependencies = Bundle::CaskDumper.formula_dependencies(kept_casks)
+        kept_formulae.concat(kept_cask_formula_dependencies)
```
MikeMcQuaid:
```suggestion
        kept_formulae.concat += kept_cask_formula_dependencies
```

alecclarke:
@MikeMcQuaid Example test steps have been added to the PR's description!

lib/bundle/cask_dumper.rb
```diff
@@ -29,5 +32,14 @@ def dump(casks_required_by_formulae)
         (full_name_casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
       ]
     end
+
+    def formula_dependencies(cask_list)
+      cask_info_response, _status = Open3.capture2("brew", "cask", "info", *cask_list, "--json=v1")
```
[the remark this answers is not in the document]
alecclarke: I thought about using `backticks` here, but even though this isn't accepting direct user input I'm always a bit cautious about using string interpolation when running shell commands. What do you think? Am I being overcautious?

MikeMcQuaid:
Fantastic PR @alecclarke, great work! I hope to see more PRs for you, you rock! We couldn't run the project without people like you contributing!

alecclarke:
Thanks for the feedback @MikeMcQuaid ! I will keep the PRs coming!

PR #577, opened 2019-11-01
Gemfile.lock: downgrade "bundled with".

MikeMcQuaid:
This will allow use with the version bundled with Ruby 2.6.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,4 +100,4 @@ DEPENDENCIES
   simplecov-cobertura
 
 BUNDLED WITH
-   2.0.1
+   1.17.2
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -100,4 +100,4 @@ DEPENDENCIES
   simplecov-cobertura
 
 BUNDLED WITH
-   2.0.1
```
jacobbednarz: Shame the Ruby version doesn't include [bundler autoswitch](https://bundler.io/guides/bundler_2_upgrade.html#version-autoswitch) to help automatically switching between desired versions.

MikeMcQuaid: Yeh :(

PR #597, opened 2019-11-29
Update docs to mention HOMEBREW_BUNDLE_FILE

joehorsnell:
Updating the docs to mention the `HOMEBREW_BUNDLE_FILE` env var.


MikeMcQuaid:
Good call, thanks again @joehorsnell!

PR #625, opened 2020-01-15
Fix brew bundle check -v

sjackman:
The option `brew bundle check -v` had no effect and should behave identically to `brew bundle check --verbose`.

### DIFF
--- a/lib/bundle/commands/check.rb
+++ b/lib/bundle/commands/check.rb
@@ -9,11 +9,11 @@ module Bundle
       FAILURE_MESSAGE = "brew bundle can't satisfy your Brewfile's dependencies."
 
       def output_errors?
-        ARGV.include?("--verbose")
+        ARGV.verbose?
       end
 
       def exit_on_first_error?
-        !ARGV.include?("--verbose")
+        !ARGV.verbose?
       end
 
       def run


lib/bundle/commands/check.rb
```diff
@@ -9,11 +9,11 @@ module Check
       FAILURE_MESSAGE = "brew bundle can't satisfy your Brewfile's dependencies."
 
       def output_errors?
-        ARGV.include?("--verbose")
+        ARGV.verbose?
```
MikeMcQuaid: We're currently removing these: https://github.com/Homebrew/brew/issues/5730.

I think at this point any extension should be to use `cli/parser` like Homebrew `cmd` blocks.

sjackman: Despite present plans, I suggest merging this PR regardless, as it fixes the current bug, and the proposed code is more correct than the previous code.

MikeMcQuaid: The tests will need fixed for this PR to be merged. Additionally, when they are fixed, the tests will need amended again when the `ARGV.verbose?` method is removed (somewhat imminently). Given that I'm probably 👎 but won't block you merging this if you disagree.

sjackman: I'm not sufficiently fussed to have a strong opinion. I'll wait for the imminent Homebrew/brew change. Closing.

PR #641, opened 2020-02-10
Add trailing newline to lock file

daniel-mohemian:
Adds a trailing newline to the lock file to make its last line conform to the [POSIX standard for lines](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206):

> 3.206 Line
>
> A sequence of zero or more non- \<newline\> characters plus a terminating \<newline\> character.

I'm not that fluent in rspec so I wasn't sure how to add a test that mocks `Pathname` to actually  check what is written to the lock file in the locker spec. Let me know if I should have a deeper look into that!

### DIFF
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -72,7 +72,7 @@ module Bundle
       json = JSON.pretty_generate(lock)
       begin
         lockfile.unlink if lockfile.exist?
-        lockfile.write(json)
+        lockfile.write(json.to_s + "\n")
       rescue Errno::EPERM, Errno::EACCES, Errno::ENOTEMPTY => e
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense to me! 
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @daniel-mohemian

daniel-mohemian:
Thanks @MikeMcQuaid!

Out of curiosity: How do releases work for external commands like `bundler`? Will they be included in one of the next `brew` releases?

MikeMcQuaid:
There are no releases, you `brew update` always to the latest version/`master`.

PR #643, opened 2020-02-12
Use `Homebrew::CLI::Parser`

MikeMcQuaid:
This helps us with consistent help output, parsing invalid options and accessing options without using `ARGV` directly.

While we're here: cleanup some old checks that are no longer needed and simplify `cmd/brew-bundle.rb`.

---

Note CI will fail on here as I haven't fixed the tests (yet).


lib/bundle/brew_dumper.rb
```diff
@@ -28,11 +28,11 @@ def dump
       end
       requested_formula.map do |f|
         brewline = ""
-        brewline += "# #{f[:desc]}\n" if ARGV.include?("--describe") && f[:desc]
+        brewline += "# #{f[:desc]}\n" if Homebrew.args.describe? && f[:desc]
```
jacobbednarz: much nicer! 🍭

jacobbednarz:
looks great as a starting point 🥇

lib/bundle/brew_dumper.rb
colindean: That's awesome.

PR #656, opened 2020-03-08
Add support for Whalebrew to bundle

### LINKED ISSUE #655: Support for whalebrew
I've been using Docker (alot) lately using an approach similar to [Jess Frazelle's `dockerfunc` setup](https://github.com/jessfraz/dotfiles/blob/master/.dockerfunc). More recently, I've started using [Whalebrew](https://github.com/whalebrew/whalebrew) which elegantly shims in Docker images inplace of a full binary. For example, take a look at the contents of [`/usr/local/bin/wget`](https://github.com/whalebrew/whalebrew-packages/blob/master/wget/Dockerfile).

This is working pretty-ok at the moment but I have hit a couple of times where I've used a Homebrew package instead of the intended Whalebrew wrapped-package due to to managing the packages via a Brewfile. This got me thinking, would it make sense to add Whalebrew as an addon for Homebrew Bundle? I don't know how used Whalebrew is at the moment so if this is just me, it's probably not worth it but if there is interest from others, it may be worth enabling something.

If this was to go ahead, the Brewfile would look a little something like this:

```rb
brew "wget"
brew "whalebrew"
whalebrew "httpie" 
``` 

Or perhaps more like `mas` where you don't need to explicitly mention the source brew package.

```rb
brew "wget"
whalebrew "httpie"
```

Thoughts? I'm happy to POC something together if there is any interest outside of myself.

jacobbednarz:
This introduces support for Whalebrew, a slightly different approach
to managing packages within Homebrew into homebrew-bundle. The
reasoning behind this is outlined in #655 but in short, if you manage
your environment with `brew` packages, it becomes increasingly
difficult to use this tool alongside it as you need to manage the
Brewfile _and_ a custom wrapper for installing Whalebrew packages
(can be done, just needs handrolled Ruby). By supporting this in
`homebrew-bundle` I'm hoping it allows these two installation methods
to co-exist and not step on each other.

An upside of this is that it's initial implementation has a small footprint
and should we choose not to pursue this, it's easily ripped out without
a mass clean up effort. Thanks to some of the decisions made in the 
tool upstream, the implementation for this is quite simple as you don't 
really need to concern yourself with hard things (like upgrades) as 
Docker can manage those under the covers without us doing extra 
legwork.

### Todo

- [x] Get CI passing on Linux + Mac
- [x] Add support to lock files 
- [x] README update
- [x] CLI output update

Closes #655


lib/bundle/whalebrew_dumper.rb
```diff
@@ -0,0 +1,28 @@
...
+    def reset!
+      @images = nil
+    end
+
+    def images
+      return [] unless Bundle.whalebrew_installed?
+
+      @images ||= `whalebrew list 2>/dev/null`.split("\n")
+      @images.reject { |image| image.include? "COMMAND   IMAGE" }
+             .map { |image| image.sub(/\w*\s+/, "") }
```
jacobbednarz: This is the only part of this PR that I'm not entirely happy with but 🤷‍♂ . The issue here is that the output is this:

```
$ whalebrew list
COMMAND   IMAGE
dig       whalebrew/dig
wget      whalebrew/wget
```

The first line ditches the headers (I'm going to PR upstream to have a flag to remove so we may be able to remove it later) and then the second replaces everything at the beginning up until the first chunk of whitespace to get rid of the column formatting. I looked at splitting on `\t` however it didn't seem consistent despite that [being the delimiter upstream](https://github.com/whalebrew/whalebrew/blob/master/cmd/list.go#L37). If there are options here to make this saner, I'm all ears!

jacobbednarz:
I'll check out those CI failures as they are on Linux but I completely forgot that's a supported platform now too 🤦‍♂

MikeMcQuaid:
Neat! This implementation is simple enough I'm 👍 on this being included once the tests are passing. Would also probably warrant a README addition and some `bundle.rb` help text changes . Great work so far, @jacobbednarz!

jacobbednarz:
Thanks @MikeMcQuaid 🙇 

I've updated the to-do list left for this PR however I'm interested to see if you can run these tests locally as I'm unable to replicate the issue popping up in CI.

MikeMcQuaid:
> I've updated the to-do list left for this PR however I'm interested to see if you can run these tests locally as I'm unable to replicate the issue popping up in CI.

Smells like mocking that isn't working as expected. Try to (temporarily) install/uninstall `whalebrew` on your machine.

jacobbednarz:
I did end up removing `whalebrew` but did miss that I had the executable in my $GOPATH which was catching these tests. Reproducing failure locally now, 👍

jacobbednarz:
Alright, this is now functioning with all the required bits. Will look addressing the code coverage next.

lib/bundle/locker.rb
```diff
@@ -128,6 +130,15 @@ def mas_list
       end
     end
 
+    def whalebrew_list
+      @whalebrew_list ||= begin
+        Bundle::WhalebrewDumper.images.each_with_object({}) do |image, name_versions|
+          version = `docker image inspect #{image} --format '{{ index .RepoDigests 0 }}' | awk -F':' '{print $2}'`.chomp
```
MikeMcQuaid: Can this line be shortened? Additionally, would be nice to parse this with Ruby rather than with `awk`.

cmd/bundle.rb
```diff
@@ -57,6 +57,8 @@ def bundle_args
              description: "`list` tap dependencies."
       switch "--mas",
              description: "`list` Mac App Store dependencies."
+      switch "--whalebrew",
```
MikeMcQuaid: Also needs updates on https://github.com/Homebrew/homebrew-bundle/blob/05f7b1ac5c4a74b49ed46ef8d6ea2a51d8c198cb/cmd/bundle.rb#L12 and https://github.com/Homebrew/homebrew-bundle/blob/05f7b1ac5c4a74b49ed46ef8d6ea2a51d8c198cb/cmd/bundle.rb#L20

lib/bundle/locker.rb
colindean:
```suggestion
          version = `docker image inspect #{image} --format '{{ index .RepoDigests 0 }}'`.split(':')[1].chomp
```

lib/bundle/whalebrew_dumper.rb
colindean: It'd be awesome if `whalebrew list` had a `--json[=v1]` option.

lib/bundle/locker.rb
MikeMcQuaid: `.second`  or `_, version, =` is nicer than `[1]` imo

cmd/bundle.rb
jacobbednarz: good pick up! addressed in 9abddde

lib/bundle/locker.rb
jacobbednarz: fair call; i've gone with multiple assignments instead of using an index as I find it easier to read. fixed in 253fb06

lib/bundle/whalebrew_dumper.rb
jacobbednarz: That would be very nice! I think the intention is the match the `docker` output as closely as possible so unsure how asking for a JSON output would go down 😛 My PR to remove the header row is at whalebrew/whalebrew#99

cmd/bundle.rb
```diff
@@ -9,15 +9,15 @@ def bundle_args
       usage_banner <<~EOS
         `bundle` <subcommand>
 
-        Bundler for non-Ruby dependencies from Homebrew, Homebrew Cask and the Mac App Store.
+        Bundler for non-Ruby dependencies from Homebrew, Homebrew Cask, Mac App Store and Whalebrew.
```
colindean:
```suggestion
        Bundler for non-Ruby dependencies from Homebrew, Homebrew Cask, Mac App Store, and Whalebrew.
```

MikeMcQuaid: Nope, no Oxford comma: https://docs.brew.sh/Prose-Style-Guidelines#typographical-conventions

jacobbednarz:
Narrowed this down to the following condition

```
Run options: include {:ids=>{"./spec/bundle/locker_spec.rb"=>["1:3:1", "1:4:1:2:1"]}}

Randomized with seed 3255

Bundle::Locker
  .lock
    writes Brewfile.lock.json
      on Linux
        returns true
  .whalebrew_list
    returns a hash of the name and layer checksum (FAILED - 1)

Failures:

  1) Bundle::Locker.whalebrew_list returns a hash of the name and layer checksum
     Failure/Error: expect(locker.whalebrew_list).to eq({"whalebrew/wget" => "abcd1234"})

       expected: {"whalebrew/wget"=>"abcd1234"}
            got: {}

       (compared using ==)

       Diff:
       @@ -1,2 +1 @@
       -"whalebrew/wget" => "abcd1234",

     # ./spec/bundle/locker_spec.rb:47:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:67:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:67:in `block (2 levels) in <top (required)>'

Finished in 0.02943 seconds (files took 0.15799 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/bundle/locker_spec.rb:44 # Bundle::Locker.whalebrew_list returns a hash of the name and layer checksum

Randomized with seed 3255
```

Looks like I'm missing something in the Linux tests or it's not resetting something for Whalebrew support as removing the Linux context block allows this to pass.

jacobbednarz:
After much back and forth with the code coverage, we're good for review! @MikeMcQuaid @colindean would you mind taking another pass over this and providing any feedback you have and we'll get this merged.

lib/bundle/locker.rb
```diff
@@ -132,6 +134,15 @@ def mas_list
       end
     end
 
+    def whalebrew_list
+      @whalebrew_list ||= begin
+        Bundle::WhalebrewDumper.images.each_with_object({}) do |image, name_versions|
+          _, version = `docker image inspect #{image} --format '{{ index .RepoDigests 0 }}'`.split(':')
```
MikeMcQuaid:
```suggestion
          , version = `docker image inspect #{image} --format '{{ index .RepoDigests 0 }}'`.split(':')
```

think that will work

lib/bundle/whalebrew_dumper.rb
```diff
@@ -0,0 +1,24 @@
...
+             .uniq
+    end
```
MikeMcQuaid:
```suggestion
      @images ||= begin
        `whalebrew list 2>/dev/null`.split("\n")
                                    .reject { |image| image.match? /^COMMAND / }
                                    .map { |image| image.sub(/\w*\s+/, "") }
                                    .uniq
      end
    end
```

spec/bundle/commands/install_command_spec.rb
```diff
@@ -16,15 +16,23 @@
...
 
   context "when a Brewfile is found" do
+    let(:brewfile_contents) do
+      <<~HEREDOC
+        tap 'phinze/cask'
+        brew 'mysql', conflicts_with: ['mysql56']
+        cask 'google-chrome'
+        mas '1Password', id: 443987910
+        whalebrew 'whalebrew/wget'
+      HEREDOC
```
MikeMcQuaid:
```suggestion
      <<~EOS
        tap 'phinze/cask'
        brew 'mysql', conflicts_with: ['mysql56']
        cask 'google-chrome'
        mas '1Password', id: 443987910
        whalebrew 'whalebrew/wget'
      EOS
```

but generally: 👏 on making this a heredoc

spec/bundle/locker_spec.rb
```diff
@@ -40,6 +40,17 @@
     end
   end
 
+  describe ".whalebrew_list" do
+    before do
+      allow(Bundle::WhalebrewDumper).to receive(:images).and_return(["whalebrew/wget"])
+      allow(locker).to receive(:`).with("docker image inspect whalebrew/wget --format '{{ index .RepoDigests 0 }}'").and_return("whalebrew/wget@sha256:abcd1234")
```
MikeMcQuaid:
```suggestion
      allow(locker).to receive(:`).with("docker image inspect whalebrew/wget --format '{{ index .RepoDigests 0 }}'")
                                  .and_return("whalebrew/wget@sha256:abcd1234")
```

or whatever doesn't wrap in the GitHub diff UI

spec/bundle/locker_spec.rb
```diff
@@ -68,6 +80,9 @@
           ]
         EOS
         allow(locker).to receive(:`).with("brew list --versions").and_return("mysql 8.0.18")
+        allow(locker).to receive(:`).with("whalebrew list").and_return("COMMAND   IMAGE\nwget      whalebrew/wget")
+        allow(locker).to receive(:`).with("docker image inspect whalebrew/wget --format '{{ index .RepoDigests 0 }}'").and_return("whalebrew/wget@sha256:abcd1234")
```
MikeMcQuaid:
```suggestion
        allow(locker).to receive(:`).with("docker image inspect whalebrew/wget --format '{{ index .RepoDigests 0 }}'")
                                    .and_return("whalebrew/wget@sha256:abcd1234")
```

or whatever doesn't wrap in the GitHub diff UI

spec/bundle/whalebrew_dumper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  subject(:dumper) { described_class }
+
+  describe '.images' do
+    before do
+      dumper.reset!
+      allow(Bundle).to receive(:whalebrew_installed?).and_return(true)
+    end
+
+    it 'removes the header' do
+      allow(dumper).to receive(:`).with("whalebrew list 2>/dev/null").and_return("COMMAND   IMAGE\nwget      whalebrew/wget")
```
MikeMcQuaid: Use a heredoc here and avoid wrapping in GitHub diff UI

spec/bundle/whalebrew_dumper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(dumper.images).to_not include("COMMAND")
+      expect(dumper.images).to_not include("IMAGE")
+    end
+
+    it 'dedupes items' do
+      allow(dumper).to receive(:`).with("whalebrew list 2>/dev/null").and_return("COMMAND   IMAGE\nwget      whalebrew/wget\nwget      whalebrew/wget")
```
MikeMcQuaid: Use a heredoc here and avoid wrapping in GitHub diff UI

spec/bundle/whalebrew_dumper_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      allow(dumper).to receive(:images).and_return(["whalebrew/wget", "whalebrew/dig"])
+    end
+
+    context "images are installed" do
+      it "returns correct listing" do
+        expect(dumper.images).to eq(["whalebrew/wget", "whalebrew/dig"])
+      end
+
+      it "dumps usable output for Brewfile" do
+        expect(dumper.dump).to eql(%Q{whalebrew "whalebrew/wget"\nwhalebrew "whalebrew/dig"})
```
MikeMcQuaid: Use a heredoc here

jacobbednarz:
Thanks @MikeMcQuaid! I've gone through and followed up on all the feedback with the exception of [removing the assignment to `_` in the string split](https://github.com/Homebrew/homebrew-bundle/pull/656/files#r394280366) as I was getting a syntax error. Happy to explore this further but I can't imagine too many issues popping up atm because of it.

MikeMcQuaid:
No worries! Will give @colindean a bit to have a look but I'm happy to ship.

colindean:
I will look shortly. If I haven't gotten to it by 10:00 p.m. Eastern, ship it.

jacobbednarz:
Oh no rush! I’m happy to wait a couple of days if you want to go through
it. There is no deadline here :)

spec/bundle/commands/install_command_spec.rb
```diff
@@ -16,15 +16,23 @@
   end
 
   context "when a Brewfile is found" do
+    let(:brewfile_contents) do
+      <<~EOS
```
colindean: 🎉

jacobbednarz:
Thank you both for checking it out!

PR #660, opened 2020-03-12
Skip tapping homebrew/cask when on Linux

issyl0:
- Casks are automatically skipped on Linux. When there's a Brewfile used on both macOS and Linux, there's usually Casks in it. This leads to `tap "homebrew/cask"` being present and causes the Cask tap to be installed on Linux too, even though Casks don't work on Linux and `cask` entries in Brewfiles are auto-skipped.
- A recent side effect of this behaviour is that because `brew update` now shows Cask data too, Linux users who cannot make use of Casks see a lot of noise, and may get their hopes up that Casks are available.
- Alternatives to this feature included:
    - Putting up with it.
    - Skipping outputting `brew update` info for Casks if on Linux.
    - Adding `tap "homebrew/cask" if OS.mac?` to my Brewfiles.

### DIFF
--- a/lib/bundle/extend/os/linux/skipper.rb
+++ b/lib/bundle/extend/os/linux/skipper.rb
@@ -3,8 +3,18 @@
 module Bundle
   module Skipper
     class << self
+      def macos_only_entry?(entry)
+        [:cask, :mas].include?(entry.type)
+      end
+
+      def macos_only_tap?(entry)
+        entry.type == :tap && entry.name == "homebrew/cask"
+      end
+
       def skip?(entry, silent: false)
-        return generic_skip?(entry) unless [:cask, :mas].include?(entry.type)
+        unless macos_only_entry?(entry) || macos_only_tap?(entry)
+          return generic_skip?(entry)
+        end
         return true if silent
 
         puts Formatter.warning "Skipping #{entry.type} #{entry.name} (on Linux)"


issyl0:
Some example output:

Brewfile:

```
tap "homebrew/cask"

brew "wget"
brew "ack"

cask "omnifocus"
```

Output:

```
╭─issyl0@rigel /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-bundle ‹skip-tapping-homebrew/cask-on-linux› 
╰─ $ brew bundle
Skipping tap homebrew/cask (on Linux)
Using wget
Using ack
Skipping cask omnifocus (on Linux)
```

lib/bundle/extend/os/linux/skipper.rb
```diff
@@ -4,7 +4,10 @@ module Bundle
   module Skipper
     class << self
       def skip?(entry, silent: false)
-        return generic_skip?(entry) unless [:cask, :mas].include?(entry.type)
+        unless [:cask, :mas].include?(entry.type) ||
```
jacobbednarz: I find this a little hard to grok and needed to read it a couple of times before understanding what we're aiming for here. Could we use some methods to wrap these to better convey what we're doing?

```rb
unless skippable_plugins?(entry) || cask_tap?(entry)
  return generic_skip?(entry)
end

# ...

def skippable_plugins?(entry)
  [:cask, :mas].include?(entry.type)
end

def cask_tap?(entry)
  entry.type == :tap && entry.name == "homebrew/cask"
end
```

I don't love the `skippable_plugins` method name here but just to demonstrate the idea. Not a deal breaker though if others find this readable enough.

issyl0: I considered this, but couldn't think of any decent method names. :grinning: I agree it makes more sense when split up - but I wonder if the `unless` is the thing that's really confusing us?

lib/bundle/extend/os/linux/skipper.rb
```diff
@@ -4,7 +4,10 @@ module Bundle
...
+               (entry.type == :tap && entry.name == "homebrew/cask")
+          return generic_skip?(entry)
+        end
```
colindean:
```suggestion
        mac_only_types = [:cask, :mas]
        mac_only_taps = ["homebrew/cask"]

        unless mac_only_types.include?(entry.type) ||
               (entry.type == :tap && mac_only_taps.include?(entry.name)
          return generic_skip?(entry)
        end
```

Could move those vars outside the `skip?` method scope, too.

issyl0:
@jacobbednarz @colindean Thanks for the suggestions! In the end I kept the `unless` but went with a mixture of your choices of naming.

lib/bundle/extend/os/linux/skipper.rb
```diff
@@ -3,8 +3,18 @@
...
+        [:cask, :mas].include?(entry.type)
+      end
+
+      def macos_only_tap?(entry)
+        entry.type == :tap && entry.name == "homebrew/cask"
+      end
+
       def skip?(entry, silent: false)
-        return generic_skip?(entry) unless [:cask, :mas].include?(entry.type)
+        unless macos_only_entry?(entry) || macos_only_tap?(entry)
```
MikeMcQuaid: `unless ... |||` is pretty hard to follow/parse. Could this be written as an `if` (or two)?

Issue #664, opened 2020-03-16
Flakey failure for MAS installer


jacobbednarz:
Steps to reproduce:

1. Checkout `master`. 491180c6f8bbd2b1496ab25aee4889b7c3de8caf at the time of raising the issue.
1. Run `bundle exec rspec spec --seed 60330`. See consistent failures. 
1. Run `bundle exec rspec spec --seed 11827`. See consistent test suite passes.

```
Failures:

  1) Bundle::MacAppStoreInstaller when mas is installed when mas is not signed in tries to sign in with mas
     Failure/Error: expect { do_install }.to raise_error(RuntimeError)
       expected RuntimeError but nothing was raised
     # ./spec/bundle/mac_app_store_installer_spec.rb:64:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:67:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:67:in `block (2 levels) in <top (required)>'

Finished in 11.88 seconds (files took 0.20816 seconds to load)
213 examples, 1 failure

Failed examples:

rspec ./spec/bundle/mac_app_store_installer_spec.rb:61 # Bundle::MacAppStoreInstaller when mas is installed when mas is not signed in tries to sign in with mas
```

Reproducible on Mac OS 10.14.6

colindean:
I'm able to reproduce this, too. I couldn't make heads or tails of it in the ~30 minutes I investigated it.

jacobbednarz:
Now that the whalebrew support has landed, I'll try to look into this one shortly.

jacobbednarz:
Minimum reproduction case is 

```
rspec ./spec/bundle/mac_app_store_dumper_spec.rb[1:3:1] ./spec/bundle/mac_app_store_installer_spec.rb[1:4:2:1] ./spec/bundle_utils_spec.rb[1:1:1,1:1:2,1:2:1,1:2:2] --seed 60330
```

PR #670, opened 2020-04-01
mac_app_store_installer: fix flaky test, remove dead code.

MikeMcQuaid:
@colindean @jacobbednarz I'm not convinced it's a good ROI filing more flaky failure issues. They take a while to debug, they aren't harming end-users and this is not a particularly flaky test suite in general.

Fixes https://github.com/Homebrew/homebrew-bundle/issues/664

### DIFF
--- a/lib/bundle/mac_app_store_installer.rb
+++ b/lib/bundle/mac_app_store_installer.rb
@@ -25,8 +25,7 @@ module Bundle
 
       unless Bundle.mas_signedin?
         puts "Not signed in to Mac App Store." if Homebrew.args.verbose?
-        Bundle.system "mas", "signin", "--dialog", "" if MacOS.version < :mojave
-        raise "Unable to install #{name} app. mas not signed in to Mac App Store." unless Bundle.mas_signedin?
+        raise "Unable to install #{name} app. mas not signed in to Mac App Store."
       end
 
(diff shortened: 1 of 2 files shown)


jacobbednarz:
I'm on the fence about ignoring flakey tests for exactly the reason you've mentioned. The test suite is currently pretty stable and when I'm looking at upgrades or introducing functionality, I lean _very_ heavily on the existing test suite for confidence in the change and I wouldn't want that to go away for others because we started ignoring the stability.

Yes, they are difficult to debug and they don't always harm the end user but it can make maintainer life difficult if they do start creeping in.

Either way, appreciate you grabbing the two I raised :star2: I had intentions of getting back to it but unfortunately it slipped down the priority list and didn't get looked at.

MikeMcQuaid:
> I'm on the fence about ignoring flakey tests for exactly the reason you've mentioned.

I'm not suggesting "ignoring" as much as "don't submit issues".

> Either way, appreciate you grabbing the two I raised 🌟 I had intentions of getting back to it but unfortunately it slipped down the priority list and didn't get looked at.

This is the thing I'm talking about mainly, I guess. I have zero objection to anyone who finds flaky tests submitting fixes for them. I do object a bit to issues being opened that end up falling on me to fix.

PR #672, opened 2020-04-06
Don't lowercase URL when those are specified in Brewfile

pinkahd:
If a Brewfile defines a specific version of a package and this
package originates form an URL, the name shouldn't be lowercase.

An example of such file could be:

```
brew "https://raw.githubusercontent.com/Homebrew/homebrew-core/50c928c870001577ce6b7e28edf43e05df699852/Formula/etcd.rb"
```

Previous to this case this was falling with this error:

```
curl: (22) The requested URL returned error: 404 Not Found
Error: Failure while executing; `/usr/bin/curl -q --globoff --show-error --user-agent Homebrew/2.2.11\ \(Macintosh\;\ Intel\ Mac\ OS\ X\ 10.15.4\)\ curl/7.64.1 --fail --progress-bar --retry 3 --location --remote-time --continue-at 0 --output /Users/alexandru/Library/Caches/Homebrew/Formula/etcd.rb https://raw.githubusercontent.com/homebrew/homebrew-core/50c928c870001577ce6b7e28edf43e05df699852/formula/etcd.rb` exited with 22. Here's the output:

curl: (22) The requested URL returned error: 404 Not Found
```

The reason of the failure was due to the name been transformed
to lowercase.

### DIFF
--- a/lib/bundle/dsl.rb
+++ b/lib/bundle/dsl.rb
@@ -89,7 +89,8 @@ module Bundle
     HOMEBREW_TAP_FORMULA_REGEX = %r{^([\w-]+)/([\w-]+)/([\w+-.@]+)$}.freeze
 
     def self.sanitize_brew_name(name)
-      name = name.downcase
+      name = name.downcase unless name.match(/\A#{URI.regexp(['http', 'https'])}\z/)
+
       if name =~ HOMEBREW_CORE_FORMULA_REGEX
         Regexp.last_match(1)
       elsif name =~ HOMEBREW_TAP_FORMULA_REGEX
--- a/spec/bundle/dsl_spec.rb
+++ b/spec/bundle/dsl_spec.rb
@@ -18,6 +18,7 @@ describe Bundle::Dsl do
         cask 'firefox', args: { appdir: '~/my-apps/Applications' }
         mas '1Password', id: 443987910
         whalebrew 'whalebrew/wget'
+        brew 'https://raw.githubusercontent.com/Homebrew/homebrew-core/50c928c870001577ce6b7e28edf43e05df699852/Formula/etcd.rb'
       EOS
     end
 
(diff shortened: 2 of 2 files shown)


pinkahd:
Is there a reason why URL's aren't supported?

My case here is that I need a specific version of `etcd` installed on the system. From my understanding the way to do that in `brew` is to install a past formula. (`brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/50c928c870001577ce6b7e28edf43e05df699852/Formula/etcd.rb`)unless I'm missing something.

MikeMcQuaid:
> Is there a reason why URL's aren't supported?

Run `brew install $URL` and you'll see the reason 😉.

In short: they will break and when they do they can't be fixed. We provide tooling (`brew extract`) to make your own specific versions you depend on that can be fixed when they break.

pinkahd:
Thanks for point that out.

I know this is not the best place to ask this but I've done exactly what you've said extracted the package and create a new tap. But now it tries to build the package... Since it can't find a pre-build package in:

```
==> Downloading https://homebrew.bintray.com/bottles-etcd/etcd-3.3.15.mojave.bottle.1.tar.gz
curl: (22) The requested URL returned error: 404 Not Found
```

Is this expected or did I messed something up?

When installing from an URL I can get the correct version `==> Downloading https://homebrew.bintray.com/bottles/etcd-3.3.15.mojave.bottle.1.tar.gz` Is there a way to make my tap install this? Building the package from scratch I don't think should be the desired result since I cannot assume re-building it will generate a replicable build.

---

If this is not the case I still think URL installs should be allow, if `brew install` command allows them I see no reason why Brewfile shouldn't. Or at least have a way to specify a version of a package in the Brewfile.

MikeMcQuaid:
> Is this expected or did I messed something up?

Somewhat expected. You'll want to remove the `bottle do` block. See https://github.com/Homebrew/brew/issues/6059

> If this is not the case I still think URL installs should be allow, if `brew install` command allows them I see no reason why Brewfile shouldn't.

There is a "warning" telling you not to use them. They are likely to be deprecated and removed in future.
> Or at least have a way to specify a version of a package in the Brewfile.

This is not how Homebrew works. It's a rolling-release package manager. Read https://docs.brew.sh/Versions

MikeMcQuaid:
Sorry, passing on this due to reasons above.

PR #687, opened 2020-05-02
Updates link to mas-cli

jeroenj:
It redirected as expected but it's cleaner to use the correct URL.


colindean:
Thanks for the update!

PR #745, opened 2020-07-27
Refactor usages of global `args`.

reitermarkus:
Depends on https://github.com/Homebrew/brew/pull/8097.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -79,25 +79,36 @@ module Homebrew
   end
 
   def bundle
-    bundle_args.parse
+    args = bundle_args.parse
 
     # Keep this after the .parse to keep --help fast.
     require_relative "../lib/bundle"
 
     begin
-      case subcommand = Homebrew.args.named.first.presence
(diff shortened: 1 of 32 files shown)


reitermarkus:
Actually it's the other way around: Homebrew/brew#8097 depends on this, since it removes the global options.

reitermarkus:
I think we need to tag a new `brew` release before merging this.

mistydemeo:
Even with this branch, I'm noticing that `brew bundle exec` is broken:

```
/usr/local/Homebrew/Library/Homebrew/extend/ENV.rb:13:in `activate_extensions!'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/commands/exec.rb:16:in `run'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/cmd/bundle.rb:99:in `bundle'
/usr/local/Homebrew/Library/Homebrew/brew.rb:111:in `<main>'
```

PR #758, opened 2020-08-05
Adds support for multi-line desc in brew formulae

### LINKED ISSUE #757: Improper line breaks processing for formulas with multiline descriptions when use `brew bundle --describe dump`.
Hello. I have a formula installed from tap with such description:
```
msoap/tools/shell2http: stable 1.13
Executing shell commands via simple http server

Settings through 2 command line arguments, path and shell command.
```
The formulæ `desc` itself contains line break characters:
```
desc "Executing shell commands via simple http server\n\nSettings through 2 command line arguments, path and shell command.\n\n"
```

And that's what we get in our Brewfile for this formulæ after `brew bundle --describe dump` run:
```
# Executing shell commands via simple http server

Settings through 2 command line arguments, path and shell command.


brew "msoap/tools/shell2http"
```
I know that `desc` must contain an one-line description, but maybe it's better to strip line breaks?

colindean:
Closes #757

### DIFF
--- a/lib/bundle/brew_dumper.rb
+++ b/lib/bundle/brew_dumper.rb
@@ -28,7 +28,7 @@ module Bundle
       end
       requested_formula.map do |f|
         brewline = ""
-        brewline += "# #{f[:desc]}\n" if describe && f[:desc]
+        brewline += desc_comment(f[:desc]) if describe && f[:desc]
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Makes sense, thanks @colindean.

Issue #793, opened 2020-09-18
Proposal: add a `bundle converge` command that does `bundle cleanup --force` + `bundle`


aramprice:
Not sure if this would be useful for other folks but I find myself often "converging" my machine to ensure that I have all, and only the things specified in my Brewfile are installed.

Curious if other folks might find this useful. If so happy to work on a PR. No attachment to `converge` as the command's name.

issyl0:
Another idea: run `bundle` automatically at the end of `brew cleanup --force`?

MikeMcQuaid:
I think a separate command would be a bit overkill for this. Either having the two separate commands we have now or a e.g. `brew bundle cleanup --install` (naming could change) makes sense to me.

aramprice:
I agree that adding another subcommand seems crowded maybe. The urge to keep the breadth of subcommands limited.

To me both `install` as part of `cleanup`, or `cleanup --install` feel a little cognitively surprising, though you all probably know the user base better.

If you all are interested can probably cobble something together for a PR over the next few weeks/month if you all are up for making the decision about where to hang this functionality. Here are the options so far, with a few additional ideas thrown in:

1. teach `cleanup --force` to run `install`
1. teach `cleanup` a new `--install` flag
1. teach `install` a new `--cleanup` (or `--converge`, or ??) flag
1. add `converge` command
1. ...?

jacobbednarz:
I feel this _could_ hang off `brew bundle cleanup` but looking a bit broader for use cases, I wonder if a single command would be more viable if it was instead targetted at drift detection and a flag available to that command was a way to resync the desired and actual. It seems like this could be separated and became used in automation much like `bundle check` is now.

MikeMcQuaid:
> * teach `cleanup --force` to run `install`
> * teach `cleanup` a new `--install` flag
> * teach `install` a new `--cleanup` ~(or `--converge`, or ??)~ flag

I'm game for any of these.

Hanging off `cleanup` makes a little more sense to me because `install` gets a lot more usage than `cleanup` and this functionality would be an extension for a greater proportion of existing users of `cleanup` compared to existing users of `install`.

aramprice:
My vote would be for `install --cleanup`. To me running `--cleanup` as part of installation feels more intuitive than running `--install` as part of cleaning up (removing software).

Maybe adding the flag to `install` would help introduce more folks to the idea of `cleanup`? Not sure if that's desirable or not.

[...]

PR #795, opened 2020-09-29
Add `--cleanup` flag to `bundle install`

aramprice:
When `--cleanup` is passed `bundle install` will remove any software not
present in the `Brewfile`.

The behavior of running:

```
brew bundle install --cleanup
```

should be equivalent to running:

```
brew bundle install && brew bundle cleanup --force
```

Fixes https://github.com/Homebrew/homebrew-bundle/issues/793

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -54,6 +54,8 @@ module Homebrew
       switch "-f", "--force",
              description: "`dump` overwrites an existing `Brewfile`. " \
                           "`cleanup` actually performs its cleanup operations."
+      switch "--cleanup",
+             description: "`install` performs cleanup operation, same as running `cleanup --force```"
       switch "--no-lock",
              description: "`install` won't output a `Brewfile.lock.json`."
       switch "--all",
@@ -94,6 +96,15 @@ module Homebrew
           no_upgrade: args.no_upgrade?,
           verbose:    args.verbose?,
         )
+
+        if args.cleanup?
+          Bundle::Commands::Cleanup.run(
+            global: args.global?,
+            file:   args.file,
+            force:  true,
+            zap:    args.zap?,
+          )
+        end
       when "dump"
         Bundle::Commands::Dump.run(
           global:     args.global?,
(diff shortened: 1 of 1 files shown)


aramprice:
I am a little uncertain of this implementation because there was no place to add tests. The file I modified [`cmd/bundle.rb`](https://github.com/Homebrew/homebrew-bundle/blob/master/cmd/bundle.rb) did not have a spec file.

I could have added the call to `Cleanup.run()` down inside of the [`Bundle::Commands::Install.run()`](https://github.com/Homebrew/homebrew-bundle/blob/12702c2b9111a447bb89cd887771d20cffab4e63/lib/bundle/commands/install.rb#L8) method but this felt a bit odd since I would have ended up passing a number of additional flags to `Bundle::Commands::Install.run()` for the sole purpose of then passing those flags down to [`Bundle::Commands::Cleanup.run()`](https://github.com/Homebrew/homebrew-bundle/blob/12702c2b9111a447bb89cd887771d20cffab4e63/lib/bundle/commands/cleanup.rb#L20).

Maybe there's an additional abstraction here looking to emerge.

cmd/bundle.rb
```diff
@@ -54,6 +54,8 @@ def bundle_args
       switch "-f", "--force",
              description: "`dump` overwrites an existing `Brewfile`. " \
                           "`cleanup` actually performs its cleanup operations."
+      switch "--cleanup",
+             description: "`install` performs cleanup operation, same as running `cleanup --force```"
```
MikeMcQuaid, line 58:
```suggestion
             description: "`install` performs cleanup operation, same as running `cleanup --force`"
```

cmd/bundle.rb
```diff
@@ -54,6 +54,8 @@ def bundle_args
       switch "-f", "--force",
              description: "`dump` overwrites an existing `Brewfile`. " \
                           "`cleanup` actually performs its cleanup operations."
+      switch "--cleanup",
+             description: "`install` performs cleanup operation, same as running `cleanup --force`"
```
MikeMcQuaid, line 58:
```suggestion
             description: "`install` performs cleanup operation, same as running `cleanup --force`."
```

MikeMcQuaid (approved):

MikeMcQuaid:
Great work here, thanks @aramprice!

colindean:
Thank you, @aramprice!

PR #810, opened 2020-11-05
Include descriptions from Casks in dump --describe output

### LINKED ISSUE #809: dump --describe only works on formulae
Looks like `brew bundle dump --describe` outputs the `desc`ription for formulae only, and not Casks. Even though they too now have a `desc` stanza…

danielbayley:
Fixes #809.

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -6,6 +6,7 @@ module Bundle
 
     def reset!
       @casks = nil
+      @cask_info = nil
     end
 
     def casks
@@ -16,25 +17,33 @@ module Bundle
             .uniq
     end
 
-    def dump(casks_required_by_formulae)
+    def dump(casks_required_by_formulae, describe: false)
       [
-        (casks & casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
-        (casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
+        (casks & casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
+        (casks - casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
       ]
     end
 
-    def formula_dependencies(cask_list)
-      return [] unless cask_list.present?
+    def cask_line(cask, describe)
+      desc = cask_info&.flat_map.find { |info| info["token"] == cask }&["desc"]
+      comment = "# #{desc}\n" if describe && !desc.blank?
+      "#{comment}cask \"#{cask}\""
+    end
 
-      cask_info_response = `brew cask info #{cask_list.join(" ")} --json=v1`
-      cask_info = JSON.parse(cask_info_response)
+    def cask_info
+      @cask_info ||= begin
+        output = `brew cask info #{casks.join(" ")} --json=v1`
+        JSON.parse(output)
+      rescue JSON::ParserError => e
+        opoo "Failed to parse `brew cask info --json`: #{e}"
+        []
+      end
+    end
 
+    def formula_dependencies
       cask_info.flat_map do |cask|
         cask.dig("depends_on", "formula")
       end.compact.uniq
-    rescue JSON::ParserError => e
-      opoo "Failed to parse `brew cask info --json`: #{e}"
-      []
     end
   end
 end
--- a/lib/bundle/dumper.rb
+++ b/lib/bundle/dumper.rb
@@ -17,7 +17,7 @@ module Bundle
       content = []
       content << TapDumper.dump
       casks_required_by_formulae = BrewDumper.cask_requirements
-      cask_before_formula, cask_after_formula = CaskDumper.dump(casks_required_by_formulae)
+      cask_before_formula, cask_after_formula = CaskDumper.dump(casks_required_by_formulae, describe: describe)
       content << cask_before_formula
       content << BrewDumper.dump(describe: describe, no_restart: no_restart)
       content << cask_after_formula
--- a/spec/bundle/cask_dumper_spec.rb
+++ b/spec/bundle/cask_dumper_spec.rb
@@ -81,6 +81,7 @@ describe Bundle::CaskDumper do
 
     context "when multiple casks have the same dependency" do
       before do
+        described_class.reset!
         allow(described_class)
           .to receive(:`)
           .with("brew cask info foo bar --json=v1")


lib/bundle/cask_dumper.rb
```diff
@@ -16,13 +16,19 @@ def casks
             .uniq
     end
 
-    def dump(casks_required_by_formulae)
+    def dump(casks_required_by_formulae, describe: false)
+      to_line = ->(cask) { "#{desc_comment(cask)}cask \"#{cask}\"" }
```
MikeMcQuaid, line 20: Could this be a dedicated method that has `desc_comment` inlined?

lib/bundle/cask_dumper.rb
```diff
@@ -16,13 +16,19 @@ def casks
...
-        (casks & casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
-        (casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
+        (casks & casks_required_by_formulae).map(&to_line).join("\n"),
+        (casks - casks_required_by_formulae).map(&to_line).join("\n"),
       ]
     end
 
+    def desc_comment(cask)
+      desc = Cask::CaskLoader.load(cask).desc
+      desc.nil? ? "" : "# #{desc}\n"
```
MikeMcQuaid, line 29:
```suggestion
      desc = Cask::CaskLoader.load(cask).desc
      return "" if desc.blank?
      
      "# #{desc}\n"
```

MikeMcQuaid (commented):
Nice work! Would be good to have some tests for this.

MikeMcQuaid:
Oops, CI failing in this case due to missing test coverage: https://github.com/Homebrew/homebrew-bundle/pull/810/checks?check_run_id=1358404591#step:9:87

lib/bundle/cask_dumper.rb
danielbayley: It get awkward because you can't pass arguments (`describe`, in this case) to `map`… What exactly did you have in mind?

MikeMcQuaid: Where/how does describe change behaviour? I don't see it being used here.

danielbayley: > Where/how does describe change behaviour? I don't see it being used here.

Yeah I realised this, and fixed it… just forgot to push that change.

lib/bundle/cask_dumper.rb
```diff
@@ -16,13 +16,24 @@ def casks
             .uniq
     end
 
-    def dump(casks_required_by_formulae)
+    def dump(casks_required_by_formulae, describe: false)
+      to_line = lambda { |cask|
+        comment = desc_comment(cask) if describe
+        "#{comment}cask \"#{cask}\""
+      }
```
MikeMcQuaid, line 23: I see the `describe` usage now. Should be fine to have `desc_comment` take a `describe` argument and just use `.map {|c| desc_comment(c, describe)`. I think the lambda usage as is harms rather than helps readability.

lib/bundle/cask_dumper.rb
```diff
@@ -16,13 +16,19 @@ def casks
             .uniq
     end
 
-    def dump(casks_required_by_formulae)
+    def dump(casks_required_by_formulae, describe: false)
       [
-        (casks & casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
-        (casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
+        (casks & casks_required_by_formulae).map { |cask| desc_comment(cask, describe) }.join("\n"),
+        (casks - casks_required_by_formulae).map { |cask| desc_comment(cask, describe) }.join("\n"),
```
MikeMcQuaid, line 22:
```suggestion
        (casks & casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
        (casks - casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
```

danielbayley:
> Would be good to have some tests for this.

@MikeMcQuaid I started writing them, but not sure how to get around the `NameError: uninitialized constant Bundle::CaskDumper::Cask`, due to relying on `Cask::CaskLoader.load(cask).desc`…

So far I just have:
~~~ rb
context "when `describe` is false" do
  subject(:dump) { described_class.dump(dumper.casks, describe: false) }

  before do
    described_class.reset!
  end

  it "does not output a comment with dependency description" do
    expect(dump).not_to include("#")
  end
end
~~~

lib/bundle/cask_dumper.rb
```diff
@@ -16,13 +16,19 @@ def casks
...
       ]
     end
 
+    def desc_comment(cask, describe)
```
MikeMcQuaid, line 26:
```suggestion
    def cask_line(cask, describe)
```

MikeMcQuaid (commented):
Looks good after my suggested changes! Just needs some tests added so CI will pass now.

danielbayley:
> Just needs some tests added so CI will pass now.

See above [comment](https://github.com/Homebrew/homebrew-bundle/pull/810#issuecomment-723001325)☝️

MikeMcQuaid:
Sorry, missed that!

> @MikeMcQuaid I started writing them, but not sure how to get around the `NameError: uninitialized constant Bundle::CaskDumper::Cask`, due to relying on `Cask::CaskLoader.load(cask).desc`…

A `require` in the class and/or the test should handle that.

danielbayley:
> A `require` in the class and/or the test should handle that.

@MikeMcQuaid That's what I thought, but trying to figure out exactly what/how to `require`, without playing _Whac-A-Mole_ with errors…
~~~ rb
# (HOMEBREW_LIBRARY const seems inaccessible)
$LOAD_PATH.unshift "#{HOMEBREW_PREFIX}/Homebrew/Library/Homebrew"
require "cask"
# NameError: uninitialized constant Cask::Artifact::AbstractArtifact::Predicable
~~~
I'm running `bundle exec rspec spec/bundle/cask_dumper_spec.rb` to test this FYI.

MikeMcQuaid:
@danielbayley https://github.com/Homebrew/homebrew-bundle/blob/9aed384937bdf545b5c4e3e1fcc0a2bd686dd979/lib/bundle/cask_dumper.rb#L29-L30 may be a better approach than calling cask internals given that.

danielbayley:
> @danielbayley
> 
> https://github.com/Homebrew/homebrew-bundle/blob/9aed384937bdf545b5c4e3e1fcc0a2bd686dd979/lib/bundle/cask_dumper.rb#L29-L30
> 
> may be a better approach than calling cask internals given that.

Well doing that inside the `cask_line` method, shelling out for _every_ cask obviously has horrible performance, making the command basically unusable… So I have refactored it some and pushed the changes, which work nicely. However this has broken a bunch of tests, so not sure exactly how to proceed, fixing or rewriting them… @MikeMcQuaid What do you think?

lib/bundle/cask_dumper.rb
```diff
@@ -6,6 +6,7 @@ module CaskDumper
 
     def reset!
       @casks = nil
+      @cask_info_response = nil
```
MikeMcQuaid, line 9:
```suggestion
      @cask_info = nil
```

lib/bundle/cask_dumper.rb
```diff
@@ -16,25 +17,30 @@ def casks
...
+      desc = cask_info.flat_map.filter { |info| info["token"] == cask }.first["desc"]
+      comment = "# #{desc}\n" if describe && !desc.blank?
+      "#{comment}cask \"#{cask}\""
+    end
 
-      cask_info_response = `brew cask info #{cask_list.join(" ")} --json=v1`
-      cask_info = JSON.parse(cask_info_response)
+    def cask_info
+      @cask_info_response ||= `brew cask info #{casks.join(" ")} --json=v1`
+      JSON.parse(@cask_info_response)
```
MikeMcQuaid, line 35:
```suggestion
      @cask_info ||= begin
        output = `brew cask info #{casks.join(" ")} --json=v1`
        JSON.parse(output)
```

lib/bundle/cask_dumper.rb
```diff
@@ -16,25 +17,30 @@ def casks
...
-        (casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
+        (casks & casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
+        (casks - casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
       ]
     end
 
-    def formula_dependencies(cask_list)
-      return [] unless cask_list.present?
+    def cask_line(cask, describe)
+      desc = cask_info.flat_map.filter { |info| info["token"] == cask }.first["desc"]
```
MikeMcQuaid, line 28:
```suggestion
      desc = cask_info.flat_map.find { |info| info["token"] == cask }&["desc"]
```

MikeMcQuaid (commented):
Looking good! I believe my suggestions should get CI passing and me be 👍🏻 on the actual code. After that it'll just be a matter of adding some more tests to get the coverage up.

danielbayley:
> I believe my suggestions should get CI passing

They didn't 😅 @MikeMcQuaid I don't see where `formula_dependencies` is even being called outside of the specs…?

lib/bundle/cask_dumper.rb
```diff
@@ -16,25 +17,32 @@ def casks
...
+    end
 
-      cask_info_response = `brew cask info #{cask_list.join(" ")} --json=v1`
-      cask_info = JSON.parse(cask_info_response)
+    def cask_info
+      @cask_info ||= begin
+        output = `brew cask info #{casks.join(" ")} --json=v1`
+        JSON.parse(output)
+      rescue JSON::ParserError => e
+        opoo "Failed to parse `brew cask info --json`: #{e}"
```
MikeMcQuaid, line 38:
```suggestion
        opoo "Failed to parse `brew cask info --json`: #{e}"
        []
```

lib/bundle/cask_dumper.rb
```diff
@@ -16,25 +17,32 @@ def casks
...
-        (casks - casks_required_by_formulae).map { |cask| "cask \"#{cask}\"" }.join("\n"),
+        (casks & casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
+        (casks - casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
       ]
     end
 
-    def formula_dependencies(cask_list)
-      return [] unless cask_list.present?
+    def cask_line(cask, describe)
+      desc = cask_info.flat_map.find { |info| info["token"] == cask }&["desc"] if cask_info
```
MikeMcQuaid, line 28:
```suggestion
      desc = cask_info.flat_map.find { |info| info["token"] == cask }&["desc"]
```

MikeMcQuaid:
> They didn't 😅 @MikeMcQuaid I don't see where `formula_dependencies` is even being called outside of the specs…?

It says in the CI failure message: https://github.com/Homebrew/homebrew-bundle/pull/810/checks?check_run_id=1373792286#step:9:56

danielbayley:
@MikeMcQuaid I reintroduced the arguments to `formula_dependencies` and `cask_info` to pass the CI… but not sure if that feels like the right approach.

MikeMcQuaid:
> but not sure if that feels like the right approach.

I agree. I'd suggest changing the tests (https://github.com/Homebrew/homebrew-bundle/blob/8588817eb92495a4461d3d2ed5af4903afb638cb/spec/bundle/cask_dumper_spec.rb#L65-L92) to test the functionality without arguments being passed (and doing stubbing/mocking if necessary).

lib/bundle/cask_dumper.rb
```diff
@@ -16,25 +17,33 @@ def casks
...
+        (casks & casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
+        (casks - casks_required_by_formulae).map { |cask| cask_line(cask, describe) }.join("\n"),
       ]
     end
 
-    def formula_dependencies(cask_list)
-      return [] unless cask_list.present?
+    def cask_line(cask, describe)
+      desc = cask_info&.flat_map.find { |info| info["token"] == cask }&["desc"]
+      comment = "# #{desc}\n" if describe && !desc.blank?
```
MikeMcQuaid, line 29:
```suggestion

      comment = if describe
        desc = cask_info.flat_map.find { |info| info["token"] == cask }&["desc"]
        "# #{desc}\n" if desc.present?
      end
```

`cask_info&` is unnecessary because `cask_info` will never return `nil`. If it did: `formula_dependencies` would break too.

MikeMcQuaid:
@danielbayley Please try to run the tests locally rather than repeatedly pushing, thanks.

danielbayley:
> doing stubbing/mocking if necessary

@MikeMcQuaid I was attempting to do that before… but I'm not familiar with RSpec. How do I override the `casks` and/or `cask_info` method/s in the `before do` block so that `formula_dependencies` picks up `["foo"]` instead of passing it directly as before?

lib/bundle/cask_dumper.rb
MikeMcQuaid: This was not resolved.

MikeMcQuaid:
@danielbayley `allow(Bundle).to receive(:cask_installed?).and_return(true)` is an example of this. See other uses of `allow` and `expect` in the relevant test files.

jacobbednarz:
This PR will need to be updated as the `cask` functionality has been updated to use the newer version of JSON output.

danielbayley:
@BrewTestBot Coming back to it!

Issue #815, opened 2020-11-20
Cask with formula of same name is treated as formula, upgrade fails


kbd:
`osquery` has a formula and a cask of the same name:

```shell
$ brew search osquery
==> Formulae
osquery
==> Casks
osquery ✔
```

As you can see, I have the cask installed but not the formula.

In my `Brewfile`, I have the line:

```
cask "osquery"
```

But when Bundle goes to update the cask, I get: 

```
...
Using libreoffice
Using meld
Upgrading osquery cask. It is installed but not up-to-date.
Warning: Treating osquery as a formula. For the cask, use homebrew/cask/osquery
Error: osquery not installed
Installing osquery has failed!
Using rq
Using slack
...
```

When I specifically upgraded the `osquery` cask it completed successfully:

```
$ brew cask upgrade osquery
Warning: Calling brew cask upgrade is deprecated! Use brew upgrade --cask instead.
==> Upgrading 1 outdated package:
osquery 4.5.0 -> 4.5.1
==> Upgrading osquery
==> Downloading https://pkg.osquery.io/darwin/osquery-4.5.1.pkg
######################################################################## 100.0%
==> Verifying SHA-256 checksum for Cask 'osquery'.
==> Removing launchctl service com.facebook.osqueryd
Password:
==> Uninstalling packages:
...
installer: The install was successful.
==> Purging files for version 4.5.0 of Cask osquery
🍺  osquery was successfully upgraded!
```

Is this a bug in Bundle? Or must I specify:

```
cask "homebrew/cask/osquery"
```

(is that correct?) to get it to install the cask even though the Brewfile is explicit that it's a cask?

MikeMcQuaid:
CC @reitermarkus as this looks like a possible regression from the new cask upgrade stuff.

PR #817, opened 2020-11-22
Switch to non-cask commands.

reitermarkus:
Fixes https://github.com/Homebrew/homebrew-bundle/issues/815 and closes https://github.com/Homebrew/homebrew-bundle/pull/816.


MikeMcQuaid:
Thanks @reitermarkus!

PR #792, opened 2020-09-13
Fix `brew cask` deprecation warnings

issyl0:
- A lot of the separation between Homebrew formulae and casks was deprecated in [Homebrew 2.5.0](https://brew.sh/2020/09/08/homebrew-2.5.0/), released last week. To avoid `brew bundle` showing deprecation warnings to users, update the invocations.
- Fixes #791.

PR #820, opened 2020-11-25
cask_dumper: Replace `brew cask info` for `brew info --cask`

jacobbednarz:
Follows on #817 and #792 by updating the `brew cask info` commands to
use the new syntax of `brew info --cask` instead.

cc @issyl0


issyl0 (approved):
Refactoring too fast, can't keep up!

Issue #822, opened 2020-11-26
JSON v2 Parsing Issue


astromonkee:
Hi,

The recent changes here https://github.com/Homebrew/homebrew-bundle/commit/b3da7adcbb4a889724d5bcd96d05937d122964f6 is causing some issues when I execute `brew bundle`. 

```Error: no implicit conversion of String into Integer
Please report this bug:
  https://github.com/Homebrew/homebrew-bundle/issues
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:92:in `block in brew_list_info'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:91:in `each'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:91:in `each_with_object'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:91:in `brew_list_info'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:44:in `block in lock'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:36:in `each'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/locker.rb:36:in `lock'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/installer.rb:57:in `install'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/commands/install.rb:10:in `run'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle/cmd/bundle.rb:94:in `bundle'
/usr/local/Homebrew/Library/Homebrew/brew.rb:119:in `<main>'
```

It looks like it's related to the format of the JSON when using `v2`. Here are the outputs from my machine.
`brew info --json=v1 --installed --quiet`
https://gist.github.com/astromonkee/c9d5b6c4119abe4d0790126ee3979c27

`brew info --json=v2 --installed --quiet`
https://gist.github.com/astromonkee/1c6f290111d8d7969b33d805efe65465

PR #823, opened 2020-11-26
Fix formulae parsing issue

astromonkee:
Possible fix for https://github.com/Homebrew/homebrew-bundle/issues/822


jacobbednarz:
Thanks @astromonkee!

PR #832, opened 2020-12-01
Create `add` and `remove` subcommands for `brew bundle`

johndbritton:
As discussed in https://github.com/Homebrew/homebrew-bundle/issues/818

Add two subcommands `brew bundle add` and `brew bundle remove` which will add or remove a specified formula to the `Brewfile`. When the `--global` option is specified the formula is added or removed from the global `Brewfile`

I've figured out how to add new subcommands and also figured out how to read in the `Brewfile` and check if the requested formula is already in the `Brewfile`.

I'm looking for suggestions on how to add to the `Brewfile` and write it out to disk. My hope is that there is some in memory way to manipulate the Brewfile and call dump. I looked at `Bundle::Dumper` but that seems to operate based on what's installed on the system, not what was parsed from the `Brewfile`.

Additionally, looking for feedback on the command line syntax for how we should handle formulae that are not `:brew` but are `:cask`, `:tap`, `:mas`, or any other type. For example `brew bundle add cask atom`.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -110,7 +110,7 @@ module Homebrew
       when "add"
         Bundle::Commands::Add.run(
           global: args.global?,
-          file:   args.file
+          file:   args.file,
         )
       when "dump"
         Bundle::Commands::Dump.run(
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -107,6 +107,11 @@ module Homebrew
             zap:    args.zap?,
           )
         end
+      when "add"
+        Bundle::Commands::Add.run(
+          global: args.global?,
+          file:   args.file
+        )
       when "dump"
         Bundle::Commands::Dump.run(
           global:     args.global?,
--- a/lib/bundle/commands/add.rb
+++ b/lib/bundle/commands/add.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Bundle
+  module Commands
+    module Add
+      module_function
+
+      def run(*args, global: false, file: nil)
+        raise UsageError, "No arguments were specified!" if args.blank?
+
+        type = :brew # default to brew
+        name = args.first # only support one formula at a time for now
+        options = {} # we don't currently support passing options
+
+        # read the relevant Brewfile
+        parsed_entries = Bundle::Dsl.new(Brewfile.read(global: global, file: file)).entries
+
+        # check each of the entries in the specified Brewfile
+        parsed_entries.each do |entry|
+          # raise an error if the entry already exists in the Brewfile
+          # this could also be a noop, or print a friendly message
+          raise RuntimeError if entry.name == name
+        end
+
+        # need some help / pointers here
+        # is it possible to use Bundle::Dsl to create an in memory representation
+        #of the brewfile that we read, add an entry and then dump that back to the file?
+      end
+    end
+  end
+end
--- a/lib/bundle/commands/add.rb
+++ b/lib/bundle/commands/add.rb
@@ -12,20 +12,20 @@ module Bundle
         name = args.first # only support one formula at a time for now
         options = {} # we don't currently support passing options
 
-        # read the relevant Brewfile
-        parsed_entries = Bundle::Dsl.new(Brewfile.read(global: global, file: file)).entries
+        # parse the relevant Brewfile
+        dsl = Bundle::Dsl.new(Brewfile.read(global: global, file: file))
 
         # check each of the entries in the specified Brewfile
-        parsed_entries.each do |entry|
+        dsl.entries.each do |entry|
           # raise an error if the entry already exists in the Brewfile
           # this could also be a noop, or print a friendly message
           opoo "'#{name}' already exists in Brewfile."
           raise RuntimeError if entry.name == name
         end
 
-        # need some help / pointers here
-        # is it possible to use Bundle::Dsl to create an in memory representation
-        # of the brewfile that we read, add an entry and then dump that back to the file?
+        dsl.brew(name, options)
+        # execute brew bundle
+        # execute brew bundle dump
       end
     end
   end
--- a/lib/bundle/commands/add.rb
+++ b/lib/bundle/commands/add.rb
@@ -19,12 +19,13 @@ module Bundle
         parsed_entries.each do |entry|
           # raise an error if the entry already exists in the Brewfile
           # this could also be a noop, or print a friendly message
+          opoo "'#{name}' already exists in Brewfile."
           raise RuntimeError if entry.name == name
         end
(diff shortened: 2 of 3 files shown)


cmd/bundle.rb
```diff
@@ -107,6 +107,11 @@ def bundle
             zap:    args.zap?,
           )
         end
+      when "add"
+        Bundle::Commands::Add.run(
+          global: args.global?,
+          file:   args.file
+        )
```
johndbritton, line 114: Do you want these commands to be alphabetized? Looks like they're just in any order.

lib/bundle/commands/add.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Commands
+    module Add
+      module_function
+
+      def run(*args, global: false, file: nil)
+        raise UsageError, "No arguments were specified!" if args.blank?
+
+        type = :brew # default to brew
+        name = args.first # only support one formula at a time for now
+        options = {} # we don't currently support passing options
```
johndbritton, line 13: Should we handle different types of formulae? Is there an existing pattern for how you'd like this handled?

lib/bundle/commands/add.rb
```diff
@@ -0,0 +1,31 @@
...
+        # check each of the entries in the specified Brewfile
+        parsed_entries.each do |entry|
+          # raise an error if the entry already exists in the Brewfile
+          # this could also be a noop, or print a friendly message
+          raise RuntimeError if entry.name == name
+        end
+
+        # need some help / pointers here
+        # is it possible to use Bundle::Dsl to create an in memory representation
+        #of the brewfile that we read, add an entry and then dump that back to the file?
```
johndbritton, line 27: This is where I could use the most help

jacobbednarz:
> I'm looking for suggestions on how to add to the Brewfile and write it out to disk. 

From my vague recollection, I can't remember anywhere that `bundle` didn't only care about what was installed on the system. What about slurping in the `--file` value (handling the default) and appending to the appropriate section? I don't think we need to be too tricky with it; something as simple as grouping/detecting and appending to that section seems more than enough here but open to thoughts on it.

> Additionally, looking for feedback on the command line syntax for how we should handle formulae that are not `:brew` but are `:cask`, `:tap`, `:mas`, or any other type. For example `brew bundle add cask atom`.

It looks like other parts of the ecosystem are moving to flags which seem suitable to use here too. For example `brew bundle add 1password --cask` would create a `cask "1password"`. Thoughts?

jacobbednarz:
Actually, I lie. There is `list` which seems to cover pretty much what I outlined above.

https://github.com/Homebrew/homebrew-bundle/blob/034459da8e6cd11076c578f0ac067fd50d1b14fa/lib/bundle/commands/list.rb#L11-L15

colindean:
I think we need to handle the case where there are options to add. From a fresh `brew bundle dump` on my dev machine:

```ruby
brew "paulp/extras/coursier", args: ["HEAD"]
brew "sshfs", link: false
brew "neovim", args: ["HEAD"]
brew "md5sha1sum", link: false
```

Will this also support adding a tap to the Brewfile, perhaps with --tap?

cmd/bundle.rb
```diff
@@ -107,6 +107,11 @@ def bundle
             zap:    args.zap?,
           )
         end
+      when "add"
+        Bundle::Commands::Add.run(
+          global: args.global?,
+          file:   args.file
```
MikeMcQuaid, line 113: `brew style --fix homebrew/bundle` will fix all these for you.

cmd/bundle.rb
MikeMcQuaid: any order is fine 👍🏻

lib/bundle/commands/add.rb
MikeMcQuaid: I think it's fine to just support `brew` for now and people can extend it later if desired. As long as it's documented it can `raise UsageError` which will print the help text.

lib/bundle/commands/add.rb
```diff
@@ -0,0 +1,31 @@
...
+        options = {} # we don't currently support passing options
+
+        # read the relevant Brewfile
+        parsed_entries = Bundle::Dsl.new(Brewfile.read(global: global, file: file)).entries
+
+        # check each of the entries in the specified Brewfile
+        parsed_entries.each do |entry|
+          # raise an error if the entry already exists in the Brewfile
+          # this could also be a noop, or print a friendly message
+          raise RuntimeError if entry.name == name
```
MikeMcQuaid, line 22: Yup, friendly message would be nice.

lib/bundle/commands/add.rb
MikeMcQuaid: This is the existing dumper code for this: https://github.com/Homebrew/homebrew-bundle/blob/034459da8e6cd11076c578f0ac067fd50d1b14fa/lib/bundle/brew_dumper.rb#L18-L39

It's probably possible to do that but you may have more luck trying to inject your desired formulae state into the existing logic (overriding `@formulae` somehow).

johndbritton:
I can't figure out a way to write out a `Brewfile` from an instance of `Bundle::Dsl`, so I'm considering a different approach.

Would it be reasonable for `brew bundle add` and `brew bundle remove` to add the specified formulae to the list specified in the `Brewfile`, install those items, and then dump the new state to the `Brewfile`?

https://github.com/Homebrew/homebrew-bundle/blob/78adf4d5c77415e1a7ee6256da3bf2ebfaee76ed/lib/bundle/commands/add.rb#L15-L28

I had originally planned to keep updating the `Brewfile` separate from actually doing the install but I had assumed that there was a straightforward way to update the `Brewfile`. It seems that we're always just using `dump` to output the formulae that are currently installed on the system.

I'm open to other ideas / suggestions if anyone has opinions.

MikeMcQuaid:
> Would it be reasonable for `brew bundle add` and `brew bundle remove` to add the specified formulae to the list specified in the `Brewfile`, install those items, and then dump the new state to the `Brewfile`?

I think the main issue I'd see with this is if the user has other things that are installed that would also be dumped to the Brewfile.



> It seems that we're always just using `dump` to output the formulae that are currently installed on the system.

Yup.

Something I said in https://github.com/Homebrew/homebrew-bundle/issues/818:

> Adding and removing feels like a step below what you want, right? Presumably you want a way of keeping your (--global?) Brewfile in sync with your installations. How about if there was something like brew bundle install wget --global --dump or something similar that ran brew install for you and, if it passed, redumped the global Brewfile automatically?

How about something like that? `brew bundle [install] wget --global --dump` and `brew bundle [install] wget --dump` to write to the global and local `Brewfile`s respectively (and the `install` is optional, as already).

PR #833, opened 2020-12-01
Remove pre-2.6.0 logic.

MikeMcQuaid:
Now that 2.6.0 has been released, this is no longer needed.

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -26,13 +26,7 @@ module Bundle
     def formula_dependencies(cask_list)
       return [] if cask_list.blank?
 
-      # TODO: can be removed when Homebrew 2.6.0 ships
-      cask_info_command = if HOMEBREW_VERSION > "2.5.12"
-        "brew info --cask --json=v2 #{cask_list.join(" ")}"
-      else
-        "brew info --json=v2 #{cask_list.join(" ")}"
-      end
-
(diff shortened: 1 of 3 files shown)


jacobbednarz:
Should we give 2.6.0 some time to propagate in the wild before merging this?

jacobbednarz (approved):
👍 with a non-blocking question

colindean (approved):
Second the propagation time suggestion.

Rylan12:
That shouldn't be a factor, I think. Users will get the updated `bundle` code by running `brew update` which will also update the brew code to 2.6.0, right?

jacobbednarz:
As I understand it, providing they use `brew update` yes however, if they update individual taps/repositories somehow this could get murky. I'm 👍 to not worrying about those cases too much though as `brew update` should get them back on the right path if so 🙂

MikeMcQuaid:
> Users will get the updated `bundle` code by running `brew update` which will also update the brew code to 2.6.0, right?

This.



> I'm 👍 to not worrying about those cases too much though as `brew update` should get them back on the right path if so 🙂

Agreed.

fjarlq:
Commit 0839a47 results in this error during `brew bundle cleanup`:

    Error: Invalid usage: this command requires a formula or cask argument
    Warning: Failed to parse `brew info --cask --json=v2`:
    767: unexpected token at ''

I've submitted PR #836 to fix this.

PR #836, opened 2020-12-02
fix error in cleanup

fjarlq:


MikeMcQuaid:
Whoops, good catch, thanks!

PR #842, opened 2020-12-11
Add ability to install only certain types with "bundle install"

lumaxis:
This is an initial attempt at adding a feature I've been wanting for myself many times.

From a `Brewfile` that contains several types of dependencies, I sometimes want to install only a certain type of dependency. Say, only Casks, for example or only apps from the Mac App Store.
Instinctively, I tried adding the `--brews` or `--casks` options that I know from `brew bundle dump` but they do nothing. This adds them to `brew bundle install`.

The code here may not be super polished yet but I'm opening this early to get feedback on the general idea and the initial approach. Let me know!


lumaxis:
Hm, odd 🤔 Both of the failing tests work for me locally. Will try tomorrow to figure out what's up unless somebody else has an idea.

lumaxis:
Ah, totally forgot that Homebrew behaves differently on Linux 😄

MikeMcQuaid:
> From a `Brewfile` that contains several types of dependencies, I sometimes want to install only a certain type of dependency. Say, only Casks, for example or only apps from the Mac App Store.

Can you explain why you wish to do this? As-is it doesn't seem like a widespread use-case which is why I'm tempted to pass on this 😢

lumaxis:
@MikeMcQuaid For sure! And no worries if this doesn't work out, no hard feelings 🙂 Maybe I'm making life harder for myself than necessary.

I'm worried about my dotfiles setup breaking since usually quite some time passes in between me setting up new machine from scratch. However, with the advent of Codespaces, I've been seeing myself do that more frequently.

So what I do today is run occasional Actions workflows that verify that my whole setup installs correctly on both OSes. To not have to install all Casks on a Mac Actions runner, I currently use separate files, a `Brewfile` and a `Caskfile` but these are pretty cumbersome to maintain since `brew bundle dump` also doesn't support types. Therefore, I have to manually copy the output to the respective files.
The idea of this PR was being able to use a single file but still be able to only install Brews, for example. This would be useful for my CI case but also for when I quickly want to setup a temporary Mac that I work on without having to do the lengthy and resource intensive installation of _all_ my Casks.
Additionally, I've recently been thinking about separating my Brewfiles into "personal" and "work" ones.

If you're still not convinced (which, again, is perfectly cool), I'm happy about feedback and maybe taking this into a different direction 👍🏼 Maybe a workflow with multiple files and adding support for type filtering to `bundle dump` or so after all.

colindean:
I separate into work, personal, and both.

MikeMcQuaid:
Sorry, I think I'm unconvinced here! The CI case I'd personally critique because it's not really doing CI if it's only testing a subset.

Thanks for the PR anyway!

PR #861, opened 2020-12-31
Fix `list --whalebrew`

jarrodldavis:
- [x] Update the list command's spec to ensure `brew bundle list --whalebrew` only lists whalebrew dependencies
- [x] Fix the lister to not display brew dependencies when the specific `whalebrew` option is enabled


jacobbednarz (approved):
Thanks!

PR #864, opened 2021-01-01
Switch to getting cask names from Cask module

### LINKED ISSUE #863: Don't show casks installed by other Homebrew installations
I'm currently running two installations of Homebrew on a M1 machine  with one under `/usr/local` and the other under `/opt/homebrew`. 

When I run `brew bundle dump` (using the Homebrew installed under `/opt/homebrew`) it only outputs casks installed by the other Homebrew installation, but not casks installed under `/opt/homebrew`. Is it possible to only list casks installed under `/opt/homebrew` if the `brew bundle` command was run by the Homebrew installation under `/opt/homebrew`?

I would be happy to contribute this feature if I was given some pointers on where to start.

Tyrubias:
Originally the cask names were obtained by running `brew list --cask` and parsing the output. Now we're getting the cask with the `Cask::Caskroom.casks` method and stripping out the name.

Fixes #863 

I know @jacobbednarz had some concerns about this feature. It turned out to be a simple change so I just created this pull request.

This is my first time contributing so I might've missed something. I'd be happy to have feedback!

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -11,7 +11,7 @@ module Bundle
     def casks
       return [] unless Bundle.cask_installed?
 
-      @casks ||= `brew list --cask 2>/dev/null`.split("\n")
+      @casks ||= Cask::Caskroom.casks.map(&:full_name).sort(&tap_and_name_comparison)
       @casks.map { |cask| cask.chomp " (!)" }
             .uniq
     end
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -13,7 +13,7 @@ module Bundle
 
       require "cask/caskroom"
 
-      @casks ||= Cask::Caskroom.casks.map(&:full_name).sort(&tap_and_name_comparison)
+      @casks ||= Cask::Caskroom.casks.map(&:full_name)
       @casks.map { |cask| cask.chomp " (!)" }
             .uniq
     end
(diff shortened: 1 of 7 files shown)


jacobbednarz:
You'll need to get the test suite green before this can be merged (likely a mocking issue). 

I'm on the fence here still as it seems like a deliberate decision to shell out to the list commands instead of reaching into a class not provided by this codebase. @MikeMcQuaid do you have any context on this decision?

Tyrubias:
I'm not very familiar with RSpec. I created a `stub/cask.rb` to mock `Cask::Caskroom` and tried adding
`allow(Cask).to receive(:casks).and_return([])` in `spec/bundle/cask_dumper_spec.rb` but I receive the error `uninitialized constant Cask`.

Do I need to edit `spec/spec_helper.rb` to include my new stubbed module in any way?

Tyrubias:
I figured out why my mocks weren't working. I'm not very experienced with this codebase, so some of my coding choices might be illogical. I'd be happy to have feedback 😃

MikeMcQuaid (approved):

MikeMcQuaid:
> I'm on the fence here still as it seems like a deliberate decision to shell out to the list commands instead of reaching into a class not provided by this codebase. @MikeMcQuaid do you have any context on this decision?

Yeh, this was before Casks were integrated into the Homebrew/brew codebase.

Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @Tyrubias!

PR #867, opened 2021-01-02
Don't get full cask name when dumping casks

### LINKED ISSUE #866: brew bundle doesn't handle external taps correctly
After PR #864 was merged, `brew bundle` doesn't handle external taps correctly, and always wants to install (or uninstall for `cleanup`) external taps.

Steps to reproduce:
```
$ cat Brewfile-test
tap "jabenninghoff/edge"
cask "jabenninghoff/edge/transcriptions"

$ brew bundle --file Brewfile-test install
Installing transcriptions

$ brew bundle check --file Brewfile-test
```

I expected the `brew bundle check` to note that the Brewfile dependencies are satisfied.

This also affects `brew clean` which always wants to remove casks from external taps.

Tyrubias:
When I switched to using `Cask::Caskroom` to get installed casks, I got the full name of the Cask (including the tap name, if any) instead of just the token. This PR should emulate the old behavior.

Fixes #866

### DIFF
--- a/spec/bundle/dumper_spec.rb
+++ b/spec/bundle/dumper_spec.rb
@@ -16,8 +16,8 @@ describe Bundle::Dumper do
     Bundle::WhalebrewDumper.reset!
     Bundle::BrewServices.reset!
 
-    chrome = instance_double("Cask::Cask", full_name: "google-chrome")
-    java = instance_double("Cask::Cask", full_name: "java")
+    chrome = instance_double("Cask::Cask", to_s: "google-chrome")
+    java = instance_double("Cask::Cask", to_s: "java")
 
     allow(Cask::Caskroom).to receive(:casks).and_return([chrome, java])
     allow(Bundle::CaskDumper).to receive(:`).and_return("google-chrome\njava")
--- a/spec/bundle/dumper_spec.rb
+++ b/spec/bundle/dumper_spec.rb
@@ -16,15 +16,16 @@ describe Bundle::Dumper do
     Bundle::WhalebrewDumper.reset!
     Bundle::BrewServices.reset!
 
-    chrome = instance_double("Cask::Cask", to_s: "google-chrome")
-    java = instance_double("Cask::Cask", to_s: "java")
+    chrome = instance_double("Cask::Cask", full_name: "google-chrome", to_s: "google-chrome")
+    java = instance_double("Cask::Cask", full_name: "java", to_s: "java")
+    iterm2beta = instance_double("Cask::Cask", full_name: "homebrew/cask-versions/iterm2-beta", to_s: "iterm2-beta")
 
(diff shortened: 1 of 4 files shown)


spec/bundle/dumper_spec.rb
```diff
@@ -16,8 +16,8 @@
     Bundle::WhalebrewDumper.reset!
     Bundle::BrewServices.reset!
 
-    chrome = instance_double("Cask::Cask", full_name: "google-chrome")
-    java = instance_double("Cask::Cask", full_name: "java")
+    chrome = instance_double("Cask::Cask", to_s: "google-chrome")
+    java = instance_double("Cask::Cask", to_s: "java")
```
jacobbednarz, line 20: As we identified a regression from a previous change, I'd think we would want to add a test case covering that scenario. Thoughts?

Tyrubias: Seems like a good idea. Would that involve adding a double that's a cask from a tap? I'll think about how to implement that.

Tyrubias: I added a test case but I don't think it's actually useful. Could I have some advice on a good way to test this?

jacobbednarz: Which part don't you don't is useful?

spec/bundle/dumper_spec.rb
```diff
@@ -16,15 +16,16 @@
     Bundle::WhalebrewDumper.reset!
     Bundle::BrewServices.reset!
 
-    chrome = instance_double("Cask::Cask", full_name: "google-chrome")
-    java = instance_double("Cask::Cask", full_name: "java")
+    chrome = instance_double("Cask::Cask", full_name: "google-chrome", to_s: "google-chrome")
+    java = instance_double("Cask::Cask", full_name: "java", to_s: "java")
+    iterm2beta = instance_double("Cask::Cask", full_name: "homebrew/cask-versions/iterm2-beta", to_s: "iterm2-beta")
```
jacobbednarz, line 21: I'm not sure if this is the correct assertion as the issue we're having is with things like `cask "some/external/cask"` listed in the Brewfile ([example](https://github.com/jacobbednarz/j/blob/main/config/homebrew/Brewfile#L197)). I'm not entirely positive this solves the issue.

Tyrubias: In this case shouldn't I be checking a cask whose full name does include the tap and asserting that the output Brewfile doesn't contain one?

MikeMcQuaid:
Thanks again @Tyrubias!

spec/bundle/dumper_spec.rb
MikeMcQuaid: Looks right to me 👍🏻

jacobbednarz: Thanks for confirming!

PR #882, opened 2021-01-14
Use new cask syntax

EricFromCanada:
Migrate `brew cask install` mentions to `brew install --cask`.


jacobbednarz (approved):
thanking you!

PR #898, opened 2021-02-01
Don't shell out for casks or taps.

MikeMcQuaid:
Instead use the native APIs.

While we're here, make some class/module methods private too.


MikeMcQuaid:
Note to self: check the Linux coverage % after this is merged.

Issue #897, opened 2021-01-31
Ability to specify --greedy against 'cask' markers


aegershman:
For `cask`s which requires `--greedy` to be passed for evaluating upgrade eligibility, is there a way to specify the `--greedy` flag (or an equivalent) in a `Brewfile` when performing `brew bundle {install,check}` operations?

e.g., something such as `cask "spotify", greedy: true`? I thought about trying to use `args`, but it uses a key-value hash rather than arbitrary extra params being able to be passed in. And I'm under the impression something like `cask "spotify", args: {"greedy":true}` is... well, wrong.

Apologies if I am missing something straightforward, pI didn't see anything about this in issues or docs](https://docs.brew.sh/Manpage#bundle-subcommand), but forgive me if I missed it or I'm misunderstanding something.

It's definitely not the highest-priority thing in the world, nor end of the world if it isn't possible. _I will live_ with doing `brew outdated --cask --greedy` / `brew upgrade <thing> --cask --greedy`. But, figured I'd ask.

Thanks for your time 👍

MikeMcQuaid:
I don't believe this is supported, sorry.

MikeMcQuaid:
CC @EricFromCanada in case you have more knowledge/thoughts here.

aegershman:
No worries, thanks 👍

PR #900, opened 2021-02-03
Add support for greedy cask upgrades 

EricFromCanada:
Allow currently-installed casks with `auto_updates true` or `version :latest` to be upgraded if `greedy: true` is appended to their Brewfile entry. Solves #897.

Also, start using [`explicit_s`](https://github.com/Homebrew/brew/pull/10462) to generate explicit cask config lines.


MikeMcQuaid (approved):

MikeMcQuaid:
@EricFromCanada looks great, thanks for this! Got some missing code coverage here, though.

MikeMcQuaid:
Thanks again @EricFromCanada!

PR #917, opened 2021-02-11
Use consistent array/hash indentation

Bo98:
https://github.com/Homebrew/brew/pull/10594


MikeMcQuaid (approved):
:+1: but a chicken-and-egg here because this CI won't pass until Homebrew/brew has changed.

PR #930, opened 2021-02-26
exec: Allow the use of commands not in the PATH

ticky:
This extends `brew bundle exec` with the ability to run a command which isn't in the path, for example, my use case was running a `./configure` script using the Brewfile magic, but this wasn't supported.

My patch basically just skips the storage of the path to the command in `command_path` as well as the related null if the command has a directory separator in it anywhere, and then uses `exec` as normal.

In effect, this brings it up to acting more like a shell would, meaning you can run a configure script via `brew bundle exec` as `./configure`, but not as `configure`. On the other hand, you can now run something like `brew bundle exec bin/setup`, or pass an absolute path to an executable.

Specs have been added for the three types of paths this should interact with; absolute and relative paths, and relative paths which do not contain a `.` indicating the current directory.

Note that this branch does not have 100% code coverage as requested in CONTRIBUTING.md, but this is not due to this change, and in fact already the case on `master`: https://github.com/Homebrew/homebrew-bundle/runs/1976844086?check_suite_focus=true#step:9:47

### DIFF
--- a/lib/bundle/commands/exec.rb
+++ b/lib/bundle/commands/exec.rb
@@ -17,10 +17,10 @@ module Bundle
         command = args.first
 
         # For commands which aren't either absolute or relative
-        unless command.include? "/"
+        if command.exclude? "/"
           # Save the command path, since this will be blown away by superenv
           command_path = which(command)
-          raise "command was not found in your PATH: #{command}" if command_path.nil?
+          raise "command was not found in your PATH: #{command}" if command_path.blank?
 
           command_path = command_path.dirname.to_s
         end
@@ -47,7 +47,7 @@ module Bundle
         ENV.prepend_path "PATH", pkgconfig.opt_bin.to_s if pkgconfig.any_version_installed?
 
         # Ensure the Ruby path we saved goes before anything else, if the command was in the PATH
-        ENV.prepend_path "PATH", command_path unless command_path.nil?
+        ENV.prepend_path "PATH", command_path if command_path.present?
 
         exec(*args)
       end
--- a/lib/bundle/commands/exec.rb
+++ b/lib/bundle/commands/exec.rb
@@ -16,11 +16,14 @@ module Bundle
 
         command = args.first
 
-        # Save the command path, since this will be blown away by superenv
-        command_path = which(command)
-        raise "command was not found in your PATH: #{command}" if command_path.nil?
+        # For commands which aren't either absolute or relative
+        unless command.include? "/"
+          # Save the command path, since this will be blown away by superenv
+          command_path = which(command)
+          raise "command was not found in your PATH: #{command}" if command_path.nil?
 
-        command_path = command_path.dirname.to_s
+          command_path = command_path.dirname.to_s
+        end
 
         brewfile = Bundle::Dsl.new(Brewfile.read(global: global, file: file))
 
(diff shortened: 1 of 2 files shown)


ticky:
Oh, and an additional note: I don't believe a documentation update is required, as the documentation for `brew bundle exec` already just says `<command>`, which is actually why I thought this would handle relative paths already!

https://github.com/Homebrew/homebrew-bundle/blob/ccdee2544f1ad5ca4e6862cca0107422bf51be0d/cmd/bundle.rb#L41-L42

mistydemeo (commented):
This makes sense to me! One idea - what if, instead of checking for a `/`, we check if `File.exist?(command)`?

ticky:
> This makes sense to me! One idea - what if, instead of checking for a `/`, we check if `File.exist?(command)`?

That would make the behaviour pretty different from the terminal norms; you can't run a bare file in the current directory without `./`, but `File.exist?` will return true for that sort of path.

lib/bundle/commands/exec.rb
```diff
@@ -16,11 +16,14 @@ def run(*args, global: false, file: nil)
 
         command = args.first
 
-        # Save the command path, since this will be blown away by superenv
-        command_path = which(command)
-        raise "command was not found in your PATH: #{command}" if command_path.nil?
+        # For commands which aren't either absolute or relative
+        unless command.include? "/"
```
MikeMcQuaid, line 20:
```suggestion
        if command.exclude? "/"
```

lib/bundle/commands/exec.rb
```diff
@@ -16,11 +16,14 @@ def run(*args, global: false, file: nil)
...
+          # Save the command path, since this will be blown away by superenv
+          command_path = which(command)
+          raise "command was not found in your PATH: #{command}" if command_path.nil?
```
MikeMcQuaid, line 23:
```suggestion
          raise "command was not found in your PATH: #{command}" if command_path.blank?
```

lib/bundle/commands/exec.rb
```diff
@@ -43,8 +46,8 @@ def run(*args, global: false, file: nil)
         pkgconfig = Formulary.factory("pkg-config")
         ENV.prepend_path "PATH", pkgconfig.opt_bin.to_s if pkgconfig.any_version_installed?
 
-        # Ensure the Ruby path we saved goes before anything else
-        ENV.prepend_path "PATH", command_path
+        # Ensure the Ruby path we saved goes before anything else, if the command was in the PATH
+        ENV.prepend_path "PATH", command_path unless command_path.nil?
```
MikeMcQuaid, line 50:
```suggestion
        ENV.prepend_path "PATH", command_path if command_path.present?
```

MikeMcQuaid (commented):
Feature makes sense to me! A couple of style suggestions (that `brew style` may flag on some of them).

ticky:
> that `brew style` may flag on some of them

`brew style` is green already for me, but I'll look at updating with your feedback! ☺️

ticky:
Okay, this is updated with the style suggestions you proposed, and I've confirmed it's all still working.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @ticky!

colindean:
This is awesome! Thank you so much, @ticky!

PR #950, opened 2021-04-26
Use codecov action instead of gem

colindean:
Closes #949

~~WIP: Something's messed up with my Homebrew installation so I took a pass at this just editing in GitHub while brew was doing stuff in the background. May or may not work; I'll fix it tomorrow if it doesn't work.~~ It worked! https://codecov.io/gh/Homebrew/homebrew-bundle/tree/6018480b61eb8a1703a7758a99bbe554cfe92328

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,5 +43,5 @@ jobs:
 
     - name: Run RSpec tests
       run: bundle exec rspec
-      env:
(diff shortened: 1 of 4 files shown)


jacobbednarz (approved):

colindean:
Rebased and cleaned up better in https://github.com/Homebrew/homebrew-bundle/pull/951

PR #951, opened 2021-04-26
Use codecov action instead of gem

colindean:
* Adds codecov action at end of test run
* Adds simplecov-cobertura gem
* Remove codecov gem
* No need to require simplecov, at least according to [this][1]

Closes #949 since it obsoletes the need to upgrade the codecov gem

[1]: https://github.com/simplecov-ruby/simplecov#getting-started


jacobbednarz (approved):

MikeMcQuaid:
Looks right, nice work @colindean!

PR #966, opened 2021-05-22
dependabot: ignore updates to `actions/stale` and `dessant/lock-threads`

Rylan12:
See https://github.com/Homebrew/brew/issues/11401

These actions should be updated by dependabot in Homebrew/.github so, to avoid extra PRs, let's ignore them here.


issyl0 (approved):
Good call. Thanks!

Issue #803, opened 2020-10-24
Remove tap warning for tap-prefixed formulae


jasonkarns:
Some behavior of brew-bundle surprised me...

Given a Brewfile that contains a tap-prefixed formulae, say `brew "nejckorasa/tap/dckr"`, the output of `brew-bundle` generates a warning about tap being necessary:

```$ brew bundle
Warning: 'nejckorasa/tap/dckr' formula is unreadable: No available formula with the name "nejckorasa/tap/dckr" 
Please tap it and then try again: brew tap nejckorasa/tap
Installing nejckorasa/tap/dckr
Homebrew Bundle complete! 1 Brewfile dependencies now installed.
```

The warning seems unnecessary for two reasons:
1. installing the formula directly will automatically tap as necessary (without the warning):
    ```
    $ brew install nejckorasa/tap/dckr
    ==> Tapping nejckorasa/tap
    Cloning into '/usr/local/Homebrew/Library/Taps/nejckorasa/homebrew-tap'...
    remote: Enumerating objects: 93, done.
    remote: Total 93 (delta 0), reused 0 (delta 0), pack-reused 93
    Unpacking objects: 100% (93/93), 18.58 KiB | 279.00 KiB/s, done.
    Tapped 2 formulae (120 files, 59KB).
    ==> Installing dckr from nejckorasa/tap
    ==> Downloading https://github.com/nejckorasa/mac-docker-go/archive/v1.1.2.tar.gz
    Already downloaded: /usr/local/cache/Homebrew/downloads/47a5d32dc932fc4ef42244d0293476000b788a203905d4ca167559ef8e91474a--mac-docker-go-1.1.2.tar.gz
    🍺  /usr/local/Cellar/dckr/1.1.2: 4 files, 6.8KB, built in 3 seconds
    ```
2. brew-bundle itself still proceeds to tap and install the formula without requiring an explicit tap

So the tap warning seems unnecessary for brew-bundle; and is inconsistent with direct brew-install (which doesn't warn about tap at all).

[...]

PR #982, opened 2021-07-20
Skip dependents if a tap fails

quad:
If a `tap` fails, then any subsequent commands can grossly misbehave. For example:

```
tap 'example/formula', 'user@git.example.com:example/homebrew-formulas.git'

brew 'example/formula/internal-tool'
```

If `tap` fails, then bundle currently continues and will attempt to tap (and clone) `https://github.com/example/homebrew-formula` to much sad failure.

This PR changes `bundle` to skip any dependent installations when a tap fails. The other stanzas, `brew`, `cask`, `mas`, `whalebrew`, and others are unchanged and do not abort; my rationale is that their failures are isolated and don't impact each other.

### DIFF
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -39,6 +39,10 @@ module Bundle
         when :skipped
           puts "Using #{entry.name}"
           success += 1
+        when :aborted
+          puts Formatter.error("#{verb} #{entry.name} has failed! Aborting!")
+          failure += 1
+          break
         else
           puts Formatter.error("#{verb} #{entry.name} has failed!")
           failure += 1
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -40,9 +40,9 @@ module Bundle
           puts "Using #{entry.name}"
           success += 1
         when :aborted
-          puts Formatter.error("#{verb} #{entry.name} has failed! Aborting!")
+          Bundle::Skipper.skip entry
+          puts Formatter.error("#{verb} #{entry.name} has failed! Skipping dependents...")
           failure += 1
-          break
         else
           puts Formatter.error("#{verb} #{entry.name} has failed!")
           failure += 1
--- a/lib/bundle/skipper.rb
+++ b/lib/bundle/skipper.rb
@@ -32,8 +32,7 @@ module Bundle
       alias generic_skip? skip?
 
       def skip(entry)
-        skipped_entries[entry.type]&.<< entry.name ||
-          (skipped_entries[entry.type] = [entry.name])
+        (skipped_entries[entry.type] ||= []) << entry.name
       end
 
       private
--- a/lib/bundle/skipper.rb
+++ b/lib/bundle/skipper.rb
@@ -32,8 +32,8 @@ module Bundle
       alias generic_skip? skip?
 
       def skip(entry)
-        skipped_entries[entry.type] &.<< entry.name or
-          skipped_entries[entry.type] = [entry.name]
+        skipped_entries[entry.type]&.<< entry.name ||
+          (skipped_entries[entry.type] = [entry.name])
       end
 
       private
--- a/lib/bundle/skipper.rb
+++ b/lib/bundle/skipper.rb
@@ -12,6 +12,11 @@ module Bundle
           return true
         end
 
+        return true if skipped_entries[:tap]&.any? do |tap|
+          prefix = "#{tap}/"
+          entry.name.start_with?(prefix) || entry.options[:full_name]&.start_with?(prefix)
+        end
+
         entry_type_skips = Array(skipped_entries[entry.type])
         return false if entry_type_skips.empty?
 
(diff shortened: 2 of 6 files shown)


MikeMcQuaid:
Thanks for this @quad, I really like this idea. I'd love to see the scope narrowed a little bit so e.g. `tap "cantino/mcfly"` only causes an immediate failure if e.g. `brew "cantino/mcfly/mcfly"` is present. 

Thoughts?

quad:
I thought about that but went this PR for the following reasons:

1. The installer would need to look ahead, making it more complex
2. A tap prefix is optional. e.g. `brew "internalpackage"` and `brew "example/formula/internalpackage"` are the same
3. Taps can change homebrew's behaviour, e.g. services or … bundle 😁 
4. When would a tap not be used?

I don't have the confidence I can get the corner cases right. _Especially_ since this PR is driven by a corner-case.

MikeMcQuaid:
> * A tap prefix is optional. e.g. `brew "internalpackage"` and `brew "example/formula/internalpackage"` can resolve to the same formula

They can but: the latter will try to tap it and the prior will not. `brew bundle dump` always outputs the latter format (and may do for casks, I'm not sure).

> * Taps can change homebrew's behaviour, e.g. services or … bundle 😁

Kinda; these add behaviour/commands but arguably don't change it. They are also both auto tapped when run.

> 4\. When would a tap not be used?

I've seen cases where people want to have a tap just so they can install from the packages available in it but might not want it to be fatal.

--- 

I'm still open to merging this as-is. @Homebrew/bundle any thoughts from you folks?

gromgit:
Since:

1. the majority of installs would likely be from core taps
2. core taps aren't likely to be b0rked
3. b0rked taps are not very likely if the Brewfile is recent

a more user-friendly approach might be:

1. If a tap fails, print an error and flag the tap name, but continue.
2. If a formula/cask begins with a flagged tap name, print a warning and continue.
3. Consolidate and print a list of failures at the end (this might already be happening, but I haven't had any bundle failures so far, so I can't be sure).

MikeMcQuaid:
> * the majority of installs would likely be from core taps
> * core taps aren't likely to be b0rked
> * b0rked taps are not very likely if the Brewfile is recent

👍🏻 agreed with all this.

> * If a formula/cask begins with a flagged tap name, print a warning and continue.

I think this should be extended to "don't try to install it and print a warning stating why". This would avoid the need for lookahead while also partially addressing e.g. https://github.com/Homebrew/homebrew-bundle/issues/803..

> 3\. Consolidate and print a list of failures at the end (this might already be happening, but I haven't had any bundle failures so far, so I can't be sure).

We don't do this but the output is minimal enough that I think it's probably not necessary.

quad:
> > * A tap prefix is optional. e.g. `brew "internalpackage"` and `brew "example/formula/internalpackage"` can resolve to the same formula
> 
> They can but: the latter will try to tap it and the prior will not. `brew bundle dump` always outputs the latter format (and may do for casks, I'm not sure).

👌 

> > * Taps can change homebrew's behaviour, e.g. services or … bundle 😁
> 
> Kinda; these add behaviour/commands but arguably don't change it. They are also both auto tapped when run.

External commands from internal taps aren't auto-tapped.

> > 4. When would a tap not be used?
> 
> I've seen cases where people want to have a tap just so they can install from the packages available in it but might not want it to be fatal.

👌 Good point!

quad:
> a more user-friendly approach might be:
> 
>     1. If a tap fails, print an error and flag the tap name, but continue.
>     2. If a formula/cask begins with a flagged tap name, print a warning and continue.
>     3. Consolidate and print a list of failures at the end (this might already be happening, but I haven't had any bundle failures so far, so I can't be sure).

I will make this change tomorrow morning AEST. It fixes my issue, so yay.

😢 that I'll be the one to make the install loop more complex. I was impressed / easily able to contribute thanks to its straight-forwardness!

MikeMcQuaid:
@quad definitely don't worry about that, we're super glad to have you involved! ❤️

quad:
@MikeMcQuaid 😰

MikeMcQuaid:
@quad some failing tests here and missing test coverage but I think we're really close 😍

quad:
@MikeMcQuaid I made that line more idiomatic.

(It's been a minute since I've written Ruby in anger.)

quad:
@MikeMcQuaid Tests are passing. I'll investigate coverage.

MikeMcQuaid:
@quad CI still failing I'm afraid. I think I've now made it so I don't need to re-approve CI each time you push a new commit, though.

MikeMcQuaid:
@quad You may also wish to rebase on https://github.com/Homebrew/homebrew-bundle/pull/985 which will probably be merged before this PR is. Alternatively, just fix up CI failures here and I can fix up the merge conflicts for you.

quad (commented):
@MikeMcQuaid Linux build fixed.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @quad!

PR #992, opened 2021-08-05
Add groups for Brewfiles

jphastings:
✨ Extends the Brewfile DSL to allow 'group' declarations, and a new cli flag that excludes the given flags from being installed.

I ended up writing this before I realised how far I'd gotten, I figured I'd pause here and get feedback before I continue into tests and polish. **Does this seem interesting and/or supportable by the community?** I'd certainly find it useful (which is why I built it!) but maintenance of broadly unused code is hard!

## Example

A new `Brewfile` might look like this:

```Brewfile
# Naturally groups needn't be used at all, existing syntax is preserved entirely
brew "go"

# Groups can be specified as an option to all declarations
cask "slack", group: :work

# Specify a group with a block too
group :gaming do
  cask "discord"
  cask "steam"
end

# Multiple groups can be set at once, if needed
group :home, :work do
  brew "ipfs", restart_service: true
  mas "Synalyze It!", id: 417074533

  # The following line will break the Brewfile asgroups can't be declared as options within a group block
  cask "alfred", group: :anything
end
```

A user with the above `Brewfile` could bootstrap a new computer for home use like this:

```bash
$ brew bundle --without=gaming,work
## 
Installing go
Skipping slack (work group excluded)
Skipping discord (gaming group excluded)
Skipping steam (gaming group excluded)
Installing ipfs
Installing Synalyze It!
```
Notes:
- Items not in any groups can never be excluded, in the current implementation
- The last two are in the 'home' group, which isn't excluded, even though the 'work' group is — this matches `Gemfile`/bundler usage of the `--without-groups` parameter.
- I'm tempted to also offer an `--with` or `--with-only` option — but it is a little confusing as, unlike `bundle` this code doesn't have the concept of _optional_ groups.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -45,6 +45,8 @@ module Homebrew
       EOS
       flag   "--file=",
              description: "Read the `Brewfile` from this location. Use `--file=-` to pipe to stdin/stdout."
+      flag   "--without",
+             description: "Ignore entries tagged with these, comma separated, group names."
       switch "--global",
              description: "Read the `Brewfile` from `~/.Brewfile`."
       switch "-v", "--verbose",
@@ -84,6 +86,7 @@ module Homebrew
 
   def bundle
     args = bundle_args.parse
+    without_groups = args.without.split(',').map(&:to_sym)
 
     # Keep this after the .parse to keep --help fast.
     require_relative "../lib/bundle"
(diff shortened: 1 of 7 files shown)


jacobbednarz:
There has been some prior discussion in #248, and I don't think a whole heap has changed. Similar to @MikeMcQuaid I'm on the fence as I wouldn't be using it in this fashion and instead rely on individual project Brewfiles and a global Brewfile for other system wide requirements.

Like before, I'd probably leave this open and see if anyone else would be interested in it to gauge whether it's worth shipping and supporting longer term. I suspect this opens a bunch of new code paths and considerations but we can discuss those if we decide to proceed.

colindean:
I, too, wouldn't be using it in this fashion: I've got `Brewfile.all` and `Brewfile.{work.macos,personal.linux}` nowadays, plus project-specific Brewfiles in just about every work repo I touch.

I think the presence of `--without` implies the presence of `--only` or some synonym thereof. I think the complexities of this and what to do in the case of a package being in multiple groups is clearer through he maintenance of multiple files.

MikeMcQuaid:
Yeh, as-is I'm afraid I'm 👎🏻 on this. The main thing I think this doesn't really handle is how it fits in with `brew bundle dump`. If there's a better story for that: I wouldn't be necessarily opposed.

PR #993, opened 2021-08-05
workflows/tests: Test that `bundle`ing and running some software works

issyl0:
- In light of 79441868a6a2e1002d8e080c8045483c41b411e3, to avoid future regressions, I was inspired by the `brew services start` tests in [the`brew services` repo](https://github.com/Homebrew/homebrew-services/blob/master/.github/workflows/tests.yml) to make a very simple functional test that this actually installs what it says it does.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,5 +43,11 @@ jobs:
 
     - name: Run RSpec tests
       run: bundle exec rspec
+
+    - name: Test that installing works
+      run: |
+        echo 'brew "hello"' > ./Brewfile
+        brew bundle
+        hello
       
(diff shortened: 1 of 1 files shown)


colindean (approved):

colindean:
This gives me an idea for even more tests. I've been using [bats](https://github.com/bats-core/bats-core) for a lot of stuff at work and wouldn't mind flexing that a little more along with digging into GitHub Actions (haven't touched it yet!). I could do up some public API tests that treat the `brew bundle` command as an opaque box. This has worked well for me for some CLI projects I've done through the years. 

Thoughts, team?

MikeMcQuaid:
Looks good, thanks @issyl0!

PR #1008, opened 2021-09-17
installer.rb: change verb to "Upgrading" on upgradable formula/cask

amancevice:
Currently `brew bundle` logs upgrading formulae as `Installing <formula/cask>`.

This commit changes the verbiage to `Upgrading <fomula/cask>` when the Installer class method `{formula,cask}_upgradable?()` returns `true`

### DIFF
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -17,9 +17,11 @@ module Bundle
         cls = case type
         when :brew
           options = entry.options
+          verb = "Upgrading" if Bundle::BrewInstaller.formula_upgradable?(name)
           Bundle::BrewInstaller
         when :cask
           options = entry.options
+          verb = "Upgrading" if Bundle::CaskInstaller.cask_upgradable?(name)
           Bundle::CaskInstaller
         when :mas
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can you explain the motivation for this? I'm not opposed to it, just want to hear your thoughts (as `bundler` doesn't do this, for example).

amancevice:
Sure, no problem: We maintain a common `Brewfile`  in a private tap. Users are sometimes confused when they run `brew bundle` and it says something like "Installing awscli" when they know they already have it installed. I think this change makes it a little clearer what is happening. Sort of like int the pure `brew` world where a user would run `brew install` for new things and `brew upgrade` for existing things.

colindean:
I like this ux tweak. Just this past week, I was confused when setting up a development environment for a repo on a machine that I knew was already set up, get the verbiage was installing the package I knew that was already installed.

MikeMcQuaid:
Makes sense! 
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @amancevice!

Issue #1015, opened 2021-10-30
brew bundle dump cannot dump Casks with HOMEBREW_INSTALL_FROM_API set


fanvinga:
<img width="770" alt="image" src="https://user-images.githubusercontent.com/26292849/139516504-8c43ba07-30a6-4867-b3a4-d4f9c984e040.png">

PR #1016, opened 2021-10-31
`bundle`: fix `cask_installed?` with `HOMEBREW_INSTALL_FROM_API` set

Rylan12:
Fixes https://github.com/Homebrew/homebrew-bundle/issues/1015

`Bundle::cask_installed?` now returns `true` even if the `homebrew/cask` directory doesn't exist as long as `HOMEBREW_INSTALL_FROM_API` is set.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @Rylan12!

PR #1019, opened 2021-11-04
mac_app_store_installer: handle Monterey.

MikeMcQuaid:
`mas` is not yet functional on Monterey. To track progress or to *contribute* to fixing this issue, please see: https://github.com/mas-cli/mas/issues/417

PR #1021, opened 2021-11-05
Remove `mas account` check

chris-araman:
As of macOS 12 Monterey, `mas-cli` is no longer able to query whether the user is signed into the App Store.
https://github.com/mas-cli/mas/issues/417

I propose that instead of querying `mas account`, `brew bundle` directly attempts the `mas install` operation it intends to perform. If the user is not yet signed into the App Store, the App Store app appears with a prompt to sign in. This should continue to address the concerns expressed in #233/#235 by @ffittschen.

I also reverted #1019 per my comments [here](https://github.com/Homebrew/homebrew-bundle/pull/1019#issuecomment-961501752).

FYI @phatblat

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -30,6 +30,10 @@ module Bundle
     end
 
     def mas_signedin?
+      # mas account doesn't work on Monterey (yet)
+      # https://github.com/mas-cli/mas/issues/417#issuecomment-957963271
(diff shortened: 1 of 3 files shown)


anmlnath:
thanks

MikeMcQuaid:
Thanks @chris-aramam, you rock! I've accepted the revert as-is but wanted to tweak the `mas account` check so we can still use it pre-Monterey.

PR #1083, opened 2022-04-28
Emit brew install args when in verbose mode

colindean:
I'm setting up a Brewfile that will install the dependencies for a package that is in Homebrew but we're needing to install the package itself outside of Homebrew. I added `args: ['only-dependencies']` to the brew line but it wasn't working: the package itself was getting installed. So, I added this logging facility and realized that I'd typed "only_dependencies" instead. It seems prudent to show all args passed to `brew install` when verbosity is requested.

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -236,7 +236,8 @@ module Bundle
     end
 
     def install!(verbose:)
-      puts "Installing #{@name} formula#{" with args #{@args}" if !@args.empty?}. It is not currently installed." if verbose
+      with_args = @args.empty? ? "" : " with args #{@args}"
+      puts "Installing #{@name} formula#{with_args}. It is not currently installed." if verbose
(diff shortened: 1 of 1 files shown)


jacobbednarz (approved):

lib/bundle/brew_installer.rb
```diff
@@ -236,7 +236,8 @@ def resolve_conflicts!(verbose:)
     end
 
     def install!(verbose:)
-      puts "Installing #{@name} formula. It is not currently installed." if verbose
+      with_args = @args.empty? ? "" : " with args #{@args}"
```
MikeMcQuaid, line 239:
```suggestion
      with_args = " with args #{@args}" if @args.present?
```

as it'll evaluate to `nil` otherwise which outputs the same when interpolated

jacobbednarz (approved):

MikeMcQuaid:
Thanks again @colindean!

PR #1084, opened 2022-04-28
Add support for arbitrary args to tap

### LINKED ISSUE #1063: Add support for --force-auto-update flag for taps
I could just be missing documentation, but I couldn't see a way to run 
```
brew tap --force-auto-update user/tap-repo https://user@bitbucket.org/user/homebrew-tap-repo.git
```

using a Brewfile. Does a way exist or does the feature need to be added?

marquiswang:
Fixes #1063. Allows arguments to taps in a Brewfile with:

  tap "foo/bar", "https://host/path/to/tap", args: ["force-auto-update"]

Mostly useful for adding --force-auto-update for non-Github taps.

Syntax is an array of arguments, like with `brew`, rather than a
dictionary like `cask` - this is because none of the optional arguments
of `brew tap` take a value right now. In the future, we should be able
to add support for dictionaries as well without needing to change the
format, if that becomes necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ An example `Brewfile`:
 tap "homebrew/cask"
 # 'brew tap' with custom Git URL
 tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git"
+# 'brew tap' with arguments
+tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", force_auto_update: true
+
 # set arguments for all 'brew install --cask' commands
 cask_args appdir: "~/Applications", require_sha: true
 
--- a/lib/bundle/dsl.rb
+++ b/lib/bundle/dsl.rb
@@ -74,14 +74,15 @@ module Bundle
       @entries << Entry.new(:whalebrew, name)
     end
 
-    def tap(name, clone_target = nil)
+    def tap(name, clone_target = nil, options = {})
       raise "name(#{name.inspect}) should be a String object" unless name.is_a? String
       if clone_target && !clone_target.is_a?(String)
         raise "clone_target(#{clone_target.inspect}) should be nil or a String object"
       end
 
+      options[:clone_target] = clone_target
       name = Bundle::Dsl.sanitize_tap_name(name)
-      @entries << Entry.new(:tap, name, clone_target: clone_target)
+      @entries << Entry.new(:tap, name, options)
     end
 
     HOMEBREW_TAP_ARGS_REGEX = %r{^([\w-]+)/(homebrew-)?([\w-]+)$}.freeze
--- a/lib/bundle/tap_installer.rb
+++ b/lib/bundle/tap_installer.rb
@@ -17,10 +17,13 @@ module Bundle
       return true unless preinstall
 
       puts "Installing #{name} tap. It is not currently installed." if verbose
+      args = []
+      args.append("--force-auto-update") if options[:force_auto_update]
+
       success = if options[:clone_target]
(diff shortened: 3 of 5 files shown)


colindean (changes requested):
I like and could use this myself for a company-internal tap that is frequently updated.

I'd like to see some tests for this new functionality, though.

marquiswang:
@colindean Tests added!

jacobbednarz (approved):

README.md
```diff
@@ -27,6 +27,9 @@ An example `Brewfile`:
 tap "homebrew/cask"
 # 'brew tap' with custom Git URL
 tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git"
+# 'brew tap' with arguments
+tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", args: ["force-auto-update"]
```
MikeMcQuaid, line 31:
```suggestion
tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", force_auto_update: true
```

What would you think about an API like this? Are there other arguments you need?

marquiswang: Can do. Would you prefer if we *only* support `--force-auto-update`, or allow both 
```
tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", force_auto_update: true
```

as well as

```
tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", args: ["arbitrary-argument"]
```

It is trivial to support both, as well as

```
tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", args: {arg1: true, "arg-two": "value"}
```

which would correspond to

```
--arg1 --arg-two=value
```

I will go ahead and push a version the does that. If folks prefer, I can delete all that and just support `force_auto_update`.

MikeMcQuaid: > Would you prefer if we _only_ support `--force-auto-update`

Yeh, only support I think.

marquiswang: Done!

colindean (approved):

lib/bundle/dsl.rb
```diff
@@ -74,14 +74,15 @@ def whalebrew(name)
       @entries << Entry.new(:whalebrew, name)
     end
 
-    def tap(name, clone_target = nil)
+    def tap(name, clone_target = nil, options = {})
```
MikeMcQuaid, line 77:
```suggestion
    def tap(name, clone_target = nil, force_auto_update = nil)
```

or similar?

lib/bundle/tap_installer.rb
```diff
@@ -17,10 +17,13 @@ def install(name, preinstall: true, verbose: false, **options)
       return true unless preinstall
 
       puts "Installing #{name} tap. It is not currently installed." if verbose
+      args = []
+      args.append("--force-auto-update") if options[:force_auto_update]
```
MikeMcQuaid, line 21: Could consider passing `options` as keyword arguments?

lib/bundle/dsl.rb
marquiswang: I could do that if you insist, but I think leaving it as `options = {}` is more consistent with the rest of this file. For example, `mas` also only accepts one possible option right now, `id`, and it also takes an `options` dictionary. But they are equivalent, so whichever you prefer!

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @marquiswang!

PR #1086, opened 2022-05-02
only remove lockfile if it is not symlink

quentin-fox:
Currently, the Brewfile.lock.json is always deleted prior to being written to. However, this is breaking for workflows where the Brewfile.lock.json is symlinked into the same directory as the Brewfile, as the symlink is deleted, then a regular file is written in its place. As a result, the file that the symlink references is not updated with the new version info.

This workflow probably isn't super common, but for users of `stow` (https://github.com/aspiers/stow), this might be a nice improvement. They would keep the `.Brewfile` and `.Brewfile.lock.json` in their (e.g.) `~/dotfiles/brew/` folder, and symlink both `.Brewfile` and `.Brewfile.lock.json` to the `~` directory. When running `brew bundle install --global`, it would read from the `.Brewfile` symlinked to `~/.Brewfile`, and write to the `.Brewfile.lock.json` file symlinked to `~/.Brewfile.lock.json`. Since the symlink to the lockfile would not be deleted (after this change), the changes from running `brew bundle install --global` would be reflected in the `~/dotfiles/brew/.Brewfile.lock.json` file.

I wasn't sure of the reason why the lockfile is deleted with `lockfile.unlink` prior to writing, so I didn't remove that line, but instead special-cased the instance where the lockfile exists, but is a symlink. Another possible fix would be to never unlink the lockfile? Calling `lockfile.write` already replaces the entire contents of the file, so I think this only breaks workflows where the lockfile is a directory... but not 100% sure on that case.

### DIFF
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -14,9 +14,9 @@ module Bundle
 
       # no need to call realpath if the lockfile is not a symlink
       # unnecessary call to fs, also breaks tests, which use filenames that are not in fs
-      return lock_file_path unless lock_file_path.symlink?
+      lock_file_path = lock_file_path.realpath if lock_file_path.symlink?
 
-      lock_file_path.realpath
+      lock_file_path
     end
 
     def write_lockfile?(global: false, file: nil, no_lock: false)
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -12,9 +12,10 @@ module Bundle
       brew_file_path = Brewfile.path(global: global, file: file)
       lock_file_path = brew_file_path.dirname/"#{brew_file_path.basename}.lock.json"
 
-      # will recursively resolve any symlinks to lock file, if present
-      # so that calling `unlink` on the lockfile will remove the original file
-      # not the symlinks
+      # no need to call realpath if the lockfile is not a symlink
+      # unnecessary call to fs, also breaks tests, which use filenames that are not in fs
+      return lock_file_path if !lock_file_path.symlink?
+
       lock_file_path.realpath
     end
 
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -75,7 +75,7 @@ module Bundle
 
       json = JSON.pretty_generate(lock)
       begin
-        lockfile.unlink if lockfile.exist?
+        lockfile.unlink if lockfile.exist? and not lockfile.symlink?
         lockfile.write("#{json}\n")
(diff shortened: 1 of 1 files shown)


colindean (approved):
Once the checks are fixed, this is good in my book. The reason makes sense to me.

jacobbednarz (approved):
LGTM, the only drawbacks I can see from this is that a file isn't removed (yet still overwritten) when someone expects it to be.

lib/bundle/locker.rb
```diff
@@ -75,7 +75,7 @@ def lock(entries, global: false, file: nil, no_lock: false)
 
       json = JSON.pretty_generate(lock)
       begin
-        lockfile.unlink if lockfile.exist?
+        lockfile.unlink if lockfile.exist? && !lockfile.symlink?
```
MikeMcQuaid, line 78: Feels a bit weird to just not unlink it if it's a symlink. Instead, how about resolving the symlink and changing `lockfile` to point to the destination file instead? Thoughts?

MikeMcQuaid (commented):
Thanks for the PR!

lib/bundle/locker.rb
quentin-fox: Makes a lot of sense to me, I like that more than not deleting the file. Alternatively, we could just never unlink the file? It could break some workflows, but we wouldn't have to resolve any symlinks (not that it's hard to do).

lib/bundle/locker.rb
```diff
@@ -10,7 +10,13 @@ module Locker
 
     def lockfile(global: false, file: nil)
       brew_file_path = Brewfile.path(global: global, file: file)
-      brew_file_path.dirname/"#{brew_file_path.basename}.lock.json"
+      lock_file_path = brew_file_path.dirname/"#{brew_file_path.basename}.lock.json"
+
+      # no need to call realpath if the lockfile is not a symlink
+      # unnecessary call to fs, also breaks tests, which use filenames that are not in fs
+      return lock_file_path unless lock_file_path.symlink?
```
MikeMcQuaid, line 17:
```suggestion
      lock_file_path = lock_file_path.realpath if lock_file_path.symlink?
```

lib/bundle/locker.rb
```diff
@@ -10,7 +10,13 @@ module Locker
...
+
+      lock_file_path.realpath
```
MikeMcQuaid, line 19:
```suggestion
      lock_file_path
```

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @quentin-fox!

PR #1092, opened 2022-05-30
Remove bottle_disabled?

MikeMcQuaid:
This will be removed in https://github.com/Homebrew/brew/pull/13347.


lib/bundle/brew_dumper.rb
```diff
@@ -149,8 +149,7 @@ def formula_to_hash(formula)
 
       runtime_dependencies ||= formula.runtime_dependencies.map(&:name)
 
-      bottled_or_disabled = formula.bottle_disabled?
-      bottled_or_disabled ||= if formula.bottle_defined?
+      bottled_or_disabled = if formula.bottle_defined?
```
jacobbednarz, line 152: nitpick: worth renaming this method to be what it is used for now?

jacobbednarz (approved):
👍 with a non-blocking comment

PR #1101, opened 2022-07-04
Instruct user that cleanup without --force takes no action

colindean:
#1099 surfaced a UX improvement that directly tells the user what's going on and what action to take since the default is safety-on.

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -38,6 +38,8 @@ module Bundle
           cleanup = system_output_no_stderr(HOMEBREW_BREW_FILE, "cleanup")
           puts cleanup unless cleanup.empty?
         else
+          puts "Listing actions for cleanup, but not taking action..."
+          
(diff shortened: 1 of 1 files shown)


jacobbednarz (approved):

MikeMcQuaid (approved):
Thanks @colindean!

lib/bundle/commands/cleanup.rb
```diff
@@ -59,7 +59,9 @@ def run(global: false, file: nil, force: false, zap: false)
             puts cleanup
           end
 
-          puts "Run `brew bundle cleanup --force` to make these changes."
+          if [casks.any?, formulae.any?, taps.any?, !cleanup.empty?].any?
```
MikeMcQuaid, line 62:
```suggestion
          if casks.any? || formulae.any? || taps.any? || cleanup.any?
```

MikeMcQuaid:
Thanks @colindean!

PR #1115, opened 2022-08-19
Gracefully start services that aren't running

jeremy:
When `start_service: true`, `install` starts the service if it isn't running, whether or not it was installed.

Aligns with `check` which expects a formula with `start_service: true` to have a running service.

This is possible with `restart_service: true`, but that *always* restarts the service which is needlessly disruptive when it's already running and hasn't been upgraded.

Restarts can be avoided with `brew bundle check || brew bundle install` in tandem with `restart_service: true`, but that's drifting from "Just Works" to fairly arcane.

`brew bundle dump` is updated to reflect graceful service changes. They're now dumped with `start_service: true, restart_service: :changed` instead of `restart_service: true` so `brew bundle install` will just keep the service running and up-to-date.

Ideally, this would be the default for any formulae with a service: they'd be started and updated by default with no additional config. Formulae with optional services would use `start_service: false` to turn them off. This updates the dumper to anticipate such a change and dump formulae with services that aren't running as `start_service: false`.

In the meantime, `restart_service: :changed` could imply `start_service: true`, matching our intuition that the service should be running.

References #188, #204

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ cask_args appdir: "~/Applications", require_sha: true
 
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
-# 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
+# 'brew install --with-rmtp', always 'brew services restart'
+brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: true
+# 'brew install', always 'brew service start' if not running, 'brew services restart' on version changes, 'brew link', 'brew unlink mysql' (if it is installed)
+brew "mysql@5.6", start_service: true, restart_service: :changed, link: true, conflicts_with: ["mysql"]
 
 # 'brew install --cask'
 cask "google-chrome"
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ cask_args appdir: "~/Applications", require_sha: true
 
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', always 'brew services restart'
-brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: true
-# 'brew install', always 'brew service start' if not running, 'brew services restart' on version changes, 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", start_service: true, restart_service: :changed, link: true, conflicts_with: ["mysql"]
+# 'brew install --with-rmtp', 'brew services restart' on version changes
+brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
+# 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
+brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
 
 # 'brew install --cask'
 cask "google-chrome"
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -75,6 +75,8 @@ module Homebrew
       switch "--describe",
              description: "`dump` adds a description comment above each line, unless the " \
                           "dependency does not have a description."
+      switch "--no-start",
+             description: "`dump` does not add `start_service` to formula lines."
       switch "--no-restart",
              description: "`dump` does not add `restart_service` to formula lines."
       switch "--zap",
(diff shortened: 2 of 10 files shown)


jacobbednarz (approved):
selfishly approving this one to remove a handful of instances where i'm using `brew bundle check || brew bundle install` 😄 .

README.md
```diff
@@ -35,10 +35,10 @@ cask_args appdir: "~/Applications", require_sha: true
...
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
-# 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
+# 'brew install --with-rmtp', always 'brew services restart'
+brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: true
+# 'brew install', always 'brew service start' if not running, 'brew services restart' on version changes, 'brew link', 'brew unlink mysql' (if it is installed)
+brew "mysql@5.6", start_service: true, restart_service: :changed, link: true, conflicts_with: ["mysql"]
```
MikeMcQuaid, line 41:
```suggestion
brew "mysql@5.6", ensure_started: true, restart_service: :changed, link: true, conflicts_with: ["mysql"]
```

What do you think about this naming? I'm not sure I fully understand if that's what it's doing. I'm also ensure if we're best to add another e.g. `start_service` or add a new state to `restart_service`. Thoughts?

MikeMcQuaid (commented):
Thanks for the PR! Looking good so far, just want to chat about the DSL.

README.md
jeremy: Note that this PR uses the existing `start_service` option (it's not a new one!) anticipating a softer landing for an incremental improvement that doesn't break compatibility.

I'd love to see it motivate a DSL change as well, reflecting that common Do What I Mean case of "hey, this Brewfile runs these services, please." In my own exposure to Homebrew bundling scenarios, running the service is so commonly expected that I'd encourage it as a default that needn't be stated at all: the service is started on install and restarted on upgrade. To disable, pass `service: false` to indicate that service management is ignored entirely.

Better yet, move service declarations to a distinct DSL _and_ commandset that echoes the existing `brew services …` suite. Let software deps be software deps only. Declare services with e.g. `services "mysql@5.6", "redis"`. Invoke `brew bundle services …` to act like `brew services … --all` but scoped to the services declared in the Brewfile. Treat `brew bundle install` as implying `brew bundle services start`, with `brew bundle --no-services` to override.

DSL suggestions aside, I think this PR is a sensible step toward declarative services management, demonstrating "Do What I Mean" behavior that such a DSL could be hewn to suit.

MikeMcQuaid: > Note that this PR uses the existing start_service option (it's not a new one!) anticipating a softer landing for an incremental improvement that doesn't break compatibility.

Yup, it just reads weirdly to me to have to specify both like this.

> the service is started on install and restarted on upgrade. T

I don't agree with this because many services are installed e.g. as dependencies but never need to be run.

> I'd love to see it motivate a DSL change as well, reflecting that common Do What I Mean case of "hey, this Brewfile runs these services, please."

I've always used and had users of my `Brewfile`s use `restart_service` for this and this PR is the first I've heard complaints about the service being unnecessarily restarted.

> DSL suggestions aside, I think this PR is a sensible step toward declarative services management, demonstrating "Do What I Mean" behavior that such a DSL could be hewn to suit.

As-is: I'm afraid I disagree. With some iteration/naming changes: quite possibly 👍🏻

jeremy: * [Reverted](https://github.com/Homebrew/homebrew-bundle/pull/1115/commits/f7b441d64a103d18ccbc74d2458118fb058d261b) the dumper changes that introduced confusing dual usage of `start_service` alongside `restart_service`.
* [Proposed](https://github.com/Homebrew/homebrew-bundle/pull/1115/commits/018bf7130dc463f4da0644b67d9a3bbd852e683d) having `restart_service: :changed` also start services that aren't running.

MikeMcQuaid: > In this usage, I have yet to encounter a declared dependency that does not need to run.

This definitely happens. Some software that provides a service is needed as a build-time dependency of non-Homebrew software.

> Apologies; this PR is not a complaint. I appreciate and extensively rely on Homebrew Bundle.

No worries! Didn't take it as one ❤️ 

> That'd achieve the same outcome, that install will start a service that states restart_service: :changed and check will pass afterward.

That would work for me 👍🏻

jeremy: > > That'd achieve the same outcome, that install will start a service that states restart_service: :changed and check will pass afterward.
> That would work for me 👍🏻

(This is done, btw! PR now boils down to `restart_service: :changed` implying `start_service: true`.)

MikeMcQuaid (approved):
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @jeremy!

jeremy:
Argh - broke actual usage due to stubbing `BrewServices.start` but not adding the implementation. 🤦🏼‍♂️

Fixed in #1129

PR #1116, opened 2022-08-19
Support link `--overwrite` with `brew "formula", link: :overwrite`

### LINKED ISSUE #1062: feature request: support brew link --overwrite
To handle cases like:
```
Using kubernetes-cli
Linking kubernetes-cli formula.
Linking /usr/local/Cellar/kubernetes-cli/1.23.4...
Error: Could not symlink bin/kubectl
Target /usr/local/bin/kubectl
already exists. You may want to remove it:
  rm '/usr/local/bin/kubectl'

To force the link and overwrite all conflicting files:
  brew link --overwrite kubernetes-cli

To list all files that would be deleted:
  brew link --overwrite --dry-run kubernetes-cli
```

jeremy:
Implies `--force` for uniform usage with keg-only formulae.

Closes #1062.

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -92,6 +92,9 @@ module Bundle
 
     def link_change_state!(verbose: false)
       case @link
+      when :overwrite
+        puts "Overwriting links for #{@name} formula." if verbose
+        Bundle.system(HOMEBREW_BREW_FILE, "link", "--force", "--overwrite", @name, verbose: verbose)
       when true
         unless linked_and_keg_only?
           puts "Force-linking #{@name} formula." if verbose
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ tap "user/tap-repo", "https://user@bitbucket.org/user/homebrew-tap-repo.git", fo
 # set arguments for all 'brew install --cask' commands
 cask_args appdir: "~/Applications", require_sha: true
 
-# 'brew install'
-brew "imagemagick"
+# 'brew install', 'brew link'
+brew "imagemagick@6", link: true
 # 'brew install --with-rmtp', 'brew services restart' on version changes
 brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
-# 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
+# 'brew install', always 'brew services restart', 'brew link --overwrite', 'brew unlink mysql' (if it is installed)
+brew "mysql@5.6", restart_service: true, link: :overwrite, conflicts_with: ["mysql"]
 
 # 'brew install --cask'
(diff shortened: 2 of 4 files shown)


jacobbednarz (approved):
this seems reasonable to me given the previous discussion on it. @MikeMcQuaid do you have any reservations about the API here?

lib/bundle/brew_installer.rb
```diff
@@ -92,6 +92,9 @@ def service_change_state!(verbose:)
 
     def link_change_state!(verbose: false)
       case @link
+      when :overwrite
+        puts "Overwriting links for #{@name} formula." if verbose
+        Bundle.system(HOMEBREW_BREW_FILE, "link", "--force", "--overwrite", @name, verbose: verbose)
```
MikeMcQuaid, line 97: Are you sure `--force` is needed here? That should only be needed if it's keg-only.

MikeMcQuaid (commented):
I'm 👍🏻 with this being done. I do wonder if we want a similar DSL/status for casks to be consistent there?

lib/bundle/brew_installer.rb
jeremy: Right - it's meant to support overwriting links with a keg-only formula. Given that `--force` is effectively a no-op for non-key-only formulae, I figured it's clearer to use a single invocation that works for any formula.

MikeMcQuaid: @jeremy I think `--force` should only be passed when it's keg-only, we have that information already in memory and it feels like a cleaner implementation.

jeremy: That state ("keg-only and unlinked") is not presently available, but could be with changes to BrewDumper.

BrewDumper.formulae sets `link?` to one of three values which capture four distinct states:
| dumped | logic | keg_only | linked |
|---|---|---|---|
| `link?: true` | keg_only? && linked? | true | true |
| `link?: false` | !keg_only? && !linked? | false | false |
| `link?: nil` | keg_only? && !linked? | true | false |
| `link?: nil` | !keg_only? && linked? | false | true |

We could have BrewDumper use `link?: :overwrite` to capture the `keg_only? && !linked?` state, or we could split the compound `link?` state into `linked?` and `keg_only?` booleans.

MikeMcQuaid: > We could have BrewDumper use link?: :overwrite to capture the keg_only? && !linked? state, or we could split the compound link? state into linked? and keg_only? booleans.

Fine with either option 👍🏻

swrobel:
Would be sweet to get this merged. It's the one thing holding our team back from using a Brewfile

MikeMcQuaid:
@swrobel Note sure why the 😕 response here. You commented on a closed PR where I've given feedback on what needs to be changed. This is open source: we're all volunteers here and if this is blocking your team: you or your team can do the remaining work to get this over the line/

swrobel:
Fair enough. What I saw last night was that it had been approved by @jacobbednarz, yet left unmerged. I didn't realize changes were requested.

PR #1117, opened 2022-08-19
Let users choose to install things even if no bottle exists

johnewart:
I have seen a few previous comments somewhat related to this but I think that users should be able to forcibly install things even if no bottle exists. An environment variable seemed like the least amount of surgery to accomplish this (and in-line with other toggles in general) but if there is a different, preferred, mechanism then let me know. 

For context, I have a setup where changing my `Brewfile` causes `brew bundle` to be run after my configs get updated (using `chezmoi`) and I would like to be able to have the `Brewfile` be a single source of truth because then I am less likely to install packages ad-hoc. Without the ability to override this skip behavior I can't install some packages this way (things like luajit, neovim and tree-sitter will not install via `Brewfile` without a way to override the no-bottle behavior.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -18,6 +18,8 @@ module Homebrew
 
         You can skip the installation of dependencies by adding space-separated values to one or more of the following environment variables: `HOMEBREW_BUNDLE_BREW_SKIP`, `HOMEBREW_BUNDLE_CASK_SKIP`, `HOMEBREW_BUNDLE_MAS_SKIP`, `HOMEBREW_BUNDLE_WHALEBREW_SKIP`, `HOMEBREW_BUNDLE_TAP_SKIP`.
 
+	You can force installation of a formula that does not have a bottle by setting the `HOMEBREW_INSTALL_IF_BOTTLE_MISSING` environment variable to `true`.
+
         `brew bundle` will output a `Brewfile.lock.json` in the same directory as the `Brewfile` if all dependencies are installed successfully. This contains dependency and system status information which can be useful in debugging `brew bundle` failures and replicating a "last known good build" state. You can opt-out of this behaviour by setting the `HOMEBREW_BUNDLE_NO_LOCK` environment variable or passing the `--no-lock` option. You may wish to check this file into the same version control system as your `Brewfile` (or ensure your version control system ignores it if you'd prefer to rely on debugging information from a local machine).
 
         `brew bundle dump`:
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -18,7 +18,7 @@ module Homebrew
 
         You can skip the installation of dependencies by adding space-separated values to one or more of the following environment variables: `HOMEBREW_BUNDLE_BREW_SKIP`, `HOMEBREW_BUNDLE_CASK_SKIP`, `HOMEBREW_BUNDLE_MAS_SKIP`, `HOMEBREW_BUNDLE_WHALEBREW_SKIP`, `HOMEBREW_BUNDLE_TAP_SKIP`.
 
-	You can force installation of a formula that does not have a bottle by setting the `HOMEBREW_INSTALL_IF_BOTTLE_MISSING` environment variable to `true`.
+        You can force installation of a formula that does not have a bottle by setting the `HOMEBREW_INSTALL_IF_BOTTLE_MISSING` environment variable to `true`.
 
         `brew bundle` will output a `Brewfile.lock.json` in the same directory as the `Brewfile` if all dependencies are installed successfully. This contains dependency and system status information which can be useful in debugging `brew bundle` failures and replicating a "last known good build" state. You can opt-out of this behaviour by setting the `HOMEBREW_BUNDLE_NO_LOCK` environment variable or passing the `--no-lock` option. You may wish to check this file into the same version control system as your `Brewfile` (or ensure your version control system ignores it if you'd prefer to rely on debugging information from a local machine).
 
--- a/lib/bundle/skipper.rb
+++ b/lib/bundle/skipper.rb
@@ -18,8 +18,9 @@ module Bundle
 
           # If the user wants to install things from a bundle even if no bottle exists
           # we should let them.
-          if ENV['HOMEBREW_INSTALL_IF_BOTTLE_MISSING'] == "true"
-            puts Formatter.warning "... but HOMEBREW_INSTALL_IF_BOTTLE_MISSING is set to true. Installing anyway!" unless silent
+          if ENV["HOMEBREW_INSTALL_IF_BOTTLE_MISSING"] == "true"
+            message = "... but HOMEBREW_INSTALL_IF_BOTTLE_MISSING is set to true. Installing anyway!"
+            puts Formatter.warning message unless silent
             return false
           end
 
--- a/lib/bundle/skipper.rb
+++ b/lib/bundle/skipper.rb
@@ -13,7 +13,16 @@ module Bundle
            formula[:official_tap] &&
            !formula[:bottled]
           reason = Hardware::CPU.arm? ? "Apple Silicon" : "Linux"
+
           puts Formatter.warning "Skipping #{entry.name} (no bottle for #{reason})" unless silent
+
+          # If the user wants to install things from a bundle even if no bottle exists
+          # we should let them.
+          if ENV['HOMEBREW_INSTALL_IF_BOTTLE_MISSING'] == "true"
+            puts Formatter.warning "... but HOMEBREW_INSTALL_IF_BOTTLE_MISSING is set to true. Installing anyway!" unless silent
+            return false
+          end
+
           return true
         end
 
--- a/spec/bundle/skipper_spec.rb
+++ b/spec/bundle/skipper_spec.rb
@@ -33,6 +33,14 @@ describe Bundle::Skipper do
 
         expect(skipper.skip?(entry)).to be true
       end
+      it "returns false when HOMEBREW_INSTALL_IF_BOTTLE_MISSING is true" do
+        allow(ENV).to receive(:[]).with("HOMEBREW_INSTALL_IF_BOTTLE_MISSING").and_return("true")
+        allow(Hardware::CPU).to receive(:arm?).and_return(true)
+        allow_any_instance_of(Formula).to receive(:bottled?).and_return(false)
+
+        expect(skipper.skip?(entry)).to be false 
+      end
+
     end
 
     context "with an unlisted cask", :needs_macos do


cmd/bundle.rb
```diff
@@ -18,6 +18,8 @@ def bundle_args
 
         You can skip the installation of dependencies by adding space-separated values to one or more of the following environment variables: `HOMEBREW_BUNDLE_BREW_SKIP`, `HOMEBREW_BUNDLE_CASK_SKIP`, `HOMEBREW_BUNDLE_MAS_SKIP`, `HOMEBREW_BUNDLE_WHALEBREW_SKIP`, `HOMEBREW_BUNDLE_TAP_SKIP`.
 
+        You can force installation of a formula that does not have a bottle by setting the `HOMEBREW_INSTALL_IF_BOTTLE_MISSING` environment variable to `true`.
```
colindean, line 21:
```suggestion
        You can force the installation of a formula that does not have a bottle by setting the `HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING` environment variable.
```

lib/bundle/skipper.rb
```diff
@@ -13,7 +13,17 @@ def skip?(entry, silent: false)
...
           puts Formatter.warning "Skipping #{entry.name} (no bottle for #{reason})" unless silent
+
+          # If the user wants to install things from a bundle even if no bottle exists
+          # we should let them.
+          if ENV["HOMEBREW_INSTALL_IF_BOTTLE_MISSING"] == "true"
+            message = "... but HOMEBREW_INSTALL_IF_BOTTLE_MISSING is set to true. Installing anyway!"
+            puts Formatter.warning message unless silent
+            return false
+          end
+
```
colindean, line 26:
```suggestion
          if ENV["HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING"]
            message = "...but HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING is set, so installing from source."
            puts Formatter.warning message unless silent
            return false
          end

```

spec/bundle/skipper_spec.rb
```diff
@@ -33,6 +33,14 @@
 
         expect(skipper.skip?(entry)).to be true
       end
+      it "returns false when HOMEBREW_INSTALL_IF_BOTTLE_MISSING is true" do
+        allow(ENV).to receive(:[]).with("HOMEBREW_INSTALL_IF_BOTTLE_MISSING").and_return("true")
```
colindean, line 37:
```suggestion
      it "returns false when HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING is true" do
        allow(ENV).to receive(:[]).with("HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING").and_return("true")
```

colindean (approved):
I was initially "meh" on this but having literally hit this problem 10 minutes ago with a new Homebrew on Linux installation where something in my Brewfile doesn't have a bottle for Linux, I think I'd like for this functionality to exist.

However, IIRC, Homebrew strongly steers folks away from building from source these days so I don't know if enabling this at all is feasible while maintaining that posture. I'll leave that up to the other maintainers to decide.

lib/bundle/skipper.rb
Rylan12: How about simplifying this so that it's just a single warning message either way. Something like

```suggestion
          if ENV["HOMEBREW_BUNDLE_INSTALL_IF_BOTTLE_MISSING"]
            puts Formatter.warning "Installing #{entry.name} from source" unless silent
            return false
          else
            puts Formatter.warning "Skipping #{entry.name} (no bottle for #{reason})" unless silent
          end

```

Rylan12 (commented):
I think it's a good idea to have `brew bundle` operate the same as `brew install`. In the case of installing something without a bottle, `brew install` halts with this message:

```console
$ brew install safe-rm
Error: safe-rm: no bottle available!
You can try to install from source with:
  brew install --build-from-source safe-rm
Please note building from source is unsupported. You will encounter build
failures with some formulae. If you experience any issues please create pull
requests instead of asking for help on Homebrew's GitHub, Twitter or any other
official channels.
```

I'd recommend we use the same approach here for consistency and clarity. That would be something like a `--allow-installing-from-source` option (maybe with a shorter name) added to `brew bundle` that would have this functionality (but only for formulae that can't be built from a bottle).

Then, if `brew bundle install` encounter formulae without bottles, it can print a message saying "some formulae weren't able to be installed because they were missing bottles. To install them from source, run `brew bundle install --allow-installing-from-source`"

cmd/bundle.rb
MikeMcQuaid: Agreed, if we do want an environment variable it should be a `HOMEBREW_BUNDLE` one.

MikeMcQuaid (commented):
Thanks for the PR!

Agreed with @Rylan12 that I'm unsure about a "set it and forget it" environment variable here when it's effectively an unsupported configuration and we don't support this in Homebrew/brew for this very reason.

Can you elaborate a bit more on what formulae you're wanting to install here and why they don't have a bottle? Are they in homebrew/core or elsewhere? Are you using a non-default prefix? Are you on Linux or macOS ARM or macOS Intel?

johnewart:
> Thanks for the PR!
> 
> Agreed with @Rylan12 that I'm unsure about a "set it and forget it" environment variable here when it's effectively an unsupported configuration and we don't support this in Homebrew/brew for this very reason.
> 
> Can you elaborate a bit more on what formulae you're wanting to install here and why they don't have a bottle? Are they in homebrew/core or elsewhere? Are you using a non-default prefix? Are you on Linux or macOS ARM or macOS Intel?

In this case, I am on macOS ARM (speaking of which the arch / CPU code seems like it would complain about macOS M1 on ARM linux but that is an aside.) and there are at least three - neovim, tree-sitter and luajit (I think, now that I patched it locally I don't see the complaints anymore but I can undo it and double-check 😆 )

As for why they don't have a bottle, I don't know enough about the bottle universe, but I think perhaps it is because I am installing explicitly from HEAD (via `args: ['HEAD']`). 

If a command-line argument like `--enable-install-from-source` (or the like) is better, I can change it to use that mechanism.

Also I can't, for the life of me, seem to reproduce the linter (running `brew style` in the source tree seems to check a bunch of files but doesn't complain, I tried emulating what the GitHub action is doing in various ways to no avail) any idea how I can do it locally to avoid ye olde "push-and-pray" technique? 😄

johnewart:
> I think it's a good idea to have `brew bundle` operate the same as `brew install`. In the case of installing something without a bottle, `brew install` halts with this message:
>
> I'd recommend we use the same approach here for consistency and clarity. That would be something like a `--allow-installing-from-source` option (maybe with a shorter name) added to `brew bundle` that would have this functionality (but only for formulae that can't be built from a bottle).
> 

Seems reasonable to have them operate consistently / have the same, or at least very similar, command-line flags; I think the subtle difference in this case is that (as mentioned above) I am installing from HEAD explicitly and thereby installing from source. I would have imagined in this case that it would always install from source because I specified `--HEAD` and `brew install --HEAD` works fine manually, which is different behavior than the bundler. 

In thinking more about it, I guess another way to solve this would be to bypass the skipper if the spec includes the `--HEAD` argument. Though, in general, I would like to be able to install from source where bottles aren't available even if not building from HEAD if I tell it to do so.

Rylan12:
> if _all_ formulae in the official taps should have a bottle, and the bundle installer should only refuse to install in the absence of a bottle, then it should be very strange and exceptionally rare to get into a situation where a formula bottle can't be installed.

My understanding is that the goal is for all formulae to have bottles. At the moment, it looks like only [one formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/abi-dumper.rb) doesn't have a bottle, although some formulae are not bottled on all supported OSes/arches. However, `neovim`, `tree-sitter` and `luajit` all have bottles on all supported configurations.

> As for why they don't have a bottle, I don't know enough about the bottle universe, but I think perhaps it is because I am installing explicitly from HEAD (via `args: ['HEAD']`).

Yes, passing `--HEAD` does mean that a bottle won't be used and it will be built from source, so I think this behavior is expected. To clarify: is this PR intended to fix a bug you were encountering or is it simply a new feature?

If it's just a feature request, I think I'm leaning toward letting this be accomplished by just using `args: ["build-from-source"]`. I think that if a user chooses to run in an unsupported configuration that doesn't allow them to use bottles, it's up to them to manually adjust their `Brewfile` to accommodate for this.

I'd potentially be okay with adding a `--build-from-source` flag to the `bundle` command that would just add `--build-from-source` to the args passed to each install. Then, a user could run `brew bundle install` followed by `brew bundle install --build-from-source` to first install everything that has bottles with the bottles, and then on the second pass install everything else from source.

> Also I can't, for the life of me, seem to reproduce the linter (running `brew style` in the source tree seems to check a bunch of files but doesn't complain, I tried emulating what the GitHub action is doing in various ways to no avail) any idea how I can do it locally to avoid ye olde "push-and-pray" technique? 😄

`brew style homebrew/bundle` does the trick for me

johnewart:
> Yes, passing `--HEAD` does mean that a bottle won't be used and it will be built from source, so I think this behavior is expected. To clarify: is this PR intended to fix a bug you were encountering or is it simply a new feature?

I suspect it falls into ... both? I would have expected that passing `--HEAD` via `args:` would "just work", but it didn't so this was a way to make that happen. If `bundle` is supposed to behave the same was `brew install` should, then this is not the case and it's a bug. If I am  misinterpreting the expected behavior then it's a feature request -- Schrödinger's PR? :cat:

> If it's just a feature request, I think I'm leaning toward letting this be accomplished by just using `args: ["build-from-source"]`. I think that if a user chooses to run in an unsupported configuration that doesn't allow them to use bottles, it's up to them to manually adjust their `Brewfile` to accommodate for this.

I don't know about whether or not that will cause it to be considered coming from a bottle, but if not then think the skip logic will still decide to bypass the requested formula, since the `!formula[:bottled]` check will still cause it to be skipped. In which case, won't this yield the same outcome as me passing `--HEAD`? (I haven't tried it to see, but I will) 

> I'd potentially be okay with adding a `--build-from-source` flag to the `bundle` command that would just add `--build-from-source` to the args passed to each install. Then, a user could run `brew bundle install` followed by `brew bundle install --build-from-source` to first install everything that has bottles with the bottles, and then on the second pass install everything else from source.

At the end of the day I just want it to build from source if it can't find a bottle, particularly so if I tell it to use HEAD. Will `--build-from-source` use the same version as the bottle, except build it from source rather than fetch the pre-built version? In my case, I deliberately want the latest commit for these things and to rebuild them pretty regularly. 

> > Also I can't, for the life of me, seem to reproduce the linter (running `brew style` in the source tree seems to check a 
> 
> `brew style homebrew/bundle` does the trick for me

Weird; I tried that originally as it was the command in the GH action and it did not have the same behavior as it did just now - thanks for the sanity check; it works as-expected.

Rylan12:
> I would have expected that passing `--HEAD` via `args:` would "just work", but it didn't

Can you give specific reproduction instructions? I edited my `Brewfile` to contain `brew "htop", args: ["HEAD"]`[^1], remove `htop`, and ran `brew bundle`. Everything seemed to work as expected, and the `HEAD` version of `htop` was installed.

I also tried passing only `build-from-source` to `args` and it also worked as expected.

> Will `--build-from-source` use the same version as the bottle, except build it from source rather than fetch the pre-built version?

Yes, exactly. `--HEAD` and `build-from-source` works similarly except `--HEAD` will download the source files from the URL specified in the `head` method/block while `--build-from-source` uses the source files from the `stable` block or `url` method.

[^1]: `htop` is just a formula I like to use to test since it doesn't take long to build

MikeMcQuaid:
> In thinking more about it, I guess another way to solve this would be to bypass the skipper if the spec includes the `--HEAD` argument.

Yes, this seems like the right fix here @johnewart.

johnewart:
Sure -- I just ran `brew bundle` on my Brewfile (sans environment variable), and then ran it again with the same Brewfile but including the environment variable and stuck the contents of my Brewfile plus the output in a gist: https://gist.github.com/johnewart/e36a3e523d95c384c79b5bb216b7b444

Rylan12:
Can you post your `brew config` and `brew doctor` output? I don't understand why you're seeing the `(no bottle for Apple Silicon)` message since all of those formulae _do_ have bottles

johnewart:
Sure, updated the gist with the outputs

Rylan12:
Weird. Everything looks okay. What's the output of `brew install neovim --HEAD`? Does that work? If not, how about `brew info --json neovim` and `brew ruby -e 'puts Utils::Bottles.tag'`?

johnewart:
Done - gist? more like novel 😆

Rylan12:
Ah, okay, I see what's happening. It looks like the issue is really in [`Bundle::BrewDumper.formula_to_hash`](https://github.com/Homebrew/homebrew-bundle/blob/2e38269fe756295375640628b2ef788edc42c312/lib/bundle/brew_dumper.rb#L152-L155). Since the head version of `neovim` is currently installed, the formula that `Formula.installed` returns uses the `:head` spec. When this is the case, `Formula#bottle_defined?` will always be `false` since head formulae don't use bottles.

I'm not super familiar with all of the intricacies about `brew bundle`so not 100% sure what the best way forward is. I think adding `formula.active_spec = :stable` just above the line linked above would do the trick, but may be too specific to this case and break things in general. It

MikeMcQuaid:
> I'm not super familiar with all of the intricacies about `brew bundle`so not 100% sure what the best way forward is. I think adding `formula.active_spec = :stable` just above the line linked above would do the trick, but may be too specific to this case and break things in general.

@Rylan12 Yeh, requesting the stable spec if we're querying formula information feels like the right fix here 👍🏻

PR #1135, opened 2022-11-01
run `setup-ruby` for all OS variants

jacobbednarz:
Updates CI run `setup-ruby` on all OS versions since macos-12 has 3.0 for default however, we want to target 2.6 (for now).

Swaps the unmaintained `actions/setup-ruby` for `ruby/setup-ruby` actions while I'm here.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,8 +33,7 @@ jobs:
     - run: brew test-bot --only-tap-syntax
 
     - name: Set up Ruby
-      if: matrix.os == 'ubuntu-latest'
-      uses: actions/setup-ruby@master
+      uses: ruby/setup-ruby@v1
       with:
         ruby-version: '2.6'
 
@@ -52,5 +51,5 @@ jobs:
         echo 'brew "hello"' > ./Brewfile
         brew bundle
         hello
-      
+
     - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70


jacobbednarz:
cc @MikeMcQuaid will need a hand here if we want to merge this as the expected status is looking for `macOS-latest` as a required pass.

MikeMcQuaid:
> macos-12 has 3.0 for default

Is this GitHub Actions specific? Thanks!

jacobbednarz:
yes, sorry I missed coming back and including the link. full details at https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

MikeMcQuaid:
@jacobbednarz Gotcha. Might be nice to just update the `PATH` instead so that we use `/usr/bin/ruby` (which should be still 2.6). Thoughts?

jacobbednarz:
after some ☕ i realised we already have a step that sets up ruby but only for linux - https://github.com/Homebrew/homebrew-bundle/pull/1135/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR35-R39.

let me see if i can get that working instead of manipulating `PATH` as I'm not sure if that will work here since ruby 3.x is already installed on `PATH.

jacobbednarz:
that got it. running the `setup-ruby` action for all OS's worked better.

squashed this one so we're ready for another 👀

MikeMcQuaid:
Good catch thanks @jacobbednarz!

PR #1156, opened 2023-01-20
document `no-quarantine` setting

gromgit:
The syntax is not obvious, and it's not automatically populated, so it's probably a good idea to show how to do it right.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @gromgit!

PR #1160, opened 2023-01-25
Resolve RSpec/NoExpectationExample violations

dduugg:
This is a prerequisite to https://github.com/Homebrew/brew/pull/14419
cc @MikeMcQuaid @apainintheneck

### DIFF
--- a/spec/bundle/brew_installer_spec.rb
+++ b/spec/bundle/brew_installer_spec.rb
@@ -160,6 +160,8 @@ describe Bundle::BrewInstaller do
         expect(Bundle::BrewServices).to receive(:restart).with(formula, verbose: verbose).and_return(true)
       end
 
+      # These tests wrap expect() calls in `expectations`
+      # rubocop:disable RSpec/NoExpectationExample
       it "unlinks conflicts and stops their services" do
         expectations(verbose: false)
         described_class.preinstall(formula, restart_service: true, conflicts_with: ["mysql56"])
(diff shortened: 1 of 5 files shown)


dduugg:
Oops, this is isn't a simple pre-req, bc it triggers a `Lint/UnneededDisable` without https://github.com/Homebrew/brew/pull/14419 having enabling `RSpec/NoExpectationExample`. This should eventually resolve by merging both PRs, but I can take a less risky path if you prefer.

apainintheneck:
Yeah, I think that it'll have to be handled instead of being disabled here. There are two failing tests failing Rubocop right now because they both use the `expectations` method instead of including expectation directly.

Screen Shot 2023-01-25 at 9 14 24 PM

My first suggestion to use shared examples doesn't really make any sense. But if we look at the tests we can see that it seems designed to test the verbosity of the output (even though I don't think it's really doing that right now) so we could just add an expectation for that. I haven't tested the following to see if it passes Rubocop but it does pass RSpec locally.

### Before
https://github.com/Homebrew/homebrew-bundle/blob/fa7eafa7ce6e76069fc8c4eb7fa01e7ba2c0070a/spec/bundle/brew_installer_spec.rb#L163-L174

### After
```rb
it "unlinks conflicts and stops their services" do
  expectations(verbose: false)

  expect do
    described_class.preinstall(formula, restart_service: true, conflicts_with: ["mysql56"])
    described_class.install(formula, restart_service: true, conflicts_with: ["mysql56"])
  end.not_to output.to_stdout
end

it "prints a message" do
  expectations(verbose: true)

  expect do
    described_class.preinstall(formula, restart_service: true, conflicts_with: ["mysql56"], verbose: true)
  end.to output(/^Skipping install of mysql formula/).to_stdout
  expect do
    described_class.install(formula, restart_service: true, conflicts_with: ["mysql56"], verbose: true)
  end.to output(/^Unlinking mysql56 formula/).to_stdout
end
```

I'm sure there are other ways to work around this linter error. Just an idea.

Issue #1169, opened 2023-02-17
Locking step fails with "Error: No available tap homebrew/core." if homebrew/core isn't tapped.


faisal:
Bundling fails in lockfile generation when using brew 4 with homebrew/core untapped.

Steps:

1. `brew update` to 4.0 or later.
2. `brew untap homebrew/core`
3. `brew bundle`

Expected behavior: Homebrew completes the bundling process.

Actual behavior: Homebrew goes through the bundling process but fails in locking, reporting "Error: No available tap homebrew/core."

Notes:
- It looks like Bundle::Locker#system_macos's interaction with CoreTap assumes homebrew/core is tapped.
- Running `brew bundle --no-lock` succeeds.
- Removing the `"Homebrew/homebrew-core" => CoreTap.instance.git_head,` line from Bundle::Locker#system_macos allows the bundle to proceed for this case, but presumably would do the wrong thing if homebrew/core _is_ tapped.

colindean:
I'm not able to reproduce this immediately. 

I'm on `HOMEBREW_VERSION: 4.0.1-58-g82a36d2` on Linux (Debian 11.6) and just today updated to 4.x and untapped homebrew/core. I'm able to run brew bundle with a variety of Brewfiles I have laying around.

Can you provide the output of:

* `brew config`
* `brew doctor`

colindean:
I just caught that I have lockfiles disabled on this machine. I'm able to reproduce it now.

```
unset HOMEBREW_BUNDLE_NO_LOCK; brew bundle --file=Brewfile.all                                                                                                                
Using colindean/personal
Using bash
Using bash-completion@2
Using bat
Using git-delta
Using coreutils
Using eoconv
Using fzf
Using zoxide
Error: No available tap homebrew/core.
```

colindean:
The quick workaround is to use `--no-lock` or set `export HOMEBREW_BUNDLE_NO_LOCK=1`.

PR #1171, opened 2023-02-17
Include core tap hash in lock file only if tapped

faisal:
Fixes https://github.com/Homebrew/homebrew-bundle/issues/1169

brew-bundle assumes homewbrew/core is tapped when writing the lock file. As of Homebrew 4, that is no longer a valid assumption. This PR modifies the locking behavior to only include a hash for homebrew/core if it is tapped.

n.b. I will close this if/when 1172 lands.

### DIFF
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -138,14 +138,17 @@ module Bundle
     end
 
     def system_macos
-      [MacOS.version.to_sym.to_s, {
+      details_hash = {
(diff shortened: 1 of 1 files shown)


faisal:
https://github.com/Homebrew/homebrew-bundle/pull/1172

faisal:
Issue addressed via https://github.com/Homebrew/homebrew-bundle/pull/1172. Thank you, @issyl0 !

PR #1172, opened 2023-02-18
locker: Only get `CoreTap.instance.git_head` if not using the API

issyl0:
- Fixes https://github.com/Homebrew/homebrew-bundle/issues/1170.
- Fixes https://github.com/Homebrew/homebrew-bundle/issues/1169.
- Before, the lockfile would try to populate with `CoreTap.instance.git_head` which doesn't exist when using the API (now the default), leading to this error for the user:

```
> brew bundle --file=testBrewfile
Using asdf
Error: No available tap homebrew/core.
```

Now:

```
> brew bundle --file=testBrewfile
Using asdf

❯ tail testBrewfile.lock.json
        "HOMEBREW_VERSION": "4.0.1-60-gf0bcef6-dirty",
        "HOMEBREW_PREFIX": "/opt/homebrew",
        "Homebrew/homebrew-core": "api",
        "CLT": "14.1.0.0.1.1666437224",
        "Xcode": "14.1",
        "macOS": "13.0"
      }
    }
  }
}
```

### DIFF
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -3,6 +3,7 @@
 require "tap"
 require "os"
 require "development_tools"
+require "env_config"
 
 module Bundle
   module Locker
@@ -141,7 +142,7 @@ module Bundle
       [MacOS.version.to_sym.to_s, {
         "HOMEBREW_VERSION"       => HOMEBREW_VERSION,
         "HOMEBREW_PREFIX"        => HOMEBREW_PREFIX.to_s,
-        "Homebrew/homebrew-core" => CoreTap.instance.git_head,
+        "Homebrew/homebrew-core" => Homebrew::EnvConfig.install_from_api? ? "api" : CoreTap.instance.git_head,
         "CLT"                    => MacOS::CLT.version.to_s,
         "Xcode"                  => MacOS::Xcode.version.to_s,
         "macOS"                  => MacOS.full_version.to_s,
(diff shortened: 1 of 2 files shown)


colindean (approved):
This is a fine stopgap solution to get it working again now. We can fix the versioning bit later.

lib/bundle/locker.rb
```diff
@@ -3,6 +3,7 @@
 require "tap"
 require "os"
 require "development_tools"
+require "env_config"
```
colindean, line 6: Gonna have to stub this out

issyl0: Don't need it in the end since I'm using the full `Homebrew::EnvConfig` directly.

issyl0: Nope, I was wrong, still do, oops. I will fix the tests.

jacobbednarz (approved):

colindean (approved):
I think this is OK, but something that just came to mind is this: if someone's actually versioning this lockfile, there's going to be thrashing between a git hash and `api` or eventually some kind of `api:<timestamp or hash of formula.json>`. That thrashing would already be present because of updating brew/core, though.

faisal:
> I think this is OK, but something that just came to mind is this: if someone's actually versioning this lockfile, there's going to be thrashing between a git hash and `api` or eventually some kind of `api:<timestamp or hash of formula.json>`. That thrashing would already be present because of updating brew/core, though.

Can you give an example? I'd think `api` would cut down on churn a little. I used to version the lockfile, and the amount of churn for the tap hashes got a bit maddening. Re-bundle to pick up a single formula update, get new hashes for every single tap.

If we do want to record what exactly changed then I think recording the hashes behind this particular formula.json would be what I'll want when debugging, so I can tap and check out those exact hashes.

colindean:
@Bo98 [suggested](https://github.com/Homebrew/brew/issues/14698#issuecomment-1435472353) recording the formula.json hash of each individual formulae in the Brewfile instead of a repo hash version. I'm gonna think about that for a bit.

PR #1174, opened 2023-02-18
Drops conditional usage of removed DevelopmentTools.gcc_version

### LINKED ISSUE #1173: Remove gcc_version logic from Bundle::Locker#system_linux
From Bundle::Locker#system_linux:

> `# TODO: remove once https://github.com/Homebrew/brew/pull/13577 is merged and tagged.`

[13577](https://github.com/Homebrew/brew/pull/13577) landed, so we should be able to remove the logic setting gcc_version.

colindean:
Noticed this TODO while reading through for another task.

And just noticed #1173...

Closes #1173

### DIFF
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -156,7 +156,7 @@ module Bundle
 
     def system_linux
       [OS::Linux.os_version, system.merge({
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Thanks @colindean, fixed this up for ya.

PR #1175, opened 2023-02-18
Adds a version property to the lockfile, refactor Locker#lock

colindean:
With some potential changes coming up, we should express an explicit version number on this file format.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,4 +54,4 @@ jobs:
 
     - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
       with:
-      name: homebrew-bundle-${{ runner.os }}
+        name: homebrew-bundle-${{ runner.os }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,12 +46,13 @@ jobs:
     - name: Run RSpec tests
       run: bundle exec rspec
 
+    - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+      if: ${{ success() || failure() }}
+      with:
+        name: homebrew-bundle-${{ runner.os }}
+
     - name: Test that installing works
       run: |
         echo 'brew "hello"' > ./Brewfile
         brew bundle
         hello
-
-    - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
-      with:
-        name: homebrew-bundle-${{ runner.os }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,6 +50,7 @@ jobs:
       if: ${{ success() || failure() }}
       with:
         name: homebrew-bundle-${{ runner.os }}
+        flags: rspec-${{ runner.os }}
 
     - name: Test that installing works
       run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -53,3 +53,5 @@ jobs:
         hello
 
     - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+      with:
+      name: homebrew-bundle-${{ runner.os }}
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -74,7 +74,6 @@ module Bundle
       lock = update_lock_metadata(lock: lock)
       lock = add_entries_data(entries, lock: lock)
       lock = add_system_data(lock: lock)
-      lock
     end
 
     def update_lock_metadata(lock: {})
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -37,6 +37,7 @@ module Bundle
 
       lock = JSON.parse(lockfile.read) if lockfile.exist?
       lock ||= {}
+      lock["version"] = "1"
       lock["entries"] ||= {}
       lock["system"] ||= {}
 
--- a/lib/bundle/locker.rb
+++ b/lib/bundle/locker.rb
@@ -7,13 +7,16 @@ require "env_config"
 
 module Bundle
   module Locker
+
+    LOCKFILE_FORMAT_VERSION = "1"
+
     module_function
 
     extend Gem::Deprecate
 
     # @deprecated Use resolve_lockfile_path instead
     def lockfile(global: false, file: nil)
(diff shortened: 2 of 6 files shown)


jacobbednarz (approved):
LGTM. the only thing I would say here is if we are expecting this to be a part of the file long term, let's add a test to assert it's presence and we can't accidentally lose it.

colindean:
GHA CI SimpleCov is saying I'm missing _1_ line in macOS tests but codecov seems not to be showing _which_ line. I'm doing this on Linux right now...

.github/workflows/tests.yml
```diff
@@ -53,3 +53,5 @@ jobs:
         hello
 
     - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
+      with:
+      name: homebrew-bundle-${{ runner.os }}
```
issyl0, line 57:
```suggestion
        name: homebrew-bundle-${{ runner.os }}
```

lib/bundle/locker.rb
```diff
@@ -7,10 +7,12 @@
 
 module Bundle
   module Locker
+    LOCKFILE_FORMAT_VERSION = "1"
+
     module_function
 
-    def lockfile(global: false, file: nil)
-      brew_file_path = Brewfile.path(global: global, file: file)
+    def resolve_lockfile_path(global: false, brewfile: nil)
```
MikeMcQuaid, line 14: I think the older name makes more sense with what we use elsewhere for file paths.

lib/bundle/locker.rb
```diff
@@ -20,25 +22,58 @@ def lockfile(global: false, file: nil)
...
       return false if ENV["HOMEBREW_BUNDLE_NO_LOCK"]
 
       # handle the /dev/stdin and /dev/stdout cases
-      return false if lockfile(global: global, file: file).parent.to_s == "/dev"
+      return false if path.nil? || path.parent.to_s == "/dev"
+
+      true
+    end
 
+    def write_lockfile!(lockfile, lock_data)
```
MikeMcQuaid, line 35: Personally I don't think it makes things easier to follow to refactor methods called only once into other methods.

lib/bundle/locker.rb
```diff
@@ -20,25 +22,58 @@ def lockfile(global: false, file: nil)
...
+        lockfile.unlink if lockfile.exist?
+        lockfile.write("#{json}\n")
+      rescue Errno::EPERM, Errno::EACCES, Errno::ENOTEMPTY
+        opoo "Could not write to #{lockfile}!"
+        return false
+      end
       true
     end
 
+    def read_lockfile_if_exists(path)
```
MikeMcQuaid, line 47: Same deal.

lib/bundle/locker.rb
```diff
@@ -20,25 +22,58 @@ def lockfile(global: false, file: nil)
...
+      return false unless write_lockfile?(path: lockfile, no_lock: no_lock)
 
-      lock = JSON.parse(lockfile.read) if lockfile.exist?
-      lock ||= {}
+      lock_data = build_lock_data(entries, base: read_lockfile_if_exists(lockfile))
+
+      write_lockfile!(lockfile, lock_data)
+    end
+
+    def build_lock_data(entries, base: {})
```
MikeMcQuaid, line 63: Same deal.

lib/bundle/locker.rb
```diff
@@ -20,25 +22,58 @@ def lockfile(global: false, file: nil)
...
+      lock ||= base
+      lock = update_lock_metadata(lock: lock)
+      lock = add_entries_data(entries, lock: lock)
+      add_system_data(lock: lock)
+    end
+
+    def update_lock_metadata(lock: {})
```
MikeMcQuaid, line 70: Same deal.

lib/bundle/locker.rb
```diff
@@ -138,6 +154,21 @@ def whalebrew_list
       end
     end
 
+    def add_system_data(lock: {})
```
MikeMcQuaid, line 157: Same deal.

MikeMcQuaid (commented):
> With some potential changes coming up

What changes? Let's discuss that before we make a refactor to enable it.

lib/bundle/locker.rb
colindean: It feel like it improves the readability and testability, although I ran out of time while working on this to expand the tests to cover these smaller units. I did it initially to reason better about the steps taken: I found myself adding comments to explain the steps, and so many times when I write comments, I realize they could be method calls. I also like to keep Ruby methods short. I think Rubocop's default is like 10 lines; I'm not sure what [the brew rubocop](https://github.com/Homebrew/brew/blob/master/Library/.rubocop.yml) ends up resolving to.

lib/bundle/locker.rb
```diff
@@ -20,25 +22,58 @@ def lockfile(global: false, file: nil)
...
+      lock = add_entries_data(entries, lock: lock)
+      add_system_data(lock: lock)
+    end
+
+    def update_lock_metadata(lock: {})
+      lock["version"] = LOCKFILE_FORMAT_VERSION
+      lock
+    end
+
+    def add_entries_data(entries, lock: {})
```
colindean, line 75: On brevity, I fought the urge to refactor this, and then wound up running out of time, so that was a worthwhile effort dodged.

lib/bundle/locker.rb
colindean: I want the caller to understand that there's a resolution process behind this method based on its parameters because of the call to `Brewfile#path`.

colindean:
> What changes? Let's discuss that before we make a refactor to enable it.

@issyl0 beat me to a minimum-viable change in #1172. My intent was this:

* version the lockfile with `'version': '1'`
* do basically what Issy did as a stopgap, because we don't really have rules without a version scheme, haha
* act @Bo98's suggestion to include in each entity listed the entity's tap_git_head from the API data or from the tap info
* increment the version number `'version': '2'`

If this is unnecessary, then let's close this... but perhaps there's some value in the test helpers I added.

lib/bundle/locker.rb
MikeMcQuaid: @colindean this sort of refactoring is better done in another PR where it's made consistent with all naming across homebrew/bundle. Inconsistency across the repository is worse than having the name here be suboptimal, IMO.

MikeMcQuaid:
> If this is unnecessary, then let's close this...

Yeh, sorry, I think so. I'm still game to include some test helpers in some form if the scope can be slimmed down to just them? Happy for this PR to turn into that or it be another PR.

MikeMcQuaid:
Passing on this, sorry!

PR #1225, opened 2023-07-08
Use Formula#oldnames to silence deprecation notices

### LINKED ISSUE #1223: bundle dump spams warning messages to console regarding a deprecated function call
When running `$ brew bundle dump --force` I get this error message. Which suggests I report it here.

```
Warning: Calling Formula#oldname is deprecated! Use Formula#oldnames instead.
Please report this issue to the homebrew/bundle tap (not Homebrew/brew or Homebrew/homebrew-core), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/homebrew/homebrew-bundle/lib/bundle/brew_dumper.rb:160
```

colindean:
This doesn't really fully safely use `Formula#oldnames`, just
effectively uses what `Formula#oldname` does now.

https://github.com/Homebrew/brew/blob/4bbc0af6935816c75e0057364a84fd7fa8de071b/Library/Homebrew/formula.rb#L502-L515

Fixes #1223


colindean:
Test failure looks like a temporary failure to retrieve the latest https://formulae.brew.sh/api/cask.jws.json.

spec/bundle/brew_dumper_spec.rb
```diff
@@ -14,6 +14,7 @@
                     name:                   "foo",
                     desc:                   "foobar",
                     oldname:                "oldfoo",
+                    oldnames:               ["oldfoo"],
```
vraravam, line 17: Shouldn't the previous line also be deleted in this case?

spec/bundle/brew_dumper_spec.rb
```diff
@@ -58,6 +59,7 @@
                     name:                   "bar",
                     desc:                   "barfoo",
                     oldname:                nil,
+                    oldnames:               [],
```
vraravam, line 62: Shouldn't the previous line also be deleted in this case?

spec/bundle/brew_dumper_spec.rb
```diff
@@ -119,6 +121,7 @@
                     name:                   "baz",
                     desc:                   "",
                     oldname:                nil,
+                    oldnames:               [],
```
vraravam, line 124: Shouldn't the previous line also be deleted in this case?

spec/bundle/brew_dumper_spec.rb
colindean: It could, but the method is still present on the class this stub stubs.

jacobbednarz (approved):

MikeMcQuaid:
Thanks for workaround here @colindean and for merging @jacobbednarz!

PR #1248, opened 2023-09-17
Adds brief at_exit tutorial to README

colindean:
This has come up at least thrice now in filed issues. I've also been using it without a problem in some Brewfiles, so let's document it.

My _sole_ hesitation here is the level of support for reaching into Homebrew and HB Bundle internals like this. I'd appreciate thoughts and guidance in that area.

A possible extension of this could be the extension of the Brewfile DSL to provide some convenience methods for common tasks, e.g.

```ruby
brew 'cowsay'

post_install do
  formula 'cowsay' do |f|
    system "#{f.bin}/cowsay 'All done!'"
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,33 @@ Disable generation of the `Brewfile.lock.json` file by setting the environment v
 
 Tests can be run with `bundle install && bundle exec rspec`.
 
+## Running actions after Homebrew Bundle runs
+
+If you need to run commands after Homebrew Bundle runs, such as using a package that was just installed,
+you can set an [`at_exit`](https://ruby-doc.org/core-2.6.8/Kernel.html#method-i-at_exit) hook and
+use parts of Homebrew Bundle and Homebrew itself to determine paths and execute the commands programmatically.
+Note that errors within an `at_exit` block may have unintended behavior, so program safely.
+It's better to resolve the paths than to hardcode them.
+
+In this example `Brewfile`, after installing the [cowsay](https://formulae.brew.sh/formula/cowsay) package,
+Homebrew Bundle, or, really, Ruby, will execute the program with a friendly message of completion.
+
+```ruby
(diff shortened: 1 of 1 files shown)


jacobbednarz:
👍 to this simple documentation example however, I don't really want to be encouraging folks to do things post install for specific formulas. to me, that really feels like it is outside our realm and should instead exist in the formula itself and if it doesn't belong there, folks can use the `at_exit` with their own logic.

colindean:
Oh yes, things in at_exit should not be things that would be better off in a formula post-install. Rather, it might be the installation of some pip packages after having installed python, or a user writing a standard configuration file if one doesn't exist for some tool.

MikeMcQuaid (changes requested):
I appreciate the enthusiasm here but a few thoughts:
- this feels pretty heavy for the README. Perhaps better to have a dedicated "Tips and Tricks" page where we can just brain dump anything like this
- Really don't like digging into internals here. If we want to support doing things after a formula is installed: we should add a dedicated `after: "some_cmd"` (perhaps also with `before:`) so that hooks can be added in the existing DSL.

colindean:
Moving this out of the README makes sense. Would you prefer a top-level file or a docs directory? There are probably some other things suitable for docs, methinks. At the start of this section of the new tips doc, I can put something about "reaching into the internals might break" as a warning until we would do something more official.

colindean:
> Honestly I'd rather we just didn't document anything like that.

Are you OK with documenting what I've proposed here, then? They're internals but they're parts that are really unlikely ever to change.

MikeMcQuaid:
@colindean Would rather not, sorry; even if they aren't going to change, they are implementation details I don't want to encourage people to rely on.

colindean:
All right, I will close this then and we can document it when someone gets around to implementing it as a properly supported feature.

Issue #1202, opened 2023-05-02
Feature Request: Respect `--quiet`/`-q` cli flag


maxcbc:
**Problem**
Homebrew bundle has quite verbose output at times, such as the `Using package_x` output ([source](https://github.com/Homebrew/homebrew-bundle/blob/6b01bd338af95db59f1d6db87ebf6416d7518297/lib/bundle/installer.rb#LL43C1-L43C1)) when running `brew bundle install`.

`brew bundle --help` lists the standard homebrew `--quiet`/`-q` flag in its output, which is intended for use cases where the user wishes to suppress extraneous output. Currently homebrew bundle does not appear to respect this flag.

**My Usecase**
I use homebrew bundle as part as my standard dotfiles setup to maintain a consistent development setup across multiple macos devices and to minimise disaster recovery time.
As part of this, I run `brew install --global --quiet --cleanup` on every call to `brew` which would potentially change the state of my brewfile.
I'd like the `--quiet` flag to reduce the output from this command, so that it only tells me if it is installing/uninstalling packages/casks from my system.

**Solution**
Update homebrew bundle to respect the `--quiet`/`-q` flag by replacing relevant direct calls to `puts` with a custom log function which acts as a noop if the flag is active.

MikeMcQuaid:
Thanks for the good write-up. What would you expect the output to be with `--quiet`? Nothing unless there are errors?

maxcbc:
I think it is best to mirror the behaviour of the `brew` command, which I think would mean "Nothing unless there are errors" as you say.

Though, that aside, just as a cli user I'd expect:
- `quiet`: log errors or prompts for input only (i.e. if something goes wrong or the user needs to elevate to sudo etc)
- `default`: same as `quiet`, plus logging changes to the system (i.e. things being installed or removed).
- `verbose`: same as `default`, plus logging where changes to the system aren't required (i.e. where something is already installed, etc).

MikeMcQuaid:
> * `quiet`: log errors or prompts for input only (i.e. if something goes wrong or the user needs to elevate to sudo etc)

Agreed. I think the `default` and `verbose` already behaves as described so I think that'd be good scope for this issue.

@maxcbc Do you want to give this a go implementing it yourself? Happy to provide review and help on a partially working/CI failing draft PR?

maxcbc:
@MikeMcQuaid Sure. It'll probably be a few weeks before I can get to it, but a good opportunity to refresh my ruby skills.
[...]

PR #1249, opened 2023-09-17
[WIP] First pass at migrating logging puts to ohai/opoo, etc.

colindean:
Work toward #1202, but untested. I did this commit in a codespace
that is flailing.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -174,7 +174,7 @@ module Homebrew
       onoe e
       ohai(
         "#{Tty.bold}Please report this bug:#{Tty.reset}",
-        Formatter.url("https://github.com/Homebrew/homebrew-bundle/issues")
+        Formatter.url("https://github.com/Homebrew/homebrew-bundle/issues"),
       )
       ofail e.backtrace
     end
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -162,22 +162,18 @@ module Homebrew
       end
     rescue SystemExit => e
       Homebrew.failed = true unless e.success?
-      puts "Kernel.exit" if args.debug?
+      odebug "Kernel.exit" if args.debug?
     rescue Interrupt
-      puts # seemingly a newline is typical
-      Homebrew.failed = true
+      ofail "Interrupted"
     rescue RuntimeError, SystemCallError => e
       raise if e.message.empty?
 
       onoe e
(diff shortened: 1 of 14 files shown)


cmd/bundle.rb
```diff
@@ -162,22 +162,21 @@ def bundle
       end
     rescue SystemExit => e
       Homebrew.failed = true unless e.success?
-      puts "Kernel.exit" if args.debug?
+      odebug "Kernel.exit" if args.debug?
```
MikeMcQuaid, line 165:
```suggestion
      odebug "Kernel.exit"
```

may work as expected, worth a try.

cmd/bundle.rb
```diff
@@ -162,22 +162,21 @@ def bundle
...
     rescue Interrupt
-      puts # seemingly a newline is typical
-      Homebrew.failed = true
+      ofail "Interrupted"
```
MikeMcQuaid, line 167: Don't do this

cmd/bundle.rb
```diff
@@ -162,22 +162,21 @@ def bundle
...
     rescue RuntimeError, SystemCallError => e
       raise if e.message.empty?
 
       onoe e
-      puts e.backtrace if args.debug?
-      Homebrew.failed = true
+      ofail e.backtrace if args.debug?
```
MikeMcQuaid, line 172: Don't want to do this, it makes the output worse for no real reason

cmd/bundle.rb
```diff
@@ -162,22 +162,21 @@ def bundle
...
     rescue => e
       onoe e
-      puts "#{Tty.bold}Please report this bug:#{Tty.reset}"
-      puts "  #{Formatter.url("https://github.com/Homebrew/homebrew-bundle/issues")}"
-      puts e.backtrace
-      Homebrew.failed = true
+      ohai(
```
MikeMcQuaid, line 175: Makes output worse

MikeMcQuaid (changes requested):
As said in other issue: most of these changes don't make sense. `odebug` might do but all `ohai` and many `opoo` etc. use makes output worse, sorry.

colindean:
I'll revisit this less hurriedly and haphazardly, and with actual testing, in the future. Devcontainers setup should make it easier to do meaningful drive-by contributions like this.

PR #1250, opened 2023-09-20
bundle: Check for cask tap using HOMEBREW_LIBRARY

zhaofengli:
It's semantically more correct and also what's used [throughout](https://github.com/Homebrew/brew/blob/1f80e82dd72fce6642e4ea360f7a1feef5d78470/Library/Homebrew/brew.sh#L492) [brew](https://github.com/Homebrew/brew/blob/1f80e82dd72fce6642e4ea360f7a1feef5d78470/Library/Homebrew/cmd/update.sh#L581) (`$HOMEBREW_LIBRARY/Taps/*/*` for taps, `$HOMEBREW_REPOSITORY` for updating brew itself).

### DIFF
--- a/lib/bundle/bundle.rb
+++ b/lib/bundle/bundle.rb
@@ -43,7 +43,7 @@ module Bundle
 
     def cask_installed?
       @cask_installed ||= File.directory?("#{HOMEBREW_PREFIX}/Caskroom") &&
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):
Makes sense, thanks @zhaofengli!

PR #1253, opened 2023-09-27
Make `--force` take effect on install.

MikeMcQuaid:
Allows passing `brew bundle --force` to make installed formulae run `brew install --formula --force --overwrite` and casks `brew install --cask --force` when installed.

This is useful in environments where it's common for casks to need to be overwritten.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -55,7 +55,8 @@ module Homebrew
              description: "`install` won't run `brew upgrade` on outdated dependencies. " \
                           "Note they may still be upgraded by `brew install` if needed."
       switch "-f", "--force",
-             description: "`dump` overwrites an existing `Brewfile`. " \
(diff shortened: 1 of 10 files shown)


cmd/bundle.rb
```diff
@@ -77,6 +77,11 @@ def bundle_args
       switch "--describe",
              description: "`dump` adds a description comment above each line, unless the " \
                           "dependency does not have a description."
+      switch "--cask-force",
```
jacobbednarz, line 80: I wonder if this would be better as a more general option of `--force` that applies to both formulae and casks?

MikeMcQuaid: @jacobbednarz Agreed and I've moved it from being dump-time to install-time.

jacobbednarz (approved):

PR #1272, opened 2023-12-18
Run tests on Ruby 3.1

colindean:
Now that Homebrew 4.2.0 runs on Ruby 3.1, let's bump to that.

https://brew.sh/2023/12/18/homebrew-4.2.0/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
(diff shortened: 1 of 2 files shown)


colindean:
I have no idea if what I've done in fe114094f71534880a85eed88602cbfcd3f46a28 is rational but it solves the test failure resulting from rspec being unable to proxy CHILD_STATUS since it's frozen.

jacobbednarz (approved):

MikeMcQuaid:
Thanks both!

PR #1304, opened 2024-02-09
vscode_extension cleanup: downcase extension names before comparison while evaluating for cleanup

### LINKED ISSUE #1302: VSCode extensions that have capital letters seem to not be registered, and get uninstalled
I have a bunch of extensions that are managed using homebrew: https://gist.github.com/vraravam/8c9eae91a3750bed86b81e3a4711f842#file-brewfile-L193-L231

After a recent upgrade of VSCodium Insiders and homebrew, I noticed that while "brew bundle install" will install all the extensions, "brew bundle cleanup" will somehow delete only those extensions that start with capital letters!
Screenshot 2024-02-08 at 9 51 01 PM

The actual/valid extension names (all of the ones marked as "would be deleted") have capital letters somewhere in the extension-id as seen in the above link of the Brewfile, but for some reason they show up with lower-case - and I suspect this marks them for deletion. Please help fix asap.

vraravam:
fixes #1302 in a consistent manner such that cleanup doesn't uninstall if the case mismatch occurs

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -143,7 +143,7 @@ module Bundle
 
       def vscode_extensions_to_uninstall(global: false, file: nil)
         @dsl ||= Bundle::Dsl.new(Brewfile.read(global: global, file: file))
-        kept_extensions = @dsl.entries.select { |e| e.type == :vscode }.map(&:name).map(&:downcase)
+        kept_extensions = @dsl.entries.select { |e| e.type == :vscode }.map { |x| x.name.downcase }
 
         # To provide a graceful migration from `Brewfile`s that don't yet or
         # don't want to use `vscode`: don't remove any extensions if we don't
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -143,7 +143,7 @@ module Bundle
 
       def vscode_extensions_to_uninstall(global: false, file: nil)
         @dsl ||= Bundle::Dsl.new(Brewfile.read(global: global, file: file))
-        kept_extensions = @dsl.entries.select { |e| e.type == :vscode }.map(&:name)
+        kept_extensions = @dsl.entries.select { |e| e.type == :vscode }.map(&:name).map(&:downcase)
 
         # To provide a graceful migration from `Brewfile`s that don't yet or
         # don't want to use `vscode`: don't remove any extensions if we don't
@@ -151,7 +151,7 @@ module Bundle
         return [].freeze if kept_extensions.empty?
 
         current_extensions = Bundle::VscodeExtensionDumper.extensions
-        current_extensions.map(&:downcase) - kept_extensions.map(&:downcase)
+        current_extensions - kept_extensions
       end
 
       def system_output_no_stderr(cmd, *args)
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -151,7 +151,7 @@ module Bundle
         return [].freeze if kept_extensions.empty?
 
         current_extensions = Bundle::VscodeExtensionDumper.extensions
-        current_extensions - kept_extensions
+        current_extensions.map(&:downcase) - kept_extensions.map(&:downcase)
       end
 
       def system_output_no_stderr(cmd, *args)
--- a/lib/bundle/vscode_extension_dumper.rb
+++ b/lib/bundle/vscode_extension_dumper.rb
@@ -10,7 +10,7 @@ module Bundle
 
     def extensions
       @extensions ||= if Bundle.vscode_installed?
-        `code --list-extensions 2>/dev/null`.split("\n").map(&:downcase)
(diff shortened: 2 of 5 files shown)


lib/bundle/commands/cleanup.rb
```diff
@@ -151,7 +151,7 @@ def vscode_extensions_to_uninstall(global: false, file: nil)
         return [].freeze if kept_extensions.empty?
 
         current_extensions = Bundle::VscodeExtensionDumper.extensions
-        current_extensions - kept_extensions
+        current_extensions.map(&:downcase) - kept_extensions.map(&:downcase)
```
MikeMcQuaid, line 154: I think it might make more sense to add `.map(&:downcase)` to `kept_extensions` where it's originally queried above.

MikeMcQuaid:
```suggestion
        current_extensions - kept_extensions
```

I think it might make more sense to add `.map(&:downcase)` to `kept_extensions` where it's originally queried above.

lib/bundle/vscode_extension_dumper.rb
```diff
@@ -10,7 +10,7 @@ def reset!
 
     def extensions
       @extensions ||= if Bundle.vscode_installed?
-        `code --list-extensions 2>/dev/null`.split("\n").map(&:downcase)
+        `code --list-extensions 2>/dev/null`.split("\n")
```
MikeMcQuaid, line 13:
```suggestion
        `code --list-extensions 2>/dev/null`.split("\n").map(&:downcase)
```

makes more sense to me to keep this as-is and avoid having to have callers downcase.

lib/bundle/vscode_extension_installer.rb
```diff
@@ -38,7 +38,7 @@ def install(name, preinstall: true, no_upgrade: false, verbose: false, force: fa
     end
 
     def extension_installed?(name)
-      installed_extensions.include? name
+      installed_extensions.include? name.downcase
```
MikeMcQuaid, line 41:
```suggestion
      installed_extensions.include? name
```

similarly here

spec/bundle/commands/cleanup_command_spec.rb
```diff
@@ -76,6 +76,11 @@
       allow(Bundle::VscodeExtensionDumper).to receive(:extensions).and_return(%w[z])
       expect(described_class.vscode_extensions_to_uninstall).to eql(%w[z])
     end
+
+    it "computes which VSCode extensions to uninstall irrespective of case of the extension name" do
```
MikeMcQuaid, line 80: thanks for adding the test!

spec/bundle/vscode_extension_installer_spec.rb
```diff
@@ -27,9 +27,9 @@
         allow(described_class).to receive(:installed_extensions).and_return(["foo"])
       end
 
-      it "skips" do
+      it "skips ignoring case" do
```
MikeMcQuaid, line 30: Thanks for tweaking the test! In this case: I think it'd be worth adding a new test rather than adjusting this one so we make sure both paths work.

MikeMcQuaid (commented):
Thanks @vraravam, really great work so far! A few thoughts on some tweaks here.

lib/bundle/vscode_extension_dumper.rb
vraravam: Yes, i debated this with myself - and then chose the other implementation - since i felt it's more relevant and explicit when both the arrays are downcased in the same line (causes less of context switching). but, if you feel they can be separated (I agree from a performance perspective), then i will do that change. Just need your thoughts between the 2 reasonings.

spec/bundle/vscode_extension_installer_spec.rb
vraravam: done in latest commit

lib/bundle/vscode_extension_dumper.rb
MikeMcQuaid: > but, if you feel they can be separated (I agree from a performance perspective), then i will do that change

Yes please 👍🏻 🙇🏻

lib/bundle/vscode_extension_installer.rb
vraravam: If this is not done, then the case where the Brewfile has camelCase, whereas the system call to code returns all lower-case, then this method returns `false`. I think this line can remain due to this scenario. What do you think?

vraravam: For eg, in my Brewfile, I have "BriteSnow.vscode-toggle-quotes" (which is the exact extension Id as copied from the VSCode UI. But, the system call to `code --list-extensions` returns this as all lower case.
In this scenario, this method should return true after ignoring the case, correct?

lib/bundle/commands/cleanup.rb
vraravam: done

lib/bundle/vscode_extension_installer.rb
vraravam: I have updated the test to reflect this above scenario

PR #1305, opened 2024-02-09
fix tests if the user has a HOMEBREW_BUNDLE_FILE env var setup

vraravam:
I have an env var setup to point to the global `Brewfile`. If this is set, and following the contribution guidelines for this repo, then the tests fail. To circumvent that failure, this change in tests is required.

### DIFF
--- a/spec/bundle/commands/dump_command_spec.rb
+++ b/spec/bundle/commands/dump_command_spec.rb
@@ -28,6 +28,7 @@ describe Bundle::Commands::Dump do
 
   context "when files existed and `--force` is passed" do
     before do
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
@vraravam Thanks for the PR! I'd prefer instead if we fixed the tests that broke in this way.

vraravam:
> @vraravam Thanks for the PR! I'd prefer instead if we fixed the tests that broke in this way.

@MikeMcQuaid - fixed the tests as per your review comments.

MikeMcQuaid (approved):
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @vraravam!

PR #1308, opened 2024-02-20
mas v1.8.7 updates

MikeMcQuaid:
- Once the next version of `mas` is released: we can again rely on `mas outdated`.
- `mas_signedin?` hasn't worked for a long time and isn't coming back so just delete the code to simplify things.


MikeMcQuaid:
Will reopen this if things move forward on the `mas` side.

PR #1309, opened 2024-02-20
Don’t prefetch greedy outdated casks

jneubrand:
In some instances, calling `.outdated?(greedy: true)` causes additional network requests. (casks using `version :latest` and `sha256 :no_check` seem to trigger this)

Currently, this is done for *all installed packages* every time `brew bundle` wants to update *any* cask greedily.

This pull request changes `bundle`'s behavior to only check whether one cask is outdated at a time (for greedy updates).

### DIFF
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -14,11 +14,19 @@ module Bundle
       @cask_names ||= casks.map(&:to_s)
     end
 
-    def outdated_cask_names(greedy: false)
+    def outdated_cask_names
       return [] unless Bundle.cask_installed?
 
-      casks.select { |c| c.outdated?(greedy: greedy) }
+      casks.select { |c| c.outdated?(greedy: false) }
            .map(&:to_s)
+    end
+
+    def cask_is_outdated_using_greedy(cask)
+      return [] unless Bundle.cask_installed?
+
+      casks.select { |c| c.to_s == cask }
+           .map { |c| c.outdated?(greedy: true) }
+           .first
     end
 
     def cask_versions
--- a/lib/bundle/cask_dumper.rb
+++ b/lib/bundle/cask_dumper.rb
@@ -21,12 +21,13 @@ module Bundle
            .map(&:to_s)
     end
 
-    def cask_is_outdated_using_greedy(cask)
-      return [] unless Bundle.cask_installed?
+    def cask_is_outdated_using_greedy?(cask_name)
+      return false unless Bundle.cask_installed?
 
-      casks.select { |c| c.to_s == cask }
-           .map { |c| c.outdated?(greedy: true) }
-           .first
+      cask = casks.find { |c| c.to_s == cask_name }
+      return false if cask.nil?
+
+      cask.outdated?(greedy: true)
     end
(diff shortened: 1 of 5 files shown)


lib/bundle/cask_dumper.rb
```diff
@@ -14,13 +14,21 @@ def cask_names
...
-    def outdated_cask_names(greedy: false)
+    def outdated_cask_names
       return [] unless Bundle.cask_installed?
 
-      casks.select { |c| c.outdated?(greedy: greedy) }
+      casks.select { |c| c.outdated?(greedy: false) }
            .map(&:to_s)
     end
 
+    def cask_is_outdated_using_greedy(cask)
```
MikeMcQuaid, line 24:
```suggestion
    def cask_is_outdated_using_greedy?(cask)
```

lib/bundle/cask_dumper.rb
```diff
@@ -14,13 +14,21 @@ def cask_names
...
+      return [] unless Bundle.cask_installed?
```
MikeMcQuaid, line 25: Return `true` or `false` here

lib/bundle/cask_dumper.rb
```diff
@@ -14,13 +14,21 @@ def cask_names
...
+
+      casks.select { |c| c.to_s == cask }
```
MikeMcQuaid, line 27: Use `.find` instead of `select`ing a single cask.

MikeMcQuaid (commented):
> This pull request changes bundle's behavior to only check whether one cask is outdated at a time (for greedy updates).

Can you explain why this is better?

What network requests does this avoid?

jneubrand:
> Can you explain why this is better?
> 
> What network requests does this avoid?

For all of the following casks…

```
grep -IirE '^ +version :latest' -A 1 /usr/local/Homebrew/Library/Taps | grep 'sha256 :no_check' -B 1
```

(a few dozen nightlies from `homebrew-cask-versions`, but also many `homebrew-fonts`)

…the entire current file has to be downloaded every time to check whether the contents that would be installed are equivalent.

---

Thanks for the notes, I don't deal with ruby often 😅

MikeMcQuaid:
> …the entire current file has to be downloaded every time to check whether the contents that would be installed are equivalent.

Yeh, this is expected when using `--greedy`. What's not clear though is how your PR makes this different?

Does this mean that `brew bundle` will behave the same if your `Brewfile` lists all installed casks but if it only has e.g. one cask it'll only download that one whereas before it would download them all?

jneubrand:
> … `brew bundle` will behave the same if your `Brewfile` lists all installed casks but if it only has e.g. one cask it'll only download that one whereas before it would download them all?

Well, it won’t download a cask without the `greedy: true` flag anymore. If you only set that flag on *some* casks, this PR speeds `brew bundle`.

MikeMcQuaid:
> don’t list that kind of cask at all, this PR speeds `brew bundle`

If you don't use `greedy` in your cask file at all: it wasn't being used already.

So is it now that `greedy` will only check specific casks greedily rather than any cask being greedy triggering checking all casks greedily?

PR #1317, opened 2024-03-04
bundle/tap_dumper: check for installed taps.

MikeMcQuaid:
Otherwise we'll always output homebrew/core and homebrew/cask even when they aren't actually tapped.

CC @madAndroid


madAndroid:
Thank you for this! 🎉

PR #1328, opened 2024-03-28
Fix references to VSCode in extension installer test

Okeanos:
Replace references to 'apps' and 'mas' with the VSCode appropriate terms.


MikeMcQuaid (approved):
Thanks good catch here @Okeanos!

PR #1329, opened 2024-03-28
Add support for TeX Live package manager (`tlmgr`) packages

Okeanos:
Allow tlmgr packages to be installed, dumped, checked, updated and cleaned up.

---

First stab at this … not entirely sure it'll work as expected. Also there are a couple of `sudo` calls in there because a standard `basictex` cask installation uses paths that are not user writable by default.

There appear to be options to make this run in `user mode`, however, that has to be explicitly configured and should probably be handled via `args` in the Brewfile resulting in `sudo` being skipped for those calls then?


lib/bundle/commands/cleanup.rb
```diff
@@ -41,6 +43,12 @@ def run(global: false, file: nil, force: false, zap: false)
             Kernel.system "code", "--uninstall-extension", extension
           end
 
+          # TODO: See how we can get around the sudo requirement because of the way e.g. basictex installs itself
+          # See also https://tug.org/texlive/doc/tlmgr.html#USER-MODE
+          tlmgr_packages.each do |package|
+            Kernel.system "sudo", "tlmgr", "remove", package
```
MikeMcQuaid, line 49: I'm afraid there's no way we're going to accept this if `sudo` is needed here, sorry.

MikeMcQuaid (changes requested):
Hey @Okeanos, thanks for a(another) PR! Just to set expectations: we're definitely not going to accept this so long as it requires `sudo` and we may not accept it even if it can be made to not because I've never heard a request for this before.

Just wanted to give you added context before you put a tonne more time into this, sorry ❤️

Okeanos:
Hey, thanks for the early feedback. I expected as much to be honest and mainly wanted to have a stab at it (regardless) because I like the `vscode` integration. I'll get the syntax issues etc out of the way regardless and would, even if this get rejected by default, like some minimal feedback if it isn't too much to ask. I am specifically interested in knowing whether the places I touched (based on the vscode PR #1208 and later changes) were technically correct & complete.

Why? Mainly because outside of that PR there's pretty little documentation on the topic of integrating additional formats and there were a [few interesting ideas](https://github.com/Homebrew/homebrew-bundle/pull/1208#issuecomment-1566021890) for additional extension types.

IntellIJ for example would [likely work](https://www.jetbrains.com/help/idea/install-plugins-from-the-command-line.html) within the acceptable bounds. I just found TeX Live more interesting to start with (also IntelliJ and other JetBrains tools are a little tricky if the [Toolbox](https://www.jetbrains.com/toolbox-app/) is used to manage IntelliJ etc – the Toolbox [doesn't have CLI](https://youtrack.jetbrains.com/issue/TBX-653/toolbox-cli) and putting `cask: 'jetbrains-toolbox'` and `intellij: 'some-extension'` into the same Brewfile is going to be … interesting.

MikeMcQuaid:
> Mainly because outside of that PR there's pretty little documentation on the topic of integrating additional formats and there were a [few interesting ideas](https://github.com/Homebrew/homebrew-bundle/pull/1208#issuecomment-1566021890) for additional extension types.

Yeh, I think it would be worth documenting what we'll consider/not if that'd be useful to folks like yourself?

Could also potentially figure out how to have a plugin system here.

> IntellIJ for example would [likely work](https://www.jetbrains.com/help/idea/install-plugins-from-the-command-line.html) within the acceptable bounds.

Yes, I could see that being a bit more acceptable 👍🏻. Could think about that in the docs?

MikeMcQuaid:
Passing on this accordingly, sorry 😭

Okeanos:
> Passing on this accordingly, sorry 😭

Hey, no worries! I did that without reaching to you first after all and when I noticed the `sudo` pieces, I already knew this was more than likely not gonna make it.

colindean:
Just in case anyone returns to this, tlmgr has a [user mode](https://tug.org/texlive/doc/tlmgr.html#USER-MODE) that enables setting a user-writable directory for packages to be installed. This would require setting `TEXMFHOME` and honoring that in not only `brew bundle` but in the consuming TeX engine. Invoking `tlmgr` with `--usermode` in Homebrew _may_ fix this.

But I agree with Mike's additions in #1333: I don't perceive enough demand for built-in tlmgr support, even though I might use it myself.

colindean:
Just in case anyone returns to this, tlmgr has a [user mode](https://tug.org/texlive/doc/tlmgr.html#USER-MODE) that enables setting a user-writable directory for packages to be installed. This would require setting `TEXMFHOME` and honoring that in the `brew bundle` and the consuming TeX engine. Invoking `tlmgr` with `--usermode` in Homebrew _may_ facilitate this.

But I agree with Mike's additions in #1333: I don't perceive enough demand presently for built-in tlmgr support, even though I might use it myself.

MikeMcQuaid:
> I don't perceive enough demand presently for built-in tlmgr support, even though I might use it myself.

Agreed 👍🏻

PR #1330, opened 2024-03-28
vscode_extension_installer: use UID for installs

Bo98:
This should make sense in the general case. VSCode should always install based on the calling user rather than the middle man.

### DIFF
--- a/lib/bundle/vscode_extension_installer.rb
+++ b/lib/bundle/vscode_extension_installer.rb
@@ -25,24 +25,27 @@ module Bundle
     end
 
     def exchange_uid(&block)
-      return yield if Process.euid == Process.uid
+      euid = Process.euid
+      uid = Process.uid
+      return yield if euid == uid
 
-      old_euid = Process.euid
-      if Process::UID.re_exchangeable?
+      old_euid = euid
+      process_reexchangeable = Process::UID.re_exchangeable?
+      if process_reexchangeable
         Process::UID.re_exchange
       else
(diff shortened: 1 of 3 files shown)


lib/bundle/vscode_extension_installer.rb
```diff
@@ -24,13 +24,36 @@ def preinstall(name, no_upgrade: false, verbose: false)
...
+
     def install(name, preinstall: true, no_upgrade: false, verbose: false, force: false)
       return true unless preinstall
       return true if extension_installed?(name)
 
       puts "Installing #{name} VSCode extension. It is not currently installed." if verbose
 
-      return false unless Bundle.system "code", "--install-extension", name, verbose: verbose
+      return false unless exchange_uid do
+        Bundle.system("code", "--install-extension", name, verbose:)
```
jacobbednarz, line 55: missing the `verbose` value here

jacobbednarz (changes requested):

lib/bundle/vscode_extension_installer.rb
MikeMcQuaid: @jacobbednarz Sorry, you're mistaken: this is newer, valid Ruby syntax.

jacobbednarz: TIL! explains why my old Ruby marked this as a syntax error. thanks!

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @Bo98!

PR #1334, opened 2024-04-01
Port bundle command to use AbstractCommand

dduugg:
See https://github.com/Homebrew/homebrew-services/pull/639

Discovered via https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/cmd/bundle_spec.rb (this should be the last PR of this type, going by `setup_remote_tap` call sites in `brew` tests)

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -1,197 +1,196 @@
 # frozen_string_literal: true
 
-require "cli/parser"
+require "abstract_command"
+
 module Homebrew
-  module_function
+  module Cmd
+    class BundleCmd < AbstractCommand
+      cmd_args do
+        usage_banner <<~EOS
+          `bundle` [<subcommand>]
 
-  def bundle_args
-    Homebrew::CLI::Parser.new do
-      usage_banner <<~EOS
-        `bundle` [<subcommand>]
+          Bundler for non-Ruby dependencies from Homebrew, Homebrew Cask, Mac App Store, Whalebrew and Visual Studio Code.
 
-        Bundler for non-Ruby dependencies from Homebrew, Homebrew Cask, Mac App Store, Whalebrew and Visual Studio Code.
+          `brew bundle` [`install`]:
+          Install and upgrade (by default) all dependencies from the `Brewfile`.
 
-        `brew bundle` [`install`]:
-        Install and upgrade (by default) all dependencies from the `Brewfile`.
+          You can specify the `Brewfile` location using `--file` or by setting the `HOMEBREW_BUNDLE_FILE` environment variable.
 
-        You can specify the `Brewfile` location using `--file` or by setting the `HOMEBREW_BUNDLE_FILE` environment variable.
+          You can skip the installation of dependencies by adding space-separated values to one or more of the following environment variables: `HOMEBREW_BUNDLE_BREW_SKIP`, `HOMEBREW_BUNDLE_CASK_SKIP`, `HOMEBREW_BUNDLE_MAS_SKIP`, `HOMEBREW_BUNDLE_WHALEBREW_SKIP`, `HOMEBREW_BUNDLE_TAP_SKIP`.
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Looks good, thanks again @dduugg!

ileitch:
This change appears to be breaking environments that use `HOMEBREW_NO_AUTO_UPDATE=1`, which is common in CI.

MikeMcQuaid:
@ileitch If you set `HOMEBREW_NO_AUTO_UPDATE` and do not e.g. pin your taps at a specific version: this will happen, sorry. As per all our issue templates: if `brew update` fixes the problem, we do not consider it a bug.

ileitch:
@MikeMcQuaid Because the lock file contains [local environment info](https://github.com/Homebrew/homebrew-bundle?tab=readme-ov-file#versions-and-lockfiles), we can't use a lock file in our repo. Having to execute `brew update` in every workflow is problematic because it introduces a performance impact and potential instability.

MikeMcQuaid:
@ileitch You'll need to figure out how to solve this yourself then, sorry.

MikeMcQuaid:
@ileitch Care to explain the 👎🏻?

ileitch:
@MikeMcQuaid The points I mentioned are very real problems for teams maintaining CI environments, and how you simply dismissed them is why I downvoted your comment. In our case, we explicitly introduced `HOMEBREW_NO_AUTO_UPDATE=0` to avoid a source of breakages and performance issues. I don't think we can just go back to implicit updates, and we may need to find another approach that doesn't depend on `brew bundle`, or perhaps even Homebrew altogether.

It's unfortunate you were not able to identify a migration strategy that does not introduce sudden breakages for your users, especially given the inflexibility around lock files. At least a warning of some kind would have been helpful.

MikeMcQuaid:
@ileitch I'm aware of these problems but also: Homebrew maintainers are volunteers and we are not obligated to help you. Explaining the problem and getting a no context 👎🏻 makes us disinclined to help you in future.

You not need to go back to implicit updates but, in this case, you need to be able to figure out a way to solve this problem yourself as the default, supported Homebrew configuration does not have this problem. If you're tapping a repository during your build process at whatever the latest revision is: you're asking Homebrew (a rolling-release package manager) to not update some parts of itself while running the latest versions of other parts. This is going to break and there's nothing we can do about this.

I've opened https://github.com/Homebrew/brew/pull/17003 to better explain that these issues should not be reported in these situations and that this is a known issue. I'm not sure what we can do beyond that; it's not tenable to have `master` of Homebrew/homebrew-bundle and other taps to work on arbitrary old versions of Homebrew (note: this issue would not have occurred if you were on the latest stable tag).

ileitch:
I apologize for the flippant downvote, though I felt I was responding in kind to a flippant comment on your part. As an open-source maintainer myself, I completely understand you are under no obligation to help me, nor do I expect you to. I appreciate the full explanation in your last comment, it provides the level of detail I think your initial comment was lacking. 

Perhaps there's a way to catch these errors as they occur, and explain the dangers around `HOMEBREW_NO_AUTO_UPDATE` and missing lock files. `HOMEBREW_NO_AUTO_UPDATE` is a fairly commonly used feature, in my experience, and I'm sure there are many users out there that will run into this sharp corner over the next few days.

MikeMcQuaid:
> I apologize for the flippant downvote

Thanks, I appreciate the apology ❤️ 

> Perhaps there's a way to catch these errors as they occur, and explain the dangers around `HOMEBREW_NO_AUTO_UPDATE` and missing lock files.

That's exactly the intent of https://github.com/Homebrew/brew/pull/17003.

MikeMcQuaid:
@ileitch https://github.com/Homebrew/homebrew-bundle/pull/1335 will now print an error if `brew bundle` is too old, too.

PR #1345, opened 2024-04-19
workflows: pin setup-ruby action

Moisan:
Pin the version of the `setup-ruby` action to full length commit SHA as described in the [security hardening for GitHub Actions guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
     - run: brew test-bot --only-tap-syntax
 
     - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@6bd3d993c602f6b675728ebaecb2b569ff86e99b # v1.174.0
       with:
         bundler-cache: true
 


jacobbednarz:
understandable that we want to be more in control of the versions, however, what are we actually aiming to achieve with this? given we use dependabot, the versions will be bumped automatically with a PR anyway. 

are we looking to mitigate the chance of a bump happening in CI before we have the opportunity to test it? or are we looking to be more across the minor version bumps? 

not against this by any means but do want to understand the driver a bit better. the cases where i've seen people do this in the past, it hasn't actually made any difference to their workflows and people just bump the versions anyway (it was more of a compliance checkbox).

MikeMcQuaid (approved):
Thanks @Moisan!

MikeMcQuaid:
> understandable that we want to be more in control of the versions, however, what are we actually aiming to achieve with this?

Tags are mutable, SHAs are not.

PR #1346, opened 2024-04-23
--quiet silences "Using xyz" in `brew bundle install`

### LINKED ISSUE #1202: Feature Request: Respect `--quiet`/`-q` cli flag

rrotter:
This is intended to resolve #1202. Only addresses the `brew bundle install` subcommand because I don't see any other subcommands where `--quiet` makes sense. I don't think it's desirable to suppress other non-error output (e.g. "Homebrew Bundle complete!..." and "Downloading https://formulae.brew.sh/api/cask.jws.json") unless we added a -qq option, which would require more far reaching changes to how arg parsing works (`--quiet` is parsed by code in `homebrew/brew`, not this repo).

Fixes #1202

### DIFF
--- a/lib/bundle/installer.rb
+++ b/lib/bundle/installer.rb
@@ -4,7 +4,8 @@ module Bundle
   module Installer
     module_function
 
-    def install(entries, global: false, file: nil, no_lock: false, no_upgrade: false, verbose: false, force: false)
+    def install(entries, global: false, file: nil, no_lock: false, no_upgrade: false, verbose: false, force: false,
+                quiet: false)
       success = 0
       failure = 0
 
@@ -42,7 +43,7 @@ module Bundle
           puts Formatter.success("#{verb} #{name}")
           true
         else
-          puts "Using #{name}"
+          puts "Using #{name}" unless quiet
           false
         end
 
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -107,6 +107,7 @@ module Homebrew
             no_upgrade: args.no_upgrade?,
             verbose:    args.verbose?,
             force:      args.force?,
+            quiet:      args.quiet?,
           )
 
           cleanup = if ENV.fetch("HOMEBREW_BUNDLE_INSTALL_CLEANUP", nil)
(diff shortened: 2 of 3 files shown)


lib/bundle/installer.rb
```diff
@@ -42,7 +43,7 @@ def install(entries, global: false, file: nil, no_lock: false, no_upgrade: false
           puts Formatter.success("#{verb} #{name}")
           true
         else
-          puts "Using #{name}"
+          puts "Using #{name}" unless quiet
```
jacobbednarz, line 46: don't we need this in the positive case of this conditional as well? otherwise, we'll still get `Upgrading...`/`Tapping...`/`Installing...` in the output.

MikeMcQuaid: I think it makes sense to output when something is being done and skip output when it's a no-op like this.

Can iterate on this post-merge.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @rrotter!

PR #1348, opened 2024-04-24
enforce limit of one subcommand, list named_args

rrotter:
* Add `named_args` so Homebrew::Completions can discover subcommands
* `raise` an error if `args.named.size > 1`, unless we're running `brew bundle exec`: this fails loudly on some commands that are currently accepted and may appear valid but don't do what you'd expect. Each of the below commands currently execute the first subcommand and ignore the second silently:
  * `brew bundle install my_package`
  * `brew bundle install cleanup` (not to be confused with `brew bundle install --cleanup`!)
  * `brew bundle cleanup install`
* note: cannot use the `max: 1` annotation for named args because this would break `brew bundle exec`

I wrote this commit as part of my abortive effort on zsh completion. Sending the PR anyway, because imo it's still an improvement to usability and will be of some assistance if anyone continues working on completions for `brew bundle`.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -92,13 +92,20 @@ module Homebrew
                description: "`dump` does not add `restart_service` to formula lines."
         switch "--zap",
                description: "`cleanup` casks using the `zap` command instead of `uninstall`."
+
+        named_args %w[install dump cleanup check exec list]
       end
 
       def run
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Great idea, thanks again @rrotter!

CC @dduugg who may find this interesting.

PR #1349, opened 2024-04-24
document running linter w/ test-bot

rrotter:
Maybe this should be obvious, but as a first time committer it took me a few rounds of CI fails before I figured out how to get this check to run locally. I'm not attached to the wording, but this should somehow be more discoverable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Note: based on these criteria, we would not accept e.g. Whalebrew (but have no p
 ## Tests
 
 Tests can be run with `bundle install && bundle exec rspec`.
+Syntax linting can be run with `brew style homebrew/bundle`.
 
 ## Copyright
 


README.md
```diff
@@ -92,6 +92,7 @@ Note: based on these criteria, we would not accept e.g. Whalebrew (but have no p
 ## Tests
 
 Tests can be run with `bundle install && bundle exec rspec`.
+Syntax linting can be run with `brew test-bot --only-tap-syntax --tap homebrew/bundle`.
```
MikeMcQuaid, line 95:
```suggestion
Syntax linting can be run with `brew style --tap homebrew/bundle`.
```

is probably a better option here?

rrotter: Ah, didn't know about that one. Fount the `test-bot` command from the ci log. Pushed w/ corrected syntax (your version, but w/o `--tap`).

MikeMcQuaid (approved):
Thanks, great work again @rrotter!

PR #1365, opened 2024-05-22
Support `__FILE__` in Brewfiles

Bo98:
`__FILE__` in Brewfiles currently is set to `(eval at ...)`. Given we never read the Brewfile without creating a DSL class, we can simplify things to pass the path and use that in `instance_eval` to set the `__FILE__` to the real path.

### DIFF
--- a/lib/bundle/commands/list.rb
+++ b/lib/bundle/commands/list.rb
@@ -7,7 +7,7 @@ module Bundle
 
       def run(global: false, file: nil, all: false, casks: false, taps: false, mas: false, whalebrew: false,
               vscode: false, brews: false)
-        parsed_entries = Bundle::Dsl.new(Brewfile.read(global:, file:)).entries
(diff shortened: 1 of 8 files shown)


lib/bundle/commands/list.rb
```diff
@@ -7,7 +7,7 @@ module List
 
       def run(global: false, file: nil, all: false, casks: false, taps: false, mas: false, whalebrew: false,
               vscode: false, brews: false)
-        parsed_entries = Bundle::Dsl.new(Brewfile.read(global:, file:)).entries
+        parsed_entries = Brewfile.read(global:, file:).entries
```
colindean, line 10: This affords that `Brewfile.read` should be renamed one of:

* `Brewfile.parse` 
* `Brewfile.read_and_parse`
* `Brewfile.slurp`

PR #1390, opened 2024-06-13
Apply --force flag if set for taps

tony-sol:
I have 
```ruby
tap "homebrew/core"
tap "homebrew/cask"
```
in my Brewfile and since those casks are no longer required, `brew bundle install` will receive and error via calling `brew tap :name`.

Passing `--force` flag in `brew bundle install` has no effect due to `TapInstaller.install()` `force` argument is ignored.

appliance-of-force-flag


MikeMcQuaid:
Makes sense, thanks @tony-sol!

PR #1420, opened 2024-08-05
cleanup: exit 1 when any action would have been taken

### LINKED ISSUE #1418: `brew bundle cleanup`: exit nonzero if action would be taken (or flag to opt into nonzero exit code)
`brew bundle cleanup` without `--force` is a convenient way to figure out if I installed some package but forgot to put it in my Brewfile — but it would be nice if it also exited with a nonzero status if any action would be taken. This would make it easy to use e.g. `brew bundle cleanup` as a pre-commit hook or other scripting workflows.

For now a workaround is to check if there was any output, but that feels a bit less programmatic and subject to change if `brew bundle cleanup` ever had any output added for the "happy path". For backward compatibility, it might make sense to make this opt-in behavior, perhaps with a `--check` flag. 

Alternatively, I suppose the `brew bundle check` command could be extended to work with `--cleanup` and also have the behavior I'm describing, since the semantics are already pretty similar, and this would mirror the behavior of `brew bundle install --cleanup`. From what I can tell, `--cleanup` has no effect on `check` in the current implementation.

Edit: one other possibility would be some combination of flags to `brew bundle install` flag that would fail the command if some packages need to be removed. I'm not sure there's much of a use case for this compared to the other suggestion options though

Would you be open to a PR implementing one of these options or something equivalent? I think it would be handy for devs looking to make sure their Brewfile and environment are always in sync. Thanks!

ian-h-chamberlain:
Closes #1418 

The exception is for `brew cleanup` action, since it's just an extra step that's
not really related to the Brewfile contents.

I think this also fixes an issue where `vscode` uninstalls would not have printed the `--force` message — now they will also be considered for printing the message and the exit code.

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -32,7 +32,7 @@ module Homebrew
 
           This workflow is useful for maintainers or testers who regularly install lots of formulae.
 
-          Unless `--force` is passed, this returns with an unsuccessful exit code if changes would be made, making it useful for scripting.
+          Unless `--force` is passed, this returns a 1 exit code if anything would be removed.
 
           `brew bundle check`:
           Check if all dependencies present in the `Brewfile` are installed.
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -32,6 +32,8 @@ module Homebrew
 
           This workflow is useful for maintainers or testers who regularly install lots of formulae.
 
+          Unless `--force` is passed, this returns with an unsuccessful exit code if changes would be made, making it useful for scripting.
+
           `brew bundle check`:
           Check if all dependencies present in the `Brewfile` are installed.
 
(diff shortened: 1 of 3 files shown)


cmd/bundle.rb
```diff
@@ -32,6 +32,8 @@ class BundleCmd < AbstractCommand
 
           This workflow is useful for maintainers or testers who regularly install lots of formulae.
 
+          Unless `--force` is passed, this returns with an unsuccessful exit code if changes would be made, making it useful for scripting.
```
MikeMcQuaid, line 35:
```suggestion
          Unless `--force` is passed, this returns a 1 exit code if anything would be removed.
```

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @ian-h-chamberlain!

PR #1454, opened 2024-09-29
brew_installer: fix use of force flag

yelsa:
Fixes passing `brew bundle --force` to:
- make formula installation run `brew install <formula> --force --overwrite`
- run `brew link --overwrite <formula>` to force relinking of an unlinked non-keg by default (i.e. `link: nil`)

Previously =>
> $ brew bundle --force  --verbose
> $ Installing gnupg formula. It is not currently installed.

Now =>
> $ brew bundle --force  --verbose
> $ Installing gnupg formula with --force --overwrite. It is not currently installed.

### DIFF
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -12,8 +12,8 @@ module Bundle
       new(name, options).preinstall(no_upgrade:, verbose:)
     end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Thanks @yelsa, good catch!

PR #1455, opened 2024-09-30
Support `link: :overwrite` option

yelsa:
- set `link: :overwrite` per formula
- uses `brew link --overwrite` to force linking and overwrite all conflicting files
- use the `linked?` and `keg_only?` predicates available from `Formula`
- `--force` should only be passed when it's keg-only
- check linked/unlinked state to avoid re-calling `brew link` / `brew unlink`

Closes https://github.com/Homebrew/homebrew-bundle/issues/1062
Resolves https://github.com/Homebrew/homebrew-bundle/pull/1116#discussion_r975223823, https://github.com/Homebrew/homebrew-bundle/pull/1116#discussion_r976260667

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,8 @@ cask_args appdir: "~/Applications", require_sha: true
 
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
+# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' on version changes
+brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :changed
 # 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
 brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
 # install only on specified OS
--- a/lib/bundle/brew_installer.rb
+++ b/lib/bundle/brew_installer.rb
@@ -46,7 +46,7 @@ module Bundle
 
       if installed?
         service_change_state!(verbose:) if install_result
-        link_change_state!(verbose:, force:)
+        link_change_state!(verbose:)
       end
 
       install_result
@@ -97,29 +97,34 @@ module Bundle
       end
     end
 
-    def link_change_state!(verbose: false, force: false)
-      case @link
+    def link_change_state!(verbose: false)
+      link_args = []
+      link_args << "--force" if unlinked_and_keg_only?
+
+      cmd = case @link
+      when :overwrite
+        link_args << "--overwrite"
+        "link" unless linked?
       when true
-        unless linked_and_keg_only?
-          puts "Force-linking #{@name} formula." if verbose
-          Bundle.system(HOMEBREW_BREW_FILE, "link", "--force", @name, verbose:)
-        end
+        "link" unless linked?
       when false
-        unless unlinked_and_not_keg_only?
-          puts "Unlinking #{@name} formula." if verbose
-          Bundle.system(HOMEBREW_BREW_FILE, "unlink", @name, verbose:)
-        end
+        "unlink" if linked?
       when nil
-        if unlinked_and_not_keg_only?
-          puts "Linking #{@name} formula." if verbose
-          link_args = ["link"]
-          link_args << "--overwrite" if force
-          Bundle.system(HOMEBREW_BREW_FILE, *link_args, @name, verbose:)
-        elsif linked_and_keg_only?
-          puts "Unlinking #{@name} formula." if verbose
-          Bundle.system(HOMEBREW_BREW_FILE, "unlink", @name, verbose:)
+        if keg_only?
+          "unlink" if linked?
+        else
+          "link" unless linked?
         end
       end
+
+      if cmd
+        verb = "#{cmd}ing".capitalize
+        with_args = " with #{link_args.join(" ")}" if link_args.present?
+        puts "#{verb} #{@name} formula#{with_args}." if verbose
+        return Bundle.system(HOMEBREW_BREW_FILE, cmd, *link_args, @name, verbose:)
+      end
+
+      true
     end
 
     def self.formula_installed_and_up_to_date?(formula, no_upgrade: false)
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -37,8 +37,8 @@ cask_args appdir: "~/Applications", require_sha: true
 
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", args: ["with-rmtp"], restart_service: :changed
+# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' on version changes
+brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :changed
```
MikeMcQuaid, line 41: Not sure about this DSL; having `true`/`false`/`:overwrite` as valid values is a bit confusing. Thoughts about `link_overwrite: true` instead?

lib/bundle/brew_installer.rb
```diff
@@ -97,29 +97,34 @@ def service_change_state!(verbose:)
...
-          puts "Unlinking #{@name} formula." if verbose
-          Bundle.system(HOMEBREW_BREW_FILE, "unlink", @name, verbose:)
+        if keg_only?
+          "unlink" if linked?
+        else
+          "link" unless linked?
         end
       end
+
+      if cmd
```
MikeMcQuaid, line 120:
```suggestion
      if cmd.present?
```

lib/bundle/brew_installer.rb
```diff
@@ -97,29 +97,34 @@ def service_change_state!(verbose:)
...
+      link_args = []
+      link_args << "--force" if unlinked_and_keg_only?
+
+      cmd = case @link
+      when :overwrite
+        link_args << "--overwrite"
+        "link" unless linked?
       when true
-        unless linked_and_keg_only?
-          puts "Force-linking #{@name} formula." if verbose
```
MikeMcQuaid, line 104: This seems like a behaviour, maybe messaging, change? Can you detail the expected changes in behaviour/messaging? Thanks!

MikeMcQuaid (commented):
Thanks @yelsa! Looking good so far.

README.md
yelsa: Was thinking of `:overwrite` as a simple way to get at the extra forceful version of `true` (that also implies `true`).
Given the inherent danger of overwrite maybe it is better to have a very clear separate option like you suggest. 
Will users be confused about how `link:` `true` / `false` / `nil` interacts with `link_overwrite: true` ?

lib/bundle/brew_installer.rb
yelsa: Here goes, let me know if there's something you think I miss.

In relation to the changes in the `case` statement logic so it checks against `linked?`:
1. There were two cases where it was possible a second run of `brew bundle` would recall `brew link` or `brew unlink`. First gave a Warning, second a full on-disk run with zero changes e.g.

a.  
```
Using gnupg
Force-linking gnupg formula.
Warning: Already linked: /usr/local/Cellar/gnupg/2.4.5_1
To relink, run:
  brew unlink gnupg && brew link gnupg
```

b. 
```
Using gnupg@2.2
Unlinking gnupg@2.2 formula.
Unlinking /usr/local/Cellar/gnupg@2.2/2.2.44... 0 symlinks removed.
```
2. This behaviour now cannot occur.

The verbose output changes to be like how output is done on install. The only existing output that will change is that "`Force-linking <@formula_name> formula`" becomes "`Linking <@formula_name> formula with "--force"`".

A couple of changes in relation to `force` that have no user visible effect:
1. `brew help link` suggests `--force` is to "Allow keg-only formulae to be linked." Deleted line 103 called `--force` on both keg-only and not-keg-only. 
a. Now only add`--force` as an argument in the one case it is relevant, an unlinked keg-only keg.
b. Even though it was a no-op in the not-keg-only case, this refinement is cleaner expression of intent.
4. Can drop `force:` from the parameters list as it is redundant for linking:
a. The original global flag would have no effect unless you had also set `link: true` against a formula in your `Brewfile` 
b. As we can assume that if you explicitly set `link: true` against a formula in your `Brewfile` then you definitely want linking done we can consider the keg-only need for `--force` to be implied.

README.md
MikeMcQuaid: > Was thinking of `:overwrite` as a simple way to get at the extra forceful version of `true` (that also implies `true`).

Ok, makes sense 👍🏻 

> Will users be confused about how `link:` `true` / `false` / `nil` interacts with `link_overwrite: true` ?

Yes, maybe, unless we make it e.g. error if both are passed?

lib/bundle/brew_installer.rb
MikeMcQuaid: Thanks for explaining. This seems to make sense to me 👍🏻

README.md
yelsa: I've worked through the interactions a bit further to give more insight on:
> Not sure about this DSL; having true/false/:overwrite as valid values is a bit confusing. 

> Will users be confused about how link: true / false / nil interacts with link_overwrite: true ?

The role of `nil` looks to be particularly important. It probably is almost never actually specified in a Brewfile but it has the most complex role as an 'adopt defaults' choice that results in directly opposite actions for keg_only vs not_keg_only formulas.

Expanding on the two proposed options for extending the Brewfile dsl we have:
1. adding `:overwrite` as an additional state for the `link:` option

| Brewfile                       | Command run                                       |
|--------------------------------|---------------------------------------------------|
| brew "mysql", link: :overwrite | `brew link --overwrite`                             |
| brew "mysql", link: :true      | `brew link`                                         |
| brew "mysql", link: :false     | `brew unlink`                                       |
| brew "mysql", *[link: :nil]*     | Not_keg_only: `brew link`          Keg_only: `brew unlink` |

2. adding a new `link_overwrite:` option

| Brewfile                                                   | Command run             |
|------------------------------------------------------------|-------------------------|
| brew "gnupg", link:true, link_overwrite: true              | `brew link --overwrite`   |
| brew "gnupg", link:false, *[link_overwrite: false/nil]*      | `brew unlink`             |
| brew "gnupg", link:false, link_overwrite: true             | Not a valid combination |
|                                                            |                         |
| *Not_keg_only*                                             |                         |
| brew "gnupg", *[link:nil]*, link_overwrite: true             | `brew link --overwrite`   |
| brew "gnupg", *[link:nil/true]*, *[link_overwrite: false/nil]* | `brew link`               |
|                                                            |                         |
| *Keg_only*                                                 |                         |
| brew "gnupg@2.2", link:true, *[link_overwrite: false/nil]*   | `brew link`               |
| brew "gnupg@2.2", *[link:nil]*, *[link_overwrite: false/nil]*  | `brew unlink`             |
| brew "gnupg@2.2", *[link:nil]*, link_overwrite: true         | Not a valid combination |

Re: option 2 'Not a valid combination':
1.  `brew unlink --overwrite` is not a thing (it is nonsensical)
2. The combinations are therefore contradictory in intent - the `link` portion suggests `brew unlink` but `link_overwrite:true` suggests user expects to `brew link`. Which should take precedence?
3. A bare `brew "<formula>", link_overwrite: true` is probably going to result in user confusion as it won't work for keg_only but will work for not_keg_only. 
4. We would need to choose to error or do nothing. Based on above points i'd say it should be errors.

MikeMcQuaid: > It probably is almost never actually specified in a Brewfile but it has the most complex role as an 'adopt defaults' choice that results in directly opposite actions for keg_only vs not_keg_only formulas.

Agreed 👍🏻 



> Expanding on the two proposed options for extending the Brewfile dsl we have:

@yelsa Great work spelling these out. Intentionally or not: I'm now convinced that `link: :overwrite` is the best option here. Thanks!

MikeMcQuaid (approved):
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @yelsa!

shanesmith:
This has unfortunately broken file path formulas: https://github.com/Homebrew/homebrew-bundle/issues/1460

As mentioned in the issue I've noticed file path formulas are essentially deprecated so not sure if it's worth a revert or an attempt to fix forward.

Issue #1456, opened 2024-09-30
Option to exclude/disable vscode in dump


JtMotoX:
Would it be possible to have a `brew bundle dump` flag, something like `--no-vscode` to disable the vscode integration?  I use vscode built-in sync functionality and this is causing some issues.

Currently, I have to perform `sed -i '/^vscode /d' $HOME/Brewfile` after every dump.

jacobbednarz:
today there isn't a way to selectively output some types over others. the assumption is that if you have it installed, you will want to manage it via `brew bundle` -- in your case, that isn't true.

i don't know how widely used this would be but open to leaving this here to gauge the interest and welcome a PR if you want to see this land.

JtMotoX:
This has been requested in https://github.com/Homebrew/homebrew-bundle/issues/1209 and https://github.com/Homebrew/homebrew-bundle/issues/1212 so there are others that are interested in this option.

MikeMcQuaid:
@JtMotoX would you prefer `brew bundle dump --skip-vscode` or something like `export HOMEBREW_BUNDLE_NO_VSCODE=1`?

JtMotoX:
`--skip-vscode` would be preferred

MikeMcQuaid:
@JtMotoX you're welcome, thanks for taking the time to say thanks!

Issue #1212, opened 2023-05-31
vscode extensions have written into Brewfile event without --vscode argument 


tony-sol:
I have Brewfile, created by 
```
brew bundle dump --all --describe --force
```
But after last updates, it also writes vscode extensions.
So, i tried to run dump as
```
brew bundle dump --casks --taps --brews --describe --force
```
but it still writes vscode extensions into Brewfile

MikeMcQuaid:
Yes, as documented, `--casks` etc. only affect `brew bundle list`. Please either manually remove them or consider adding a PR for `--no-vscode` or similar.

Issue #1209, opened 2023-05-30
Request: Ability to disable vscode extension handling


duckpuppy:
I've been using a script for a very long time that installs items in my global Brewfile and then cleans up with `zap` after the fact. Now that it understands vscode extensions, it uninstalls every one of them each time I try to run it. I use the built-in vscode sync to keep extensions synced across multiple systems, not all of which are macOS/Homebrew managed.

There should be a way to disable the management of vscode extensions so that users that don't want to use brew to manage them can still benefit from the cleanup commands without losing their setup every time.

MikeMcQuaid:
Thanks for the feedback. Is there a reason you would not like to dump your VSCode extensions also to your `Brewfile`?

duckpuppy:
I'm not sure where the anger is coming from... You asked for my reasons and I simply provided them. I was unaware you were on vacation, so really that didn't enter my mind. 

I was certainly surprised that you fixed it so quickly, so a belated thank you for that. Perhaps I should have done that earlier, and that's certainly on me. I apologize for that. 

In any case, this was a catastrophic failure mode for a feature that didn't exist two days ago and was a surprise to anyone who routinely used `--cleanup`. My mention of using the sync feature was meant to be illustrative. Anyone who DOESN'T use that would have had no recourse. I had no idea that vscode extension management was incoming.

duckpuppy:
If you feel that I was being belligerent in any way. I apologize. I certainly do appreciate the work done on Homebrew.

MikeMcQuaid:
> I'm not sure where the anger is coming from

No anger here ❤️ 

> I was certainly surprised that you fixed it so quickly, so a belated thank you for that. Perhaps I should have done that earlier, and that's certainly on me. I apologize for that.

Much appreciated 🙇🏻

PR #1459, opened 2024-10-02
Add --no-vscode and HOMEBREW_BUNDLE_DUMP_NO_VSCODE

MikeMcQuaid:
This allows excluding VSCode output from `brew bundle dump` with a flag or environment variable.

While we're here
- generally refactor how we handle command arguments so this logic is pushed into `cmd/bundle` and made more consistent
- fix early RSpec failure due to an unstubbed `exec` call
- refactor some specs to be more idiomatic

Fixes https://github.com/Homebrew/homebrew-bundle/issues/1209
Fixes https://github.com/Homebrew/homebrew-bundle/issues/1212
Fixes https://github.com/Homebrew/homebrew-bundle/issues/1456

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -80,17 +80,21 @@ module Homebrew
         switch "--all",
                description: "`list` all dependencies."
         switch "--formula", "--brews",
-               description: "`list` Homebrew formula dependencies."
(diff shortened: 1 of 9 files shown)


jacobbednarz (approved):

MikeMcQuaid:
Thanks @jacobbednarz!

PR #1476, opened 2024-10-27
add explicit_s method from `Cask#Config` of `homebrew/brew` 

hyuraku:
solve  https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/config.rb#L197


MikeMcQuaid (approved):
Thanks @hyuraku!

PR #1495, opened 2024-11-11
Keep implicit taps

tt:
I have a Brewfile that contains the following line:

```ruby
brew "heroku/brew/heroku"
```

Running `brew bundle cleanup --verbose` includes the following output:

```
Would untap:
heroku/brew
```

However, this tap can't be removed as the fomula depends on it.

Considering you can install formulae without explicitly adding a `tap` declaration, it seems that these should be respected when computing which taps to keep.

This change does that: Any tap found in a `brew` declaration will automatically be kept.

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -92,8 +92,7 @@ module Bundle
       def formulae_to_uninstall(global: false, file: nil)
         @dsl ||= Brewfile.read(global:, file:)
         kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
-        kept_cask_formula_dependencies = Bundle::CaskDumper.formula_dependencies(kept_casks)
-        kept_formulae += kept_cask_formula_dependencies
+        kept_formulae += Bundle::CaskDumper.formula_dependencies(kept_casks)
         kept_formulae.map! do |f|
           Bundle::BrewDumper.formula_aliases[f] ||
             Bundle::BrewDumper.formula_oldnames[f] ||
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -153,7 +153,9 @@ module Bundle
 
       def taps_to_untap(global: false, file: nil)
         @dsl ||= Brewfile.read(global:, file:)
+        kept_formulae = self.kept_formulae(global:, file:).map(&Formulary.method(:factory))
         kept_taps = @dsl.entries.select { |e| e.type == :tap }.map(&:name)
+        kept_taps += kept_formulae.filter_map(&:tap).map(&:name)
         current_taps = Bundle::TapDumper.tap_names
         current_taps - kept_taps - IGNORED_TAPS
       end
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -11,6 +11,7 @@ module Bundle
       def reset!
         @dsl = nil
         @kept_casks = nil
+        @kept_formulae = nil
         Bundle::CaskDumper.reset!
         Bundle::BrewDumper.reset!
         Bundle::TapDumper.reset!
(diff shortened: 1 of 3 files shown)


lib/bundle/commands/cleanup.rb
```diff
@@ -145,7 +145,9 @@ def recursive_dependencies(current_formulae, formulae_names, top_level: true)
 
       def taps_to_untap(global: false, file: nil)
         @dsl ||= Brewfile.read(global:, file:)
+        kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name).map(&Formulary.method(:factory))
```
MikeMcQuaid, line 148: A `kept_formula` method that extracts/shares some logic with `formulae_to_uninstall` and memoizes to avoid doing the same logic twice would be ideal.

spec/stub/formula.rb
```diff
@@ -3,10 +3,13 @@
 require "ostruct"
 require "pathname"
 
+HOMEBREW_TAP_NAME_REGEX = %r{\A(?<tap>(?:[^/]+)/(?:[^/]+))/(?:[\w+\-.@]+)\Z}
```
MikeMcQuaid, line 6: Where is this from?

MikeMcQuaid (commented):
Thanks @tt! Good idea. Looking good so far.

lib/bundle/commands/cleanup.rb
tt: Happy to make the change but looking at `kept_casks`, it accepts two arguments and then ignores them after the first invocation. Maybe not an issue but seems like a potential footgun.

How much do you want the new method to include? All the resolution of dependencies too? I'm not sure if that would potentially yield more taps and therefore be necessary to make my chance catch all cases.

`@dsl.entries.select { |e| e.type == :brew }.map(&:name)` could arguably be a helper method on `Bundle::Dsl` also.

spec/stub/formula.rb
tt: It's a combination of [these two patterns](https://github.com/Homebrew/brew/blob/4.4.4/Library/Homebrew/tap_constants.rb#L4-L10). Is there a better way to do this considering it's only for testing?

MikeMcQuaid: @tt just an inline Ruby comment clarifying that would be handy!

lib/bundle/commands/cleanup.rb
MikeMcQuaid: > How much do you want the new method to include? All the resolution of dependencies too? I'm not sure if that would potentially yield more taps and therefore be necessary to make my chance catch all cases.

I think just consistency is the main thing. It's not clear as-is that `kept_formula` and `formulae_to_uninstall` will be consistent which is not ideal.

> How much do you want the new method to include? All the resolution of dependencies too? I'm not sure if that would potentially yield more taps and therefore be necessary to make my chance catch all cases.

Up to you. As above: I think consistency is the main thing to aim for.

> `@dsl.entries.select { |e| e.type == :brew }.map(&:name)` could arguably be a helper method on `Bundle::Dsl` also.

Leave it in this file for this PR, thanks.

lib/bundle/commands/cleanup.rb
```diff
@@ -90,24 +90,31 @@ def casks_to_uninstall(global: false, file: nil)
...
+
+        @kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
+        @kept_formulae += Bundle::CaskDumper.formula_dependencies(kept_casks)
+        @kept_formulae.map! do |f|
+          Bundle::BrewDumper.formula_aliases[f] ||
+            Bundle::BrewDumper.formula_oldnames[f] ||
+            f
+        end
+
+        @kept_formulae += recursive_dependencies(Bundle::BrewDumper.formulae, kept_formulae)
```
MikeMcQuaid, line 115:
```suggestion
        kept_formulae = @dsl.entries.select { |e| e.type == :brew }.map(&:name)
        kept_formulae += Bundle::CaskDumper.formula_dependencies(kept_casks)
        kept_formulae.map! do |f|
          Bundle::BrewDumper.formula_aliases[f] ||
            Bundle::BrewDumper.formula_oldnames[f] ||
            f
        end

        kept_formulae += recursive_dependencies(Bundle::BrewDumper.formulae, kept_formulae)
        @kept_formulae = kept_formulae
```

to avoid setting the instance variable until we've fully populated the kept_formulae

MikeMcQuaid (approved):
Looks good, thanks @tt! One small tweak and we should be good to go.

lib/bundle/commands/cleanup.rb
tt: I changed it to use a block and also realized that I missed `@kept_formulae = nil` in the `reset!` method.

MikeMcQuaid (approved):
Thanks so much for your great, first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @tt!

PR #1508, opened 2024-11-24
Ignore taps of unavailable formulae

tt:
This addresses an issue in #1495.

I set up a new machine and ran `brew bundle cleanup --verbose` which then exited with the following error:

```
Error: No available formula with the name "heroku/brew/heroku".
Please tap it and then try again: brew tap heroku/brew
```

This is because I hadn't installed anything and therefore `Formulary.factory` couldn't find the formula.

With this change, I'm making an assumption that if we can't find the formula, we don't need to keep its associated tap and therefore it's fine to ignore the exception and consequently not include the tap in `kept_taps`.

~I would like to add a test case but `TapFormulaUnavailableError` isn't available and wasn't sure it was desired to define it for testing. Let me know if I should.~ I found [spec/stub/exceptions.rb](https://github.com/Homebrew/homebrew-bundle/blob/master/spec/stub/exceptions.rb).

### DIFF
--- a/lib/bundle/commands/cleanup.rb
+++ b/lib/bundle/commands/cleanup.rb
@@ -153,7 +153,7 @@ module Bundle
 
       def taps_to_untap(global: false, file: nil)
         @dsl ||= Brewfile.read(global:, file:)
-        kept_formulae = self.kept_formulae(global:, file:).map(&method(:lookup_formula))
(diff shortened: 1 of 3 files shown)


MikeMcQuaid (approved):
Makes sense, thanks again @tt!

PR #1509, opened 2024-11-26
Remove `Brewfile.lock.json` generation.

MikeMcQuaid:
This was originally added to faciliate later functionality from other maintainers that never materialized.

I've never heard anyone say this was useful and a lot of people have complained that it's useless or annoying.

I won't merge this immediately to give people a change to comment.


p-linnane (approved):
Makes sense. I've also never understood the purpose, so your explanation makes sense.

colindean:
I have no objections. I've got a line for `Brewfile.lock.json` in every `.gitignore` in a repo with a `Brewfile`.

colindean (approved):

jacobbednarz (approved):

jacobbednarz:
works for me!

Brett-Best:
I actually found this helpful and had it committed to the repository.
It had a bunch of information about the environment in a last known working state which made finding where things broke a bit easier to debug.

It's not a big deal but perhaps it could default to not generating the lockfile and make it opt in only.

Xe:
My CI relied on this for caching homebrew packages so they didn't have to be downloaded twice. Is there a way to re-enable spacebar heating?

MikeMcQuaid:
Thanks for chiming in @georgettica @Xe! Can you elaborate on how you used/consumed this? Thanks!

jasonkarns:
Was there a deprecation period for the `--no-lock` option? I have loads of scripts and applications whose bootstrap process uses `brew bundle --no-lock` that are all failing now with `Error: invalid option: --no-lock`.

Xe:
> Thanks for chiming in @Xe! Can you elaborate on how you used/consumed this? Thanks!

GitHub actions. I wanted to avoid installing packages needlessly (this massively speeds up CI, you have no idea, it can save two digits of minutes per run) so I cache the /home/linuxbrew folder in GitHub Actions caches. Brewfile.lock.json changing means that the cache can be invalidated.

MikeMcQuaid:
@Xe Thanks! Could you open an issue for that functionality (GitHub Actions caching) as I think it'd be valuable and may make more sense in Homebrew/brew.

Issue #1002, opened 2021-09-08
Global Brewfile would be better placed somewhere under ~/.config or ~/.local as per XDG


tko:
The global Brewfile is somewhat inconveniently placed as ~/.Brewfile (macos) with no configuration or environment variables allowing to change its location. This is doubly inconvenient given the additional generation of .Brewfile.lock.json in the home folder.

Instead of adding to polluting the home folder with gazillion other dotfiles with no structure it would be better to place the files in a subfolder as per XDG.

The Brewfile should almost certainly be used as `~/.config/homebrew-bundle/Brewfile` but I'm uncertain of whether the lockfile should be better treated as configuration or data.

ref: [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

jacobbednarz:
You can toggle where the file is dumped to and read from using `--file` of you'd like it somewhere else.

MikeMcQuaid:
macOS does not follow the XDG. homebrew-bundle was originally macOS only. We cannot change the default location without breaking all existing users.

tko:
Operating systems don't follow the XDG, applications do. Many applications can look up configuration files from multiple locations in order of preference, using the first one that's found. Changing default location with a fallback to previous shouldn't break anything.

I wouldn't necessarily object to using `~/Library/Application Support` like macos apps are meant to but for terminal apps I think XDG is more appropriate.

MikeMcQuaid:
> Changing default location with a fallback to previous shouldn't break anything.

We'll review a PR to fallback to XDG locations if the necessary variables are set.

PR #1549, opened 2025-01-02
look for Brewfile in HOMEBREW_USER_CONFIG_HOME

rrotter:
Indirectly support XDG_CONFIG_HOME by leaning on `brew`'s support for it. 

when `--global` is set, we'll get the brewfile from:
- $HOMEBREW_BUNDLE_FILE_GLOBAL if set
- $XDG_CONFIG_HOME/homebrew/Brewfile (if XDG_CONFIG_HOME is set and file exists)
- $HOME/.homebrew/Brewfile (if XDG_CONFIG_HOME _is not_ set, and file exists)
- ~/.Brewfile (unconditional fallback, ~but if this doesn't exist either, we should probably use $HOMEBREW_USER_CONFIG_HOME/Brewfile as the new default~ [edit: per feedback])

Starting this as a draft because:
- looking for feedback on whether this change is welcome at all (I suspect so, per: https://github.com/Homebrew/homebrew-bundle/issues/1002#issuecomment-917613646)
- looking for feedback on what the fallback default should be if no file exists
- I still need to add tests. I'll need to add fixtures for a home dir, since currently tests are using the user's real $HOME, which we don't want to read or write from in rspec.
- need to update readme and `brew bundle --help` text

### DIFF
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -56,8 +56,9 @@ module Homebrew
         flag "--file=",
              description: "Read the `Brewfile` from this location. Use `--file=-` to pipe to stdin/stdout."
         switch "--global",
-               description: "Read the global `Brewfile`, from the homebrew config directory or `~/.Brewfile`. " \
-                            "Override this path with the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable."
+               description: "Read the `Brewfile` from `~/.Brewfile`, " \
+                            "in the `HOMEBREW_USER_CONFIG_HOME` directory, " \
+                            "or the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable, if set."
         switch "-v", "--verbose",
                description: "`install` prints output from commands as they are run. " \
                             "`check` lists all missing dependencies."
--- a/cmd/bundle.rb
+++ b/cmd/bundle.rb
@@ -56,8 +56,8 @@ module Homebrew
         flag "--file=",
              description: "Read the `Brewfile` from this location. Use `--file=-` to pipe to stdin/stdout."
         switch "--global",
-               description: "Read the `Brewfile` from `~/.Brewfile` or " \
-                            "the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable, if set."
+               description: "Read the global `Brewfile`, from the homebrew config directory or `~/.Brewfile`. " \
+                            "Override this path with the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable."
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
> * $HOME/.homebrew/Brewfile (if XDG_CONFIG_HOME _is not_ set, and file exists)
> * but if this doesn't exist either, we should probably use $HOMEBREW_USER_CONFIG_HOME/Brewfile as the new default

This is the only part I'm 👎🏻 on, I don't see any users wanting this.

> * looking for feedback on what the fallback default should be if no file exists

The current location.

MikeMcQuaid (commented):
Code mostly makes sense so far. I'd be 👍🏻 on supporting behaviour change if `XDG_CONFIG_HOME` is set and otherwise I'd like to leave everything as-is. Thanks!

rrotter:
Updated help text, added spec.

cmd/bundle.rb
```diff
@@ -56,8 +56,8 @@ class BundleCmd < AbstractCommand
         flag "--file=",
              description: "Read the `Brewfile` from this location. Use `--file=-` to pipe to stdin/stdout."
         switch "--global",
-               description: "Read the `Brewfile` from `~/.Brewfile` or " \
-                            "the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable, if set."
+               description: "Read the global `Brewfile`, from the homebrew config directory or `~/.Brewfile`. " \
+                            "Override this path with the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable."
```
MikeMcQuaid, line 60:
```suggestion
               description: "Read the `Brewfile` from `~/.Brewfile`, " \
                            "in the `HOMEBREW_USER_CONFIG_HOME` directory, " \
                            "or the `HOMEBREW_BUNDLE_FILE_GLOBAL` environment variable, if set."
```

MikeMcQuaid (approved):
Thanks @rrotter, great work!

rrotter:
@MikeMcQuaid, thanks. I went through several iterations on the help text before committing this. Not at all bothered that you decided to tweak it, but I just want to point out that I specifically avoided mentioning $HOMEBREW_USER_CONFIG_HOME in the help text (even though it's in the PR title) because it's a brew internal variable which can't be set by the user:

https://github.com/Homebrew/brew/blob/9fd2b27a353726b2ea1641b9b0477d5cc25ac595/bin/brew#L172-L177

MikeMcQuaid:
Thanks @rrotter, my bad. Fixing in https://github.com/Homebrew/homebrew-bundle/pull/1579

PR #1555, opened 2025-01-08
Add `postinstall` support for `brew` and `cask`

MikeMcQuaid:
This allows `postinstall` to be used to run a command after a formula or cask is installed.

While we're here, also adjust the `restart_service` behaviour to correctly match the comment i.e. require `always` to restart every time and otherwise just restart on an install or upgrade of a formula.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,13 @@ cask_args appdir: "~/Applications", require_sha: true
 
 # 'brew install'
 brew "imagemagick"
-# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :changed
+# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' even if no install/upgrade
+brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :always
 # 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
+brew "mysql@5.6", restart_service: :changed, link: true, conflicts_with: ["mysql"]
+# 'brew install' and run a command if installer or upgraded.
+brew "postgresql@16",
+     postinstall: "${HOMEBREW_PREFIX}/opt/postgresql@16/bin/postgres -D ${HOMEBREW_PREFIX}/var/postgresql@16"
 # install only on specified OS
 brew "gnupg" if OS.mac?
 brew "glibc" if OS.linux?
@@ -55,6 +58,8 @@ cask "firefox", args: { no_quarantine: true }
 cask "opera", greedy: true
 # 'brew install --cask' only if '/usr/libexec/java_home --failfast' fails
 cask "java" unless system "/usr/libexec/java_home", "--failfast"
+# 'brew install --cask' and run a command if installer or upgraded.
+cask "google-cloud-sdk", postinstall: "${HOMEBREW_PREFIX}/bin/gcloud components update"
 
 # 'mas install'
 mas "1Password", id: 443_987_910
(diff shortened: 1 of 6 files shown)


lib/bundle/brew_installer.rb
```diff
@@ -132,6 +135,14 @@ def link_change_state!(verbose: false)
       true
     end
 
+    def postinstall_change_state!(verbose:)
+      return true if @postinstall.blank?
+      return true unless changed?
+
+      puts "Running postinstall for #{@name}." if verbose
+      Bundle.system(@postinstall, verbose:)
```
Bo98, line 143: Does this work with the README example where it's all one string rather than separate arguments? `Bundle.system` intentionally avoids shell execution so may need another method.

Bo98 (approved):

lib/bundle/brew_installer.rb
MikeMcQuaid: @Bo98 Yup, still works, that's what I tested locally.

spec/bundle/brew_installer_spec.rb
```diff
@@ -449,8 +501,12 @@ def expectations(verbose:)
         allow_any_instance_of(described_class).to receive(:changed?).and_return(false)
       end
 
-      it "is true with {restart_service: true}" do
-        expect(described_class.new(formula, restart_service: true).restart_service_needed?).to be(true)
+      it "is false with {restart_service: true}" do
+        expect(described_class.new(formula, restart_service: true).restart_service_needed?).to be(false)
+      end
```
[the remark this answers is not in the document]
MikeMcQuaid: @Bo98 It's a behaviour change but not a breaking one. It'll still be run on installs/upgrades, just no longer unconditionally on every `brew bundle` run. `restart_service: true` will imply "on change" rather than "every time regardless of change". Also, if the service isn't running, it'll be started regardless.

Given we're doing this behaviour now for services, link and post install it feels like it makes more sense to be consistent here than hold onto previous behaviour.

Of course, if people loudly complain otherwise: we can switch it back.

Thanks for calling this out, though, it warranted further explanation! ❤️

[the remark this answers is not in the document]
Bo98: Ok if it's intentional then sounds good. Just one further question: is `brew_dumper` still outputting `true` correct with these changes?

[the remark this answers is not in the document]
MikeMcQuaid: @Bo98 I've adjusted it to output `:changed` and updated the README accordingly, good idea.

jacobbednarz (approved):

README.md
```diff
@@ -37,10 +37,13 @@ cask_args appdir: "~/Applications", require_sha: true
...
-# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' on version changes
-brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :changed
+# 'brew install --with-rmtp', 'brew link --overwrite', 'brew services restart' even if no install/upgrade
+brew "denji/nginx/nginx-full", link: :overwrite, args: ["with-rmtp"], restart_service: :always
 # 'brew install', always 'brew services restart', 'brew link', 'brew unlink mysql' (if it is installed)
-brew "mysql@5.6", restart_service: true, link: true, conflicts_with: ["mysql"]
+brew "mysql@5.6", restart_service: :changed, link: true, conflicts_with: ["mysql"]
+# 'brew install' and run a command if installer or upgraded.
+brew "postgresql@16",
+     postinstall: "${HOMEBREW_PREFIX}/opt/postgresql@16/bin/postgres -D ${HOMEBREW_PREFIX}/var/postgresql@16"
```
carlocab, line 46: It might be nice to have `postinstall` run in the same environment as `brew bundle exec` so that the formula `bin` directory is in `PATH`, etc.

carlocab (approved):

colindean (approved):
I'm very excited for this ❤️

README.md
MikeMcQuaid: Good idea but will blow up scope here a bit here so will wait until a user request or a PR comes in!

PR #1560, opened 2025-01-11
Add support for VSCodium

helgi:
VSCodium (https://vscodium.com/) is a community-driven, freely-licensed binary distribution of Microsoft’s editor VS Code.

The way extensions are exported and imported is the same as VSCode; There was a consideration to add version support (`codium --list-extensions --show-versions`) but it is not required as the IDE will update extensions itself


jacobbednarz:
i appreciate the effort here however, i'm a 👎 on the amount of duplication we have here for the multiple flavours of vscode. i would much rather see us provide an additional argument to target the binary/path for the various flavours of vscode or, the much bigger ask, generalise the plugin API so that all extensions can be extracted from core all together and push this to third party providers.

README.md
```diff
@@ -69,6 +71,9 @@ whalebrew "whalebrew/wget"
 
 # 'vscode --install-extension'
 vscode "GitHub.codespaces"
+
+# 'codium --install-extension'
+vscodium "GitHub.codespaces"
```
MikeMcQuaid, line 76: Can this not just use the existing `vscode`? If not: why not?

MikeMcQuaid (changes requested):
Thanks for the PR! I'd be 👍🏻 to adding support for VSCodium to `vscode` but 👎🏻 to add a new `vscodium` DSL, sorry!

MikeMcQuaid:
Closing due to no response.

PR #1621, opened 2025-02-25
fix(bundle): dump installed casks on Linux

caycehouse:
Adds cask dump support to Linux, by removing the default cask_installed? return. Fixes part of #1620.


MikeMcQuaid (approved):
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @caycehouse !

PR #1630, opened 2025-03-04
Add support for e.g. `brew bundle env --install`

MikeMcQuaid:
This should allow a one-liner in e.g. `eval $(brew bundle env --install)` to install all packages and setup the environment.

While we're here, also:
- make `--quiet` quieter by not outputting the success message
- remove the `--no-lock` flag as it's been hidden for a while


p-linnane (approved):
Awesome!

viniciuspalma:
Love you!

